Issue #91, opened 2023-07-06
A separate base dimension for angles


sharkdp:
This topic is something up for discussion for future SI standard versions.

Pro:
  - [1] Angles in the SI: a detailed proposal for solving the problem, https://arxiv.org/abs/2108.05704
  - [2] Angles are inherently neither length ratios nor dimensionless, https://arxiv.org/abs/1909.08389
  - [3] Brownstein K R, Angles—let’s treat them squarely, Am. J. Phys. 65 605–14 (1997).
  - [4] Dimensionless units in the SI, https://arxiv.org/abs/1409.2794
  - [5] On the status of plane and solid angles in the International System of Units (SI), https://arxiv.org/pdf/1810.12057.pdf
  - [6] On the dimension of angles and their units, https://arxiv.org/pdf/2203.12392v2.pdf

Contra:
  - [6] Dimensional angles and universal constants, https://doi.org/10.1119/1.18964
  - [7] M. Grötschel, H. Hanche-Olsen, H. Holden, Comment on ‘Angles are inherently neither length ratios nor dimensionless’.
    (and companion paper https://arxiv.org/abs/2011.05779v4)
  - [8] Comment on 'Angles in the SI: a detailed proposal for solving the problem', B P Leonard, https://iopscience.iop.org/article/10.1088/1681-7575/ac5433/meta
    (and reply to comment: https://iopscience.iop.org/article/10.1088/1681-7575/ac5434)

Review:
  - [9] Implications of adopting plane angle as a base quantity in the SI, https://arxiv.org/pdf/1604.02373.pdf

sharkdp:
See also: https://github.com/sharkdp/insect/issues/225

Issue #55, opened 2023-02-13
Config-based prelude


sharkdp:
It would be great if the prelude could not be *changed* but *configured* by the user. For example: whether or not they want US customary units. Or whether or not they want short versions of units (like `m` for `meter`, which clashes with often used variable names.. such as `m` for mass).

Something like:

config
```toml
[prelude]
short_unit_aliases = false
```

prelude:
```python
if config_setting("prelude.short_unit_aliases") {
  unit m = meter
}
```

PR #167, opened 2023-09-21
Draft: prelude with dimensionful angles

sharkdp:
see #91


archisman-panigrahi:
Will there be an option to turn this off? The idea of angles being dimensionless still did not get into mainstream physics. Having different dimensions for Torque and Energy (or Bohr radius and length) may give rise to a lot of confusions.

In my opinion (but I am saying this with my physicist hat on) it is best if this feature is turned off by default, but can be turned on by calling something like `use unit::angles_with_dimension`

sharkdp:
> Will there be an option to turn this off?

This is just a proof of concept / draft. It will not be merged as-is. And no, we will most probably not enable this by default. What I'm planning to have is something like #55, which would allow us to have one single prelude, where certain aspects could be controlled by the user (via a config file). So users could potentially opt in to this by setting `dimensionful_angles = true` in their config file.

sharkdp:
> The idea of angles being dimensionless still did not get into mainstream physics. Having different dimensions for Torque and Energy (or Bohr radius and length) may give rise to a lot of confusions.

I do agree. But note that the current situation also leads to lots of confusion. Just look at the tickets linked in https://github.com/sharkdp/numbat/issues/91#issuecomment-1736228464

PR #182, opened 2023-09-26
add metrix prefixes for barns

tomeichlersmith:
picobarn (and its shorthand `pb`) is a very common unit in High Energy Physics (HEP). Perhaps there is concern that `b` is too likely of a letter to be used elsewhere? If so, I can easily use a custom local configuration to add this shorthand for myself and HEP colleagues.

### DIFF
--- a/numbat/modules/units/misc.nbt
+++ b/numbat/modules/units/misc.nbt
@@ -9,7 +9,8 @@ unit bar: Pressure = 100 kPa
 @aliases(angstroms, Å: short)
 unit angstrom: Length = 1e-10 meter
 
-@aliases(barns)
+@metric_prefixes
+@aliases(barns, b: short)
 unit barn: Area = 1e-28 meter^2
 
 @metric_prefixes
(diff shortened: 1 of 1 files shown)


Aaron-Rumpler:
> Perhaps there is concern that `b` is too likely of a letter to be used elsewhere?
That's likely the case. For example, coming from a computer science background, I'd expect `b` to be an alias for `bit`. GNU Units on the other hand aliases `b` to `bar`.

tomeichlersmith:
Thanks for the examples from other contexts! I have removed the `b` shorthand for barns to avoid confusion, leaving the metric prefix addition for it in place.

sharkdp:
Thank you very much. I agree, it would be good to have `b` generally available for other potential uses.

@tomeichlersmith It's currently not possible to overwrite existing units, but you can use
```
@metric_prefixes
@aliases(b: short)
unit b = barn
```
to your init.nbt file to use things like `pb` for pico barn.



> High Energy Physics (HEP)

I'm always interested to hear about potential use cases. If you have some non-trivial Numbat examples to share, I'd be very happy to add them to the `examples/` folder.

PR #190, opened 2023-10-02
Allow underscores in integers with bases

irevoire:
Hey, it would be nice to allow underscores in numbers with bases.

### DIFF
--- a/numbat/src/tokenizer.rs
+++ b/numbat/src/tokenizer.rs
@@ -386,9 +386,9 @@ impl Tokenizer {
             {
                 let (base, is_digit_in_base): (_, Box<dyn Fn(char) -> bool>) =
                     match self.peek().unwrap() {
-                        'x' => (16, Box::new(|c| c.is_ascii_hexdigit())),
-                        'o' => (8, Box::new(|c| ('0'..='7').contains(&c))),
-                        'b' => (2, Box::new(|c| c == '0' || c == '1')),
+                        'x' => (16, Box::new(|c| c.is_ascii_hexdigit() || c == '_')),
+                        'o' => (8, Box::new(|c| ('0'..='7').contains(&c) || c == '_')),
+                        'b' => (2, Box::new(|c| c == '0' || c == '1' || c == '_')),
                         _ => unreachable!(),
                     };
 
--- a/numbat/src/tokenizer.rs
+++ b/numbat/src/tokenizer.rs
@@ -396,7 +396,6 @@ impl Tokenizer {
 
                 // If the first character is not a digits, that's an error.
                 if !self.peek().map(&is_digit_in_base).unwrap_or(false) {
-                    println!("here while checking {:?}", self.peek());
                     return tokenizer_error(
                         &self.current,
                         TokenizerErrorKind::ExpectedDigitInBase {
--- a/numbat/src/tokenizer.rs
+++ b/numbat/src/tokenizer.rs
@@ -800,7 +800,6 @@ fn test_tokenize_numbers() {
     "###
     );
 
-    println!("here");
     insta::assert_snapshot!(
         tokenize_reduced_pretty("0b01").unwrap(),
         @r###"
@@ -879,7 +878,6 @@ fn test_tokenize_numbers() {
         @"Error at (1, 3): `Expected base-2 digit`"
     );
 
-    println!("here");
     insta::assert_snapshot!(
         tokenize_reduced_pretty("0b1_").unwrap_err(),
         @"Error at (1, 5): `Expected base-2 digit`"
--- a/numbat/src/tokenizer.rs
+++ b/numbat/src/tokenizer.rs
@@ -394,7 +394,7 @@ impl Tokenizer {
 
                 self.advance(); // skip over the x/o/b
 
-                // If the first value is not a number that's an error.
+                // If the first character is not a digits, that's an error.
                 if !self.peek().map(&is_digit_in_base).unwrap_or(false) {
                     println!("here while checking {:?}", self.peek());
                     return tokenizer_error(
@@ -417,7 +417,7 @@ impl Tokenizer {
                     self.advance();
                 }
 
-                // Value should not end with a `_` either.
+                // Numeric literal should not end with a `_` either.
                 if last_char == Some('_')
                     || self
                         .peek()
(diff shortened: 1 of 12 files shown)


sharkdp:
Thank you very much for your contribution.

This sounds like a good idea, yes :+1: Can we please add some tests? I think we should disallow things like `0x_0` or `0x0_`. We should definitely disallow `0x_` which currently leads to a "overflow in number literal" error.

irevoire:
Hey @sharkdp, it's pretty much done, I added a lot of tests for every possible case I could think of.
To ease the addition and maintenance of new tests, I would like to introduce [insta](https://docs.rs/insta/latest/insta/) to the codebase, are you okay with this change?
We use it extensively at work, and honestly, we just won so much time since we switched on it 😅 
Writing and updating tests gets so much faster.

numbat/src/tokenizer.rs
```diff
@@ -390,13 +390,31 @@ impl Tokenizer {
 
                 self.advance(); // skip over the x/o/b
 
-                let mut has_advanced = false;
-                while self.peek().map(&is_digit_in_base).unwrap_or(false) {
+                // If the first value is not a number that's an error.
```
sharkdp, line 393:
```suggestion
                // If the first character is not a digits, that's an error.
```

numbat/src/tokenizer.rs
```diff
@@ -622,9 +640,14 @@ pub fn tokenize(input: &str, code_source_id: usize) -> Result<Vec<Token>> {
...
-fn tokenize_reduced(input: &str) -> Vec<(String, TokenKind, (u32, u32))> {
-    tokenize(input, 0)
-        .unwrap()
+fn tokenize_reduced(input: &str) -> Result<Vec<(String, TokenKind, (u32, u32))>, String> {
+    Ok(tokenize(input, 0)
+        .map_err(|e| {
+            format!(
+                "Error at {:?}: `{e}`",
+                (e.span.start.line, e.span.start.position)
+            )
```
sharkdp, line 649: :+1:

numbat/tests/interpreter.rs
```diff
@@ -43,6 +48,48 @@ fn expect_exact_failure(code: &str, expected: &str) {
     }
 }
 
+#[test]
+fn simple_value() {
```
sharkdp, line 52: Do these tests need to be full-on integration tests? These `expect_output` calls are quite heavy as we load/compile/interpret the full prelude for each of them.

I'm okay with leaving them here if it's too much work to convert them to tokenizer or parser unit tests.

numbat/src/tokenizer.rs
```diff
@@ -390,13 +390,31 @@ impl Tokenizer {
...
+                    .map(|c| is_digit_in_base(c) || c == '_')
+                    .unwrap_or(false)
+                {
+                    last_char = self.peek();
                     self.advance();
-                    has_advanced = true;
                 }
 
-                if !has_advanced
+                // Value should not end with a `_` either.
```
sharkdp, line 416:
```suggestion
                // Numeric literal should not end with a `_` either.
```

sharkdp (changes requested):
Thank you for the updates. I have just a few more comments.

> To ease the addition and maintenance of new tests, I would like to introduce [insta](https://docs.rs/insta/latest/insta/) to the codebase, are you okay with this change?

I am okay with this, but then we should probably go all in and also replace the other tokenizer and parser unit tests with insta? For the parser unit tests, this could actually be really nice and replace a lot of hand-coded ASTs. We might also be able to get rid of my custom-made `ReplaceSpans`/`replace_spans` logic by using instas `redactions` feature on the `span` fields — which would be really nice.

numbat/tests/interpreter.rs
irevoire: Oh, I didn't think about the perf.

> Do these tests need to be full-on integration tests? Also, in which way do they differ from the tokenizer unit test?

Usually, in meilisearch I duplicate the lexer/parser tests as full integration tests just to be sure that no step in the middle broke or hide a behaviour (in the parser or interpreter for example)

> These expect_output calls are quite heavy as we load/compile/interpret the full prelude for each of them.

Sooo, if I understand correctly, the last solution could be to use the `expect_output_with_context` and share the same context between each test?

sharkdp: > Usually, in meilisearch I duplicate the lexer/parser tests as full integration tests going through the whole HTTP server just to be sure that no step in the middle broke or hide a behavior (in the parser or interpreter, for example). We’ve had so many dumb issues like that in the past by only testing the parser without the interpreter.

I can certainly sympathize :smile:. I usually try to have the fine-grained tests in the unit tests / the module itself. And then one generic example in the interpreter/integration tests.

> Sooo, if I understand correctly, another solution could be to use the `expect_output_with_context` and share the same context between each test?

Yes, but the context is mutable. So we don't want to do this by default. In principle, we want a clean environment for each test. What we *could* probably do is to generate the initial context once, and then clone it? That might be a nice speedup for the unit tests.

sharkdp: So I think we can merge this as-is, unless you want to add something in this PR?

irevoire: I wanted to update the tests, as we said, but didn't really find the time to do so 😅 
If you're ok to wait, I’ll probably find the time to do it this week or next week

sharkdp: Of course, there is no rush.

irevoire:
Hey @sharkdp!
I finally got the time to take a second look at this.

On my computer the execution of your test suites goes from 23s to 2s so I guess it's a success somehow.
But I’m really not sure if I did the right thing, please take a look at [`0a60f80b`](https://github.com/sharkdp/numbat/pull/190/commits/0a60f80bdf68afbb216299ba3bc006a1228c2800).
I had to change some of your type to requires that they're `Sync` + I derived `Clone` on a lot of type without knowing if it was actually _correct_.

irevoire:
Rebased on main

sharkdp (approved):
Thank you very much!

> On my computer the execution of your test suites goes from 23s to 2s so I guess it's a success somehow.

Awesome!

> But I’m really not sure if I did the right thing, please take a look at [0a60f80b](https://github.com/sharkdp/numbat/pull/190/commits/0a60f80bdf68afbb216299ba3bc006a1228c2800).
> I had to change some of your type to requires that they're Sync + I derived Clone on a lot of type without knowing if it was actually correct.

Looks good to me :+1:

PR #191, opened 2023-10-02
Update the bnf

irevoire:
Hello, while writing a tree-sitter integration for numbat [here if you're interested](https://github.com/irevoire/tree-sitter-numbat), I realized that your bnf was not following your parser and thus rewrote most of it.

I left a few regexes at the end for the numbers, that’s not ideal, and I won't keep it in tree-sitter, but for now, it's better than nothing.

### DIFF
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -5,8 +5,8 @@
 //! statement       →   variable_decl | function_decl | dimension_decl | decorator | unit_decl | module_import | procedure_call | expression
 //!
 //! variable_decl   →   "let" identifier ( ":" type_annotation ) ? "=" expression
-//! type_annotation →   boolean | string | dimension_expr
-//! function_decl   →   "fn" identifier ( fn_decl_generic ) ? fn_decl_param "->" type_annotation "=" expression
+//! type_annotation →   "bool" | "str" | dimension_expr
+//! function_decl   →   "fn" identifier ( fn_decl_generic ) ? fn_decl_param ( "->" type_annotation ) ? ( "=" expression ) ?
 //! fn_decl_generic →   "<" ( identifier "," ) * identifier ? ">"
 //! fn_decl_param   →   "(" ( identifier ( ":" dimension_expr ) ? "," )* ( identifier ( ":" dimension_expr ) ) ? ")"
 //! dimension_decl  →   "dimension" identifier ( "=" dimension_expr ) *
@@ -18,8 +18,8 @@
 //! dimension_expr  →   dim_factor
 //! dim_factor      →   dim_power ( (multiply | divide) dim_power ) *
 //! dim_power       →   dim_primary ( power dim_exponent | unicode_exponent ) ?
-//! dim_exponent    →   number | minus dim_exponent | "(" dim_exponent ( divide dim_exponent ) ? ")"
-//! dim_primary     →   identifier | number | ( "(" dimension_expr ")"
+//! dim_exponent    →   integer | minus dim_exponent | "(" dim_exponent ( divide dim_exponent ) ? ")"
+//! dim_primary     →   identifier | "1" | ( "(" dimension_expr ")"
 //!
 //!
 //!
@@ -44,6 +44,7 @@
 //! hex_number      →   /0x[0-9a-fA-F]*/
 //! oct_number      →   /0o[0-7]*/
 //! bin_number      →   /0b[01]*/
+//! integer          →   /[0-9]([0-9_]*[0-9])?/
 //! identifier      →   [a-zA-Z_] [a-zA-Z_0-9] *
 //! boolean         →   true | false
 //! bool            →   true | false
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -2,7 +2,7 @@
 //!
 //! Grammar:
 //! ```txt
-//! statement       →   variable_decl | function_decl | dimension_decl | decorator | unit_decl | module_import | procedure_call | expression
+//! statement       →   variable_decl | function_decl | dimension_decl | unit_decl | module_import | procedure_call | expression
 //!
 //! variable_decl   →   "let" identifier ( ":" type_annotation ) ? "=" expression
 //! type_annotation →   "bool" | "str" | dimension_expr
@@ -11,7 +11,7 @@
 //! fn_decl_param   →   "(" ( identifier ( ":" type_annotation ) ? "," )* ( identifier ( ":" type_annotation ) ) ? ")"
 //! dimension_decl  →   "dimension" identifier ( "=" dimension_expr ) *
 //! decorator       →   "@" ( "metric_prefixes" | "binary_prefixes" | ( "aliases(" list_of_aliases ")" ) )
-//! unit_decl       →   "unit" ( ":" dimension_expr ) ? ( "=" expression ) ?
+//! unit_decl       →   decorator * "unit" ( ":" dimension_expr ) ? ( "=" expression ) ?
 //! procedure_call  →   ( "print" | "assert_eq" | "type" ) "(" arguments? ")"
 //! module_import   →   "use" ident ( "::" ident) *
 //!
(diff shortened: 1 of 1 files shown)


sharkdp:
> while writing a tree-sitter integration for numbat [here if you're interested](https://github.com/irevoire/tree-sitter-numbat)

Cool! What are you doing this for specifically?

> I realized that your bnf was not following your parser and thus rewrote most of it.

I was afraid it would be, so thank you very much for updating it!

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
 //!
 //! Grammar:
 //! ```txt
-//! statement       →   expression | variable_decl | function_decl | dimension_decl | unit_decl | procedure_call | module_import
+//! statement       →   variable_decl | function_decl | dimension_decl | decorator | unit_decl | module_import | procedure_call | expression
```
sharkdp, line 5: > `decorator`

I think this is wrong? A standalone decorator is not a statement. It can precede a unit definition. And together with the unit definition, it forms a statement.

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//!
+//! variable_decl   →   "let" identifier ( ":" type_annotation ) ? "=" expression
+//! type_annotation →   boolean | string | dimension_expr
```
sharkdp, line 8: I think this is wrong? A type annotation would not contain `true`/`false` or an actual string, but one of the two keywords `bool` or `str` (or a `dimension_expr`).

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//! function_decl   →   "fn" identifier ( fn_decl_generic ) ? fn_decl_param "->" type_annotation "=" expression
```
sharkdp, line 9: Return type annotations are optional.
```
fn f(x: Length) = x + 1 meter
```

Function bodies are also optional, when declaring FFI functions:
```
fn log(x: Scalar) -> Scalar
```

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//! fn_decl_generic →   "<" ( identifier "," ) * identifier ? ">"
```
sharkdp, line 10: Should this be
```suggestion
//! fn_decl_generic →   "<" ( identifier "," ) * identifier ">"
```

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//! fn_decl_param   →   "(" ( identifier ( ":" dimension_expr ) ? "," )* ( identifier ( ":" dimension_expr ) ) ? ")"
```
sharkdp, line 11: With #193 close to being merged, we could already change this to
```suggestion
//! fn_decl_param   →   "(" ( identifier ( ":" type_annotation ) ? "," )* ( identifier ( ":" type_annotation ) ) ? ")"
```

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//! dimension_decl  →   "dimension" identifier ( "=" dimension_expr ) *
+//! decorator       →   "@" ( "metric_prefixes" | "binary_prefixes" | ( "aliases(" list_of_alsiases ")" ) )
```
sharkdp, line 13: typo
```suggestion
//! decorator       →   "@" ( "metric_prefixes" | "binary_prefixes" | ( "aliases(" list_of_aliases ")" ) )
```

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//! unit_decl       →   "unit" ( ":" dimension_expr ) ? ( "=" expression ) ?
+//! procedure_call  →   ( "print" | "assert_eq" | "type" ) ( "(" arguments? ")" ) ?
```
sharkdp, line 15: Should this be
```suggestion
//! procedure_call  →   ( "print" | "assert_eq" | "type" ) "(" arguments? ")"
```

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//! module_import   →   "use" ident ( "::" ident) *
+//!
+//! dimension_expr  →   dim_factor
+//! dim_factor      →   dim_power ( (multiply | divide) dim_power ) *
+//! dim_power       →   dim_primary ( power dim_exponent | unicode_exponent ) ?
+//! dim_exponent    →   number | minus dim_exponent | "(" dim_exponent ( divide dim_exponent ) ? ")"
```
sharkdp, line 21: Numbers in dimension exponents are not as flexible as normal numbers. They are basically just integers.

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
+//! dim_primary     →   identifier | number | ( "(" dimension_expr ")"
```
sharkdp, line 22:
```suggestion
//! dim_primary     →   identifier | "1" | "(" dimension_expr ")"
```

numbat/src/parser.rs
```diff
@@ -2,29 +2,56 @@
...
-//! primary         →   number | hex-number | oct-number | bin-number | identifier | "(" expression ")"
+//! primary         →   boolean | string | hex_number | oct_number | bin_number | number | identifier | "(" expression ")"
+//!
+//! number          →   /[0-9][0-9_]*(\.([0-9][0-9_]*)?)?([eE][+-]?[0-9][0-9_]*)?/
+//! hex_number      →   /0x[0-9a-fA-F]*/
+//! oct_number      →   /0o[0-7]*/
+//! bin_number      →   /0b[01]*/
+//! identifier      →   [a-zA-Z_] [a-zA-Z_0-9] *
+//! boolean         →   true | false
+//! bool            →   true | false
```
sharkdp, line 49: Remove one of them?

numbat/src/parser.rs
```diff
@@ -332,6 +359,7 @@ impl<'a> Parser<'a> {
             if let Some(fn_name) = self.match_exact(TokenKind::Identifier) {
                 let function_name_span = self.last().unwrap().span;
                 let mut type_parameters = vec![];
+                // Parsing the generic parameters if there is any
```
sharkdp, line 362:
```suggestion
                // Parsing the generic parameters if there are any
```

sharkdp (changes requested):
Thank you very much! I think now that this is polished, we could even add it to the user documentation.

numbat/src/parser.rs
irevoire: Ooooh ok, my bad I went too fast on this part!

irevoire:
> A standalone decorator is not a statement. It can precede a unit definition. And together with the unit definition, it forms a statement.

Yeah, you’re right. I looked at the code, and it’s a bit more complex than I thought 😅

I tried to fix it by removing the decorator from the list of statements and updating the `unit` declaration to contain a list of `decorator` at the beginning. I guess it works?

> Cool! What are you doing this for specifically?

I used `numbat` to calculate how much more oil height I would get from a ladle of oil in a wok with a round or flat bottom.
It took like ~20 lines of code and was unreadable without good syntax highlighting.
Plus, I’m using [helix](https://helix-editor.com/), and being able to select parameters, comments, etc is pretty nice.

sharkdp (approved):

sharkdp:
Thank you for the updates!

> I used `numbat` to calculate how much more oil height I would get from a ladle of oil in a wok with a round or flat bottom.
> It took like ~20 lines of code and was unreadable without good syntax highlighting.
> Plus, I’m using [helix](https://helix-editor.com/), and being able to select parameters, comments, etc is pretty nice.

Thank you for sharing. I'm always interested to hear how people use (Insect or) Numbat. Concerning syntax highlighting, I have already created simple syntaxes for [vim](https://github.com/sharkdp/numbat/blob/master/assets/numbat.vim), [VS Code](https://github.com/sharkdp/numbat/tree/master/vscode-extension) and [Sublime Text / `bat`](https://github.com/sharkdp/numbat/blob/master/assets/numbat.sublime-syntax). I should probably document how to use those. Having a tree-sitter grammar in addition is great.

sharkdp:
I made a few other changes to the BNF. Now it's compatible with [this EBNF parser](https://www.bottlecaps.de/rr/ui), which can generate railroad diagrams like these: https://numbat.dev/numbat-syntax.html

PR #193, opened 2023-10-03
Add support for arbitrary types in function parameters

### LINKED ISSUE #161: Functions that take non-dtype arguments
Make it possible to define sth. like
```rs
fn append(s1: str, s2: str) -> str = "{s1}{s2}"

fn and(b1: bool, b2: bool) -> bool = if b1 then b2 else false
```

It also allows us to use `exchange_rate("USD")` etc. instead of having tens of different functions.

sharkdp:
closes #161

### DIFF
--- a/numbat/modules/core/booleans.nbt
+++ b/numbat/modules/core/booleans.nbt
@@ -1,2 +1,0 @@
-fn and(a: bool, b: bool) = if a then b else false
-fn or(a: bool, b: bool) = if a then true else b
--- a/numbat/modules/core/booleans.nbt
+++ b/numbat/modules/core/booleans.nbt
@@ -1,0 +1,2 @@
+fn and(a: bool, b: bool) = if a then b else false
+fn or(a: bool, b: bool) = if a then true else b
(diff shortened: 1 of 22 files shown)


numbat/modules/core/strings.nbt
```diff
@@ -0,0 +1,4 @@
+use core::scalar
+
+fn str_length(a: str) -> Scalar
+fn str_concat(a: str, b: str) -> str = "{a}{b}"
```
sharkdp, line 4: Think about better names. Or remove for now.

numbat/modules/core/booleans.nbt
```diff
@@ -0,0 +1,2 @@
+fn and(a: bool, b: bool) = if a then b else false
+fn or(a: bool, b: bool) = if a then true else b
```
sharkdp, line 2: Is it worth to add these? Or should we just add logical operators?

eminence: `and` and `or` are so fundamental to booleans, that I would expect that either these functions or infix operators would be provided out-of-the-box

sharkdp: I agree: https://github.com/sharkdp/numbat/issues/196

PR #208, opened 2023-10-17
Add smoot unit

Mstrodl:
From: https://en.wikipedia.org/wiki/Smoot


numbat/modules/units/misc.nbt
```diff
@@ -63,6 +63,9 @@ unit footballfield: Area = 105 m × 68 m
 # Olympic-size swimming pool (FR3)
 unit swimmingpool: Volume = 50 m × 25 m × 2 m
 
+# https://en.wikipedia.org/wiki/Smoot
+unit smoot: Length = 5 feet + 7 inch
```
sharkdp, line 67: Tests are failing because `feet` and `inch` are not defined here. This module is only importing `units::si`, and I'd like to keep it that way.

Maybe let's add a new `units::humorous` module?

Mstrodl: That's totally on me for not actually running the tests locally :facepalm:, I'll do that

Mstrodl:
(Sorry, don't review yet, need to actually include it in the prelude..)

Mstrodl:
Should be good now, sorry!

sharkdp:
Thanks for the update

PR #209, opened 2023-10-20
Add support for prefix unary plus operator

### LINKED ISSUE #120: Support prefix unary '+'?

irevoire:
Fix #120

Hey, I was looking for easy-to-implement issues just to improve my understanding of the code base.
Let me know if you have other ideas


numbat-cli/tests/integration.rs
```diff
@@ -33,8 +33,8 @@ fn pass_expression_on_command_line() {
         .arg("--expression")
         .arg("2 ++ 3")
         .assert()
-        .failure()
-        .stderr(predicates::str::contains("while parsing"));
+        .success()
+        .stdout(predicates::str::contains("5"));
```
sharkdp, line 37: This test was here to make sure that we fail with a non-zero exit status on parser errors. I suggest we revert this and change the expression instead (e.g. `2 /+ 3`.

sharkdp (approved):
Thank you very much! Just one minor comment.

numbat-cli/tests/integration.rs
irevoire: Yeah, except, hehe, now `2 / +3` is a valid expression 😎 

I updated it to `2 +/ 3`

sharkdp: Oh. I should have invested more than 2 seconds of thought into that example :smile:

sharkdp (approved):
Thank you

PR #210, opened 2023-10-20
Add a first very basic benchmarks that imports the prelude

irevoire:
Hey, I’ve noticed that you were talking about the performances of numbat on a few issues / PRs.
To be able to easily measure the impact of changes, I think it would be great to have some benchmarks.

Since I don't know yet what could be relevant, I only added the most basic benchmark: `use prelude`.
I also used `criterion` as it seems to be the most common option in the rust ecosystem.
And the way I expect people to use it is to run `cargo bench` on their computer without their changes. Then re-run it with your change, and you should see something like this:
```
     Running benches/prelude.rs (/Users/irevoire/numbat/target/release/deps/prelude-ddd949e31aa30b32)
Import prelude          time:   [13.064 ms 13.111 ms 13.161 ms]
                        change: [+0.2677% +0.7286% +1.1605%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstream"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


sharkdp (approved):
Thank you!

Yes, parsing+interpreting the prelude is an important benchmark because we do care about the startup speed, especially for the CLI. Insect was an order of magnitude slower (even after many optimizations):

|:---|---:|---:|---:|---:|
| `numbat -e "2 + 3"` | 33.1 ± 3.8 | 27.3 | 53.2 | 1.00 |
| `insect "2 + 3"` | 591.2 ± 28.6 | 551.0 | 632.1 | 17.85 ± 2.23 |

PR #212, opened 2023-10-21
Increase the usage of snapshot testing in the codebase

irevoire:
Hey, after my PR where I introduced `insta` to the code base and in [this comment](https://github.com/sharkdp/numbat/pull/190#pullrequestreview-1662007088), you mentioned that it would be nice to use insta everywhere.

I rewrote a first batch of tests on the interpreter to see what it could look like.
Before spending more time on the parser tests, I thought I would wait for your feedback on this first part 😅 

Let me know if that’s the kind of stuff you were expecting.

### DIFF
--- a/numbat/src/interpreter.rs
+++ b/numbat/src/interpreter.rs
@@ -90,12 +90,11 @@ pub trait Interpreter {
 
 #[cfg(test)]
 mod tests {
-    use crate::unit::Unit;
-    use crate::{bytecode_interpreter::BytecodeInterpreter, prefix_transformer::Transformer};
-
     use super::*;
+    use crate::{bytecode_interpreter::BytecodeInterpreter, prefix_transformer::Transformer};
+    use insta::assert_snapshot;
 
-    static TEST_PRELUDE: &str = "
+    const TEST_PRELUDE: &str = "
         dimension Scalar = 1
 
         dimension Length
@@ -122,8 +121,13 @@ mod tests {
         fn maximum<D>(xs: D…) -> D
         fn minimum<D>(xs: D…) -> D";
 
-    fn get_interpreter_result(input: &str) -> Result<InterpreterResult> {
-        let full_code = format!("{prelude}\n{input}", prelude = TEST_PRELUDE, input = input);
+    #[track_caller]
+    fn get_interpreter_result(input: impl AsRef<str>) -> Result<InterpreterResult> {
+        let full_code = format!(
+            "{prelude}\n{input}",
+            prelude = TEST_PRELUDE,
+            input = input.as_ref()
+        );
         let statements = crate::parser::parse(&full_code, 0)
             .expect("No parse errors for inputs in this test suite");
         let statements_transformed = Transformer::new()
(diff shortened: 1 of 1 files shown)


numbat/src/interpreter.rs
```diff
@@ -137,137 +141,112 @@ mod tests {
...
 
-        assert_evaluates_to(
+        assert_snapshot!(evaluates(
             "dimension Pixel
              @aliases(px: short)
              unit pixel : Pixel
-             2 * pixel",
-            Quantity::from_scalar(2.0) * Quantity::from_unit(Unit::new_base("pixel", "px")),
-        );
+             2 * pixel"), @"2 px");
```
sharkdp, line 183: This is shorter and looks nicer than what we had, but it's also not quite the same. Since you're using `format!` for comparison here, these tests  (this one and below) now depend on the `Display` implementations of `Quantity`/`Number`/`Unit`, which are not-trivial. One problem with this is that we might have to update the tests frequently when we change those Display impls (I understand that this is much more convenient now with `insta`). The other problem is that this might hide things (think of rounding, for example).

What do you think?

numbat/src/interpreter.rs
```diff
@@ -137,137 +141,112 @@ mod tests {
...
-        );
+        assert_snapshot!(evaluates("minimum(1, 2, 0, -3)"), @"-3");
+        assert_snapshot!(evaluates(
+            "minimum(2 m, 150 cm)"), @"1.5 m");
     }
 
     #[test]
     fn division_by_zero_raises_runtime_error() {
-        assert_runtime_error("1/0", RuntimeError::DivisionByZero);
+        assert_snapshot!(evaluates("1/0"), @"DivisionByZero: `Division by zero`");
```
sharkdp, line 241: This is somewhat similar. These tests now depend on the exact error message and might need to be updated whenever we change messages.

numbat/src/interpreter.rs
irevoire: Yeah, it’s what I wanted to see with you.
On one side, it hides information, on the other, don't we want to notice when the final message changes?

IMO, the other options are:
- Use the `Debug` implem
- Display the `Display` + `Debug` implem side by side (`format!("{} => {:?}")`) just to be sure we don't miss anything

numbat/src/interpreter.rs
irevoire: Yeah, that’s similar, but here, I feel like it's 100% what we want, no?
The tests are easier to review, we can _read_ the error directly, and it shows us what will effectively be returned to the end user 🤔 

And in the end, when you change an error message, you can ensure straight away from the tests that it didn't have any side effect on other stuff and that it displays as you want. (this one is easy, but some error messages may contain variables and more complex formatting).

----

You have the last word, but personally, I think the only valuable information when testing errors as a binary is the final error message.
And for a library, it’s both the error message and the error type; that’s why I snapshotted both.

numbat/src/interpreter.rs
sharkdp: > Display the `Display` + `Debug` implem side by side (`format!("{} => {:?}")`) just to be sure we don't miss anything

I like this idea. If the output is not too noisy, this might solve all issues.

numbat/src/interpreter.rs
sharkdp: > You have the last word, but personally, I think the only valuable information when testing errors as a binary is the final error message.

I think I agree.

> And for a library, it’s both the error message and the error type; that’s why I snapshotted both.

I mean.. I want both to be correct, but the important part about this test is that it asserts that divisions by zero *do raise* a runtime error (and the test doesn't care if the message is "Division by zero" or "Numerical error" or similar). Of course I also want the message to be correct, but I'm not sure if that needs a unit test. Ideally, I'd like to have that error message just once in my code base. But maybe I just need to adjust to the idea of snapshot tests. I'm okay with trying :smile:

numbat/src/interpreter.rs
irevoire: Hey, once again, I’m late 😅 

What do you think of this last commit? 
Do you think it's too noisy, or are we good, and I can start applying the same techniques to more files?

sharkdp:
If I'm being really honest, I don't see that this adds a lot of value in this particular test suite. I feel it was easier to read before. But I also have to admit, that I still haven't really worked with `insta`.

Maybe it would be better to start with tests that would more clearly (IMO) benefit from this treatment, such as the parser unit tests?

irevoire:
Ok, will do it in another PR

PR #222, opened 2023-11-02
add `help` (and `?`) commands to numbat REPL

### LINKED ISSUE #175: Add help command
I propose to add a `help` command (similar to insect), that will show examples of 
1. How to define new variables and units
2. How to load unused modules (like `use units::stoney`)
3. basic things like https://github.com/sharkdp/numbat/raw/master/assets/numbat-interactive.png

And it should have a link to the documentation https://numbat.dev/doc/

tomeichlersmith:
This resolves #175 .

I implemented an `insect`-like `help` command for the REPL (with `?` as an alias) by writing the package description, running a few examples through the REPL itself and then printing the link to the online documentation. The examples are printed along with the prompt to show the user how it would look and then the result of the examples are printed just like any other user command to show the user how it would look on their computer.

I like this so that the output of the examples are automatically updated along with any output formatting changes in numbat, but this has the downside of potentially polluting the user namespace if any of the examples have `let` or `use` in them. For now, I just avoid using those commands in the examples.

### DIFF
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -286,12 +286,7 @@ impl Cli {
                                 return Ok(());
                             }
                             "help" | "?" => {
-                                let pretty_print = match self.args.pretty_print {
-                                    PrettyPrintMode::Always => true,
-                                    PrettyPrintMode::Never => false,
-                                    PrettyPrintMode::Auto => true,
-                                };
-                                let help = help_markup(pretty_print);
+                                let help = help_markup();
                                 print!("{}", ansi_format(&help, true));
                                 // currently, the ansi formatter adds indents
                                 // _after_ each newline and so we need to manually
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -292,7 +292,11 @@ impl Cli {
                                     PrettyPrintMode::Auto => true,
                                 };
                                 let help = help_markup(pretty_print);
-                                print!("{}", ansi_format(&help, false));
+                                print!("{}", ansi_format(&help, true));
+                                // currently, the ansi formatter adds indents
+                                // _after_ each newline and so we need to manually
+                                // add an extra blank line to absorb this indent
+                                println!();
                             }
                             _ => {
                                 let result = self.parse_and_evaluate(
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -13,7 +13,7 @@ use numbat::markup as m;
 use numbat::module_importer::{BuiltinModuleImporter, ChainedImporter, FileSystemImporter};
 use numbat::pretty_print::PrettyPrint;
 use numbat::resolver::CodeSource;
-use numbat::{Context, ExitStatus, InterpreterResult, NumbatError};
+use numbat::{Context, InterpreterResult, NumbatError};
 use numbat::{InterpreterSettings, NameResolutionError};
 
 use anyhow::{bail, Context as AnyhowContext, Result};
@@ -30,7 +30,13 @@ use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::{fs, thread};
 
-type ControlFlow = std::ops::ControlFlow<numbat::ExitStatus>;
+#[derive(Debug, PartialEq, Eq)]
+pub enum ExitStatus {
+    Success,
+    Error,
+}
+
+type ControlFlow = std::ops::ControlFlow<ExitStatus>;
 
 const PROMPT: &str = ">>> ";
 
@@ -389,7 +395,6 @@ impl Cli {
                         ControlFlow::Continue(())
                     }
                     InterpreterResult::Continue => ControlFlow::Continue(()),
-                    InterpreterResult::Exit(exit_status) => ControlFlow::Break(exit_status),
                 }
             }
             Err(NumbatError::ResolverError(e)) => {
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -8,6 +8,7 @@ use highlighter::NumbatHighlighter;
 
 use itertools::Itertools;
 use numbat::diagnostic::ErrorDiagnostic;
+use numbat::help::help_markup;
 use numbat::markup as m;
 use numbat::module_importer::{BuiltinModuleImporter, ChainedImporter, FileSystemImporter};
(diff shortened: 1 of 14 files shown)


sharkdp:
> I implemented an `insect`-like `help` command for the REPL (with `?` as an alias) by writing the package description, running a few examples through the REPL itself and then printing the link to the online documentation

Thank you very much!

> The examples are printed along with the prompt to show the user how it would look and then the result of the examples are printed just like any other user command to show the user how it would look on their computer.

I like it! Maybe we could indent everything by 4 spaces to clarify what belongs to the help text and what not?



> I like this so that the output of the examples are automatically updated along with any output formatting changes in numbat, but this has the downside of potentially polluting the user namespace if any of the examples have `let` or `use` in them.

This is also problematic for another reason. If you run `numbat --no-prelude` and then type `help`, you'll get something like "unknown identifier: km".

We could create a fresh Numbat `Context` for the purpose of running the `help` examples. This way, there would be a clean separation. This also leads me to a general comment/question:


It would be great if we could somehow move this functionality into the `numbat` crate (the library, not the CLI app). This way, we could reuse this code in the web application as well.

tomeichlersmith:
Thanks for taking a look at this so quickly :smile: 

- I also like the idea of making this available to the web as well as the CLI app. (I had erroneously assumed that this code would be in the web app as well).
- Creating a fresh context to run the help will be nice for the reason you describe (and will help me familiarize myself with the `Context`) - do you think the examples should change depending on if a prelude is loaded?

tomeichlersmith (commented):
Alright, I _think_ I've made the necessary updates to enable this help to be printed in both the CLI and the web apps. There are still two more things to do

1. I need to actually test and see what the output looks like on the web app. I am not familiar with web-dev and so I'll need to figure out how to get `wasm-pack` and `npm` and stuff so I can do a local test and serve.
2. I found a very minor bug in the ANSIFormatter. When `indent` is enabled, the indenting spaces are added _after each_ new line character in the markup. This means that the next line that would be printed by the terminal emulator will have the indent. Instead of resolving this issue, I just moved the last newline from the markup into an extra `println!()` call.

numbat-cli/src/main.rs
```diff
@@ -284,6 +285,19 @@ impl Cli {
                             "quit" | "exit" => {
                                 return Ok(());
                             }
+                            "help" | "?" => {
+                                let pretty_print = match self.args.pretty_print {
+                                    PrettyPrintMode::Always => true,
+                                    PrettyPrintMode::Never => false,
+                                    PrettyPrintMode::Auto => true,
+                                };
```
sharkdp, line 293: I would switch off pretty printing regardless of user settings for these help examples. Makes the help text more concise. What do you think?

numbat/src/help.rs
```diff
@@ -0,0 +1,146 @@
+/**
+ * Print a help, linking the documentation and live-running some examples
+ * in an isolated context
+ */
```
sharkdp, line 4: Minor: use `///` for doc comments.

numbat/src/help.rs
```diff
@@ -0,0 +1,146 @@
...
+                    .fold(m::empty(), |accumulated_mk, statement| {
+                        accumulated_mk
+                            + m::nl()
+                            + m::whitespace("  ")
+                            + statement.pretty_print()
+                            + m::nl()
+                    });
+            }
+
+            match interpreter_result {
```
sharkdp, line 47: Hm. This code is now repeated (more or less) three times: in numbat-cli/src/main.rs, in numbat-wasm/src/lib.rs, and now here. It would be great if we could move this to a common place inside the `numbat` crate.

numbat-cli/src/main.rs
tomeichlersmith: Yea that makes sense, I can put that into the `help_markup` function as well to help make it more concise.

numbat/src/help.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    full_output
+}
+
+pub fn help_markup(pretty_print: bool) -> m::Markup {
+    let mut output = m::nl()
+        + m::keyword("numbat")
+        + m::space()
+        + m::text(env!("CARGO_PKG_DESCRIPTION"))
```
sharkdp, line 123: I like that you want to avoid the duplication here, but I'd rather see a proper sentence starting with "Numbat is a statically …", instead of "numbat A statically …". Let's just copy that sentence here?

A newline between the description and the first example would also be great.

numbat/src/help.rs
```diff
@@ -0,0 +1,146 @@
...
+    let _use_prelude_output = evaluate_example(&mut example_context, "use prelude", false);
+    for example in examples.iter() {
+        output += m::text(">>> ") + m::text(example) + m::nl();
+        output += evaluate_example(&mut example_context, example, pretty_print);
+        output += m::nl();
+    }
+    output += m::text("Full documentation:")
+        + m::space()
+        + m::keyword("https://numbat.dev/doc/")
+        + m::nl();
```
sharkdp, line 144: Maybe let's move this up? "You can read the full documentation at …. For now, you can try one of these examples: …"

sharkdp (changes requested):
Thank you for the updates!

sharkdp:
> * do you think the examples should change depending on if a prelude is loaded?

Hm. I think we should just show the same thing always.

> I need to actually test and see what the output looks like on the web app. I am not familiar with web-dev and so I'll need to figure out how to get `wasm-pack` and `npm` and stuff so I can do a local test and serve.

I'd also be okay with doing this for you. No worries.

> I found a very minor bug in the ANSIFormatter. When `indent` is enabled, the indenting spaces are added _after each_ new line character in the markup. This means that the next line that would be printed by the terminal emulator will have the indent. Instead of resolving this issue, I just moved the last newline from the markup into an extra `println!()` call.

Okay. If you have a good idea on how to improve the situation, please let me know.

numbat/src/help.rs
tomeichlersmith: I think I got this centralization working[^1], please look to see if it makes sense and is "clean".

[^1]: With the caveat that I still haven't tried running the wasm yet.

sharkdp (approved):
Thank you very much. I pushed a couple of updates — I hope you don't mind.

sharkdp:
It works great in the web version as well. Changes are live on https://numbat.dev/

PR #225, opened 2023-11-04
Add foot-candles unit

hotgluebanjo:
Adds support for foot-candles, analogous to lux: https://en.wikipedia.org/wiki/Foot-candle

Unsure of the `us_customary` location.


mhmdanas:
Thank you!

> Unsure of the `us_customary` location.

Should be fine I think.

sharkdp:
Thank you! Changes are live on https://numbat.dev/

PR #227, opened 2023-11-04
Add CLI config file

### LINKED ISSUE #97: User configuration
Things that a user might want to configure:

- [x] REPL Prompt
- [x] whether or not to show the "Numbat" font
- [x] Whether or not to pre-fetch currency exchange rates
- [x] Pretty printing: on/off
- [ ] Paths to the module folder
- [ ] Localization settings (decimal separator)
- [ ] whether or not to use unicode exponents like ², ³, … instead of ^2, ^3
- [ ] number of significant digits that are displayed

See also: #55

### LINKED ISSUE #139: Allow customizing behavior of exchange rate fetching
Not everybody needs exchange rate information, or uses it often; some people will also take issue with Numbat using the network without their permission or awareness. Therefore, I propose adding the following options (or at least an appropriate subset):
1. Fetching the rates on startup (the current behavior, adding a configurable time-out is also a good idea)
2. Fetching them on first currency conversion, or something similar (I'd personally use this)
3. Any other possible behaviors?

mhmdanas:
Fixes #97, fixes #139.

TODO:
- [x] Add documentation

### DIFF
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -24,11 +24,35 @@ use rustyline::{
     Validator,
 };
 use rustyline::{EventHandler, Highlighter, KeyCode, KeyEvent, Modifiers};
+use serde::Deserialize;
+use toml;
 
 use std::io::IsTerminal;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::{fs, thread};
+
+#[derive(Deserialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "kebab-case")]
+enum ExchangeRateFetchingPolicy {
+    #[default]
+    Prefetch,
+    FetchOnFirstUse,
+    DontFetch,
+}
+
+#[derive(Deserialize, Default)]
+#[serde(rename_all = "kebab-case")]
+struct Config {
+    exchange_rate_fetching_policy: ExchangeRateFetchingPolicy,
+    quiet: bool,
+    #[serde(default = "default_prompt")]
+    prompt: String,
+}
+
+fn default_prompt() -> String {
+    ">>> ".to_string()
+}
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum ExitStatus {
@@ -37,8 +61,6 @@ pub enum ExitStatus {
 }
 
 type ControlFlow = std::ops::ControlFlow<ExitStatus>;
-
-const PROMPT: &str = ">>> ";
 
 #[derive(Debug, Clone, Copy, PartialEq, ValueEnum)]
 enum PrettyPrintMode {
@@ -111,6 +133,7 @@ struct NumbatHelper {
 
 struct Cli {
     args: Args,
+    config: Config,
     context: Arc<Mutex<Context>>,
 }
 
@@ -134,12 +157,24 @@ impl Cli {
         Self {
             context: Arc::new(Mutex::new(context)),
             args,
+            config: Config::default(),
         }
     }
 
     fn run(&mut self) -> Result<()> {
         let load_prelude = !self.args.no_prelude;
         let load_init = !(self.args.no_prelude || self.args.no_init);
+
+        let user_config_path = Self::get_config_path().join("config.toml");
+
+        if let Ok(contents) = fs::read_to_string(&user_config_path) {
+            match toml::from_str(&contents) {
+                Ok(config) => self.config = config,
+                Err(_) => {
+                    bail!("Failed to parse config file")
+                }
+            }
+        }
 
         if load_prelude {
             let result = self.parse_and_evaluate(
@@ -169,16 +204,25 @@ impl Cli {
             }
         }
 
-        if load_prelude {
+        if load_prelude
+            && self.config.exchange_rate_fetching_policy != ExchangeRateFetchingPolicy::DontFetch
+        {
             {
                 self.context
                     .lock()
                     .unwrap()
                     .load_currency_module_on_demand(true);
             }
-            thread::spawn(move || {
-                numbat::Context::prefetch_exchange_rates();
-            });
+
+            match self.config.exchange_rate_fetching_policy {
+                ExchangeRateFetchingPolicy::Prefetch => {
+                    thread::spawn(move || {
+                        numbat::Context::prefetch_exchange_rates();
+                    });
+                }
+                ExchangeRateFetchingPolicy::FetchOnFirstUse
+                | ExchangeRateFetchingPolicy::DontFetch => {}
+            }
         }
 
         let pretty_print_mode =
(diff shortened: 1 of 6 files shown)


numbat-cli/src/main.rs
```diff
@@ -24,12 +24,27 @@ use rustyline::{
...
 
 use std::io::IsTerminal;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::{fs, thread};
 
+#[derive(Deserialize)]
+struct Config {
+    #[serde(
+        default = "default_prefetch_exchange_rates",
```
mhmdanas, line 38: Couldn't find a less ugly way to have a default value. :/

numbat-cli/src/main.rs
```diff
@@ -24,12 +24,27 @@ use rustyline::{
...
+        rename = "prefetch-exchange-rates"
```
mhmdanas, line 39: I prefer kebab case, but I'm also fine with using snake case if you want.

numbat-cli/src/main.rs
```diff
@@ -141,6 +156,17 @@ impl Cli {
         let load_prelude = !self.args.no_prelude;
         let load_init = !(self.args.no_prelude || self.args.no_init);
 
+        let user_config_path = Self::get_config_path().join("config.toml");
+
+        let config = match toml::from_str::<Config>(
+            &fs::read_to_string(&user_config_path).unwrap_or("".to_string()),
```
mhmdanas, line 162: Is there a better way to initialize an "empty"/default config than parsing an empty string?

numbat-cli/src/main.rs
```diff
@@ -141,6 +156,17 @@ impl Cli {
...
+        ) {
+            Ok(config) => config,
+            Err(_) => {
+                bail!("Failed to read config file")
```
mhmdanas, line 166:
```suggestion
                bail!("Failed to parse config file")
```

numbat-cli/src/main.rs
sharkdp: Sounds good. You can do this for all fields using `#[serde(rename_all = "kebab-case")]`

numbat-cli/src/main.rs
sharkdp: Yes. Let's implement `Default` for `Config` (not manually, but ideally by deriving it). And then we use `Config::default()` if no config file has been found.

numbat-cli/src/main.rs
```diff
@@ -24,12 +24,41 @@ use rustyline::{
...
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 use std::{fs, thread};
 
+#[derive(Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+enum ExchangeRateFetchingPolicy {
+    Prefetch,
+    FetchOnFirstUse,
+    DontFetch,
```
mhmdanas, line 41: Need to think about best names here (and also for the names of config options themselves).

numbat-cli/src/main.rs
```diff
@@ -24,12 +24,41 @@ use rustyline::{
...
+    }
+}
+
+#[derive(Deserialize, Default)]
+#[serde(rename_all = "kebab-case")]
+struct Config {
+    exchange_rate_fetching_policy: ExchangeRateFetchingPolicy,
+    quiet: bool,
+    #[serde(default = "default_prompt")]
+    prompt: String,
```
mhmdanas, line 50: Just noticed that with this, the derived `default` method returns a config with an empty prompt; not sure how to fix it without restoring to manually writing a `Default` impl...

numbat-cli/src/main.rs
sharkdp: I like what you have.

numbat-cli/src/main.rs
sharkdp: I think the `default_prompt` option is okay. Another thing you can do is to add a `struct Prompt(String)` and implement `Default` for that, but I think that's overkill here.

numbat-cli/src/main.rs
```diff
@@ -134,13 +157,25 @@ impl Cli {
...
 
+        let user_config_path = Self::get_config_path().join("config.toml");
+
+        if let Ok(contents) = fs::read_to_string(&user_config_path) {
+            match toml::from_str(&contents) {
+                Ok(config) => self.config = config,
+                Err(_) => {
+                    bail!("Failed to parse config file")
+                }
+            }
```
sharkdp, line 176: I think you can show the inner TOML parsing error here with something like
```suggestion
            let config = toml::from_str(&contents).context("Error while parsing config file")?;
```

numbat-cli/src/main.rs
```diff
@@ -169,16 +204,25 @@ impl Cli {
...
+
+            match self.config.exchange_rate_fetching_policy {
+                ExchangeRateFetchingPolicy::Prefetch => {
+                    thread::spawn(move || {
+                        numbat::Context::prefetch_exchange_rates();
+                    });
+                }
+                ExchangeRateFetchingPolicy::FetchOnFirstUse
+                | ExchangeRateFetchingPolicy::DontFetch => {}
+            }
```
sharkdp, line 225: Let's replace that with a simple `if self.config.exchange_rate_fetching_policy ==                  ExchangeRateFetchingPolicy::Prefetch`?

sharkdp (changes requested):
Cool — thank you very much!

numbat-cli/src/main.rs
mhmdanas: Thanks, here are my concerns though, if you want to mull on them some more:
- `"dont-fetch"` looks kind of weird to be honest
- `exchange-rate-fetching-policy` is kind of long
- maybe `show-intro` would be clearer than `quiet`, but we're already using `quiet` as a command-line option so I'm leaning towards just using `quiet`

numbat-cli/src/main.rs
mhmdanas: That's what I did initially, but then I changed it to this so that we'd be forced to account for any new/renamed values of `ExchangeRateFetchPolicy`, but it's probably unnecessary, I agree.

numbat-cli/src/main.rs
```diff
@@ -225,7 +276,7 @@ impl Cli {
     fn repl(&mut self) -> Result<()> {
         let interactive = std::io::stdin().is_terminal();
         let history_path = self.get_history_path()?;
-        let quiet_startup = self.args.quiet;
+        let intro_banner = self.args.intro_banner.unwrap_or(self.config.intro_banner);
```
sharkdp, line 279: So what we do in other applications is the following: we load the config from the file (if there is any). And then we immediately *modify* that config struct with the available environment variables (which overwrite config options) and command line arguments (which overwrite env variables and config settings). This way, there is just one central place for configuration. And we could maybe get rid of `self.args`.

sharkdp (approved):

PR #229, opened 2023-11-06
Add documentation for nixos installation

giomf:
This will add the needed documentation how to install it on a nixos based system or any system where the nix package manager is installed.

### DIFF
--- a/book/src/cli-installation.md
+++ b/book/src/cli-installation.md
@@ -34,25 +34,12 @@ yay -S numbat
 
 *... or any distribution where nix is installed.*
 
-### nix-env
-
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much. This is a lot of text. I don't really want to turn this into a NixOS tutorial. Can we decide on sentence / short paragraph + one command for installation?

giomf:
> I don't really want to turn this into a NixOS tutorial

I can understand that. I shortened the documentation. Have a look.
If its still to much we could even remove the second command.

sharkdp (approved):
Thank you!

Issue #237, opened 2023-11-08
Remove more styling when  `--pretty-print=never` is set


0xAdk:
I have recently been trying out numbat and how it compares to qalc.

My main use case is quick cli calculations
```sh
$ qalc -t '8 GiB to MiB'
8192 MiB
$ qalc -t
> 8 GiB to MiB

  8192 MiB

>
```
For numbat though I'm not able to get the same level of brevity.

The closest I can get to this with numbat is `--pretty-print=never`
```sh
$ numbat --pretty-print=never -e '8 GiB to MiB'
    = 8192 MiB    [DigitalInformation]

$ numbat -q --pretty-print=never
>>> 8 GiB to MiB

    = 8192 MiB    [DigitalInformation]

>>>
```

I was wondering if it would make sense to have formatting closer to qalc when `--pretty-print=never` is set, and if a PR would be welcome for these changes
```sh
$ numbat --pretty-print=never -e '8 GiB to MiB'
8192 MiB
$ numbat -q --pretty-print=never
>>> 8 GiB to MiB

  8192 MiB

>>>
```

Additionally It would be nice get close to `qalc -t` in use, so maybe `-e` could default to `--pretty-print=never`?
Or add an option `-t` as an alias for `--pretty-print=never`
```sh
$ numbat -e '8 GiB to MiB'
8192 MiB
$ # or
$ numbat -te '8 GiB to MiB'
8192 MiB
$ numbat -qt
>>> 8 GiB to MiB

  8192 MiB

>>>
```

0xAdk:
also, if `numbat -e` should default to no pretty printing, what about `echo 1 + 1 | numbat`

mhmdanas:
> Additionally It would be nice get close to qalc -t in use, so maybe -e could default to --pretty-print=never?

This was the default behavior in Insect, and I agree it makes sense not to pretty-print in non-interactive mode by default (even without passing `--pretty-print=never`). Adding a new command-line flag for this isn't necessary in my opinion.

> also, if numbat -e should default to no pretty printing, what about echo 1 + 1 | numbat

Both shouldn't pretty-print IMO.

sharkdp:
> I was wondering if it would make sense to have formatting closer to qalc when `--pretty-print=never` is set, and if a PR would be welcome for these changes

Yes :+1: 

> Additionally It would be nice get close to `qalc -t` in use, so maybe `-e` could default to `--pretty-print=never`?

Yes, good idea

[...]

PR #240, opened 2023-11-09
Default to `--pretty-print=never` for `-e` and stdin input

0xAdk:
see #237

Pretty sure this is a breaking change since I added a parameter to `InterpreterResult::to_markup()`

### DIFF
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -15,7 +15,7 @@ use numbat::markup as m;
 use numbat::module_importer::{BuiltinModuleImporter, ChainedImporter, FileSystemImporter};
 use numbat::pretty_print::PrettyPrint;
 use numbat::resolver::CodeSource;
-use numbat::{Context, InterpreterResult, NumbatError};
+use numbat::{Context, NumbatError};
 use numbat::{InterpreterSettings, NameResolutionError};
 
 use anyhow::{bail, Context as AnyhowContext, Result};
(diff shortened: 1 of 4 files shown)


sharkdp:

This looks great. I just merged #97. It would be great if you could rebase your branch. Let me know if you need help resolving the conflicts.

sharkdp:
> Pretty sure this is a breaking change since I added a parameter to `InterpreterResult::to_markup()`

I don't think the `numbat` crate has any users (yet). I also didn't really design it as a proper "public" library so far. But thank you for taking that into consideration!

0xAdk:
ok it's rebased, I also went ahead and changed https://github.com/sharkdp/numbat/pull/240/files#diff-723b19ccffb6afcf74a982f6dd78c112191d462e58d360c9fa49167157b1d278R136 if you don't mind.

numbat-cli/src/main.rs
```diff
@@ -133,27 +133,12 @@ impl Cli {
...
-        } else {
-            config.main.load_prelude
-        };
-        config.main.load_user_init = if args.no_prelude || args.no_init {
-            false
-        } else {
-            config.main.load_user_init
-        };
+        config.main.load_prelude &= !args.no_prelude;
+        config.main.load_user_init &= !(args.no_prelude || args.no_init);
```
sharkdp, line 137: Thank you. I knew there was a better way to write this, I was just too lazy to think yesterday :smile:

sharkdp:
Thank you very much!

PR #256, opened 2023-11-13
Add only existing modules to imported_modules list

### LINKED ISSUE #254: Attempting to use a non-existent module only results in error on first attempt
As far as I can see nothing bad happens when trying it again, however no error occurs on subsequent attempts to use the same non-existent module:

```
>>> use non_existent_module
error: while resolving imports in
  ┌─ <input:1>:1:5
  │
1 │ use non_existent_module
  │     ^^^^^^^^^^^^^^^^^^^ Unknown module

>>> use non_existent_module

>>> 
```

Bzero:
Should fix #254.

Modules were previously always added to the `imported_modules` list, this pull request changes this to only add them if the module can be read without error.

### DIFF
--- a/numbat/src/resolver.rs
+++ b/numbat/src/resolver.rs
@@ -98,8 +98,8 @@ impl Resolver {
             match statement {
                 Statement::ModuleImport(span, module_path) => {
                     if !self.imported_modules.contains(module_path) {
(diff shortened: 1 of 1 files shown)


sharkdp:

Changes are live on https://numbat.dev/

PR #265, opened 2023-11-21
Fix multiple errors spawning for the same token.

### LINKED ISSUE #260: Multiple error message for wrong assignments
We used to just have a single descriptive error message when attempting to do an assignment like `x = 3`:
```
>>> x = 3
error: while parsing
  ┌─ <input:1>:1:3
  │
1 │ x = 3
  │   ^ Trailing '=' sign. Use `let x = …` if you intended to define a new constant.
```

but now we get
```
>>> x = 3
error: while parsing
  ┌─ <input:1>:1:3
  │
1 │ x = 3
  │   ^ Trailing '=' sign. Use `let x = …` if you intended to define a new constant.

error: while parsing
  ┌─ <input:1>:1:3
  │
1 │ x = 3
  │   ^ Expected one of: number, identifier, parenthesized expression
```

presumably since #228 (FYI @irevoire)

irevoire:
Fix #260

Hey, basically, my error was that when we encountered a trailing error, we were still supposed to recover from it (by eating all the tokens until the end of the line), but I missed it in my first implementation.
So, this PR fixes it and adds a test for it.

### DIFF
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -251,10 +251,7 @@ impl<'a> Parser<'a> {
                 Ok(statement) => statements.push(statement),
                 Err(e) => {
                     errors.push(e);
-                    // skip all the tokens until we encounter a newline.
-                    while !matches!(self.peek().kind, TokenKind::Newline | TokenKind::Eof) {
-                        self.advance()
-                    }
+                    self.recover_from_error();
                 }
             }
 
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

This still has problems in edge cases where expression/statements span multiple lines, but I think that is okay:
```
>>> if 2 < 3 +
  then 4
  else 5 +
error: while parsing
  ┌─ <input:12>:1:11
  │  
1 │   if 2 < 3 +
  │ ╭──────────^
2 │ │   then 4
  │ ╰^ Expected one of: number, identifier, parenthesized expression

error: while parsing
  ┌─ <input:12>:2:3
  │
2 │   then 4
  │   ^^^^ Expected one of: number, identifier, parenthesized expression

error: while parsing
  ┌─ <input:12>:3:3
  │
3 │   else 5 +
  │   ^^^^ Expected one of: number, identifier, parenthesized expression

```

PR #273, opened 2023-12-04
Simplify binop parsing

irevoire:
Hey, just a little PR that tries to simplify adding new binop operations.

It's not _as worth it as I thought_ though, so feel free to close it if you think it actually makes the code harder to read.

Having two closures in the `parse_binop` method is a bit hard to use IMO, but overall, being able to use the same method for _7_ other parsers is interesting.

### DIFF
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -727,6 +727,31 @@ impl<'a> Parser<'a> {
         }
     }
 
+    /// Helper function to parse binary operations
+    /// - arg `op_symbol` specifiy the separator / symbol of your operation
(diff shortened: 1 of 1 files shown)


sharkdp:
I had the same thought when I reviewed your logical-operator PR — nice! If there are no performance regressions, I'm happy to accept this. I think it's definitely an improvement.

irevoire:
On my side, I didn't measure any performance regression:

```
Import prelude          time:   [16.242 ms 16.293 ms 16.345 ms]
                        change: [-0.3828% +0.1170% +0.5889%] (p = 0.63 > 0.05)
                        No change in performance detected.
```

Issue #163, opened 2023-09-20
Support for even more Unicode Symbols?


David-Talaga:
Currently using a non-alphanumeric unicode character results in an error. 

For example
```
>>> let Mass_H₂O=18.0057 grams
error: while parsing
  ┌─ <input:1>:1:11
  │
1 │ let Mass_H₂O=18.0057 grams
  │           ^ Unexpected character: '₂'
```
How much trouble would allowing a broader range of unicode characters cause? 

let ⟦H₃O⁺⟧ₓ=1.2 moles/liters

David-Talaga:
Thank you for the detailed answer. I discovered the part of Unicode that was compatible with numbat after my message. I find the Unicode subscript and superscript characters very useful for typesetting symbols, in general. My use case here is to show calculations with units for the chemistry classes I teach. We often have multiple symbols with the same dimensions. Snake case should be adequate, in principle for this purpose, but I find that a lot of my students find it confusing to learn yet another way to represent a symbol. “Extremely useful?” Probably not. “Kind of nice and cool?” Yeah. Numbat is fantastic, by the way. Thanks!-David TalagaSent from my iPhoneOn Sep 20, 2023, at 2:23 PM, David Peter ***@***.***> wrote:﻿
Thank you very much for the feedback.
We do support Unicode identifiers. We even use the official XID_START / XID_CONTINUE classes from Unicode annex #31 which was written for this exact purpose ("specifications for recommended defaults for the use of Unicode in the definitions of general-purpose identifiers").
But ⟦ and some other code points of your example are neither in XID_START, nor in XID_CONTINUE. The cited annex also has a Customization chapter, noting:

Each programming language standard has its own identifier syntax; different programming languages have different conventions for the use of certain characters such as $, @, #, and _ in identifiers. To extend such a syntax to cover the full behavior of a Unicode implementation, implementers may combine those specific rules with the syntax and properties provided here.
Each programming language can define its identifier syntax as relative to the Unicode identifier syntax, such as saying that identifiers are defined by the Unicode properties, with the addition of “$”. By addition or subtraction of a small set of language specific characters, a programming language standard can easily track a growing repertoire of Unicode characters in a compatible way.

In Numbat, we actually do this, and add certain additional characters like currency symbols ($, €, £, ¥), fractions (½, ¾, …), etc. But to be honest, I'm not sure if we can cover all use cases here. Do you think it would be extremely valuable to have something like ⟦H₃O⁺⟧ₓ as an identifier?

[...]

PR #277, opened 2023-12-13
Support subscript characters in identifiers

mhmdanas:
Partial fix for #163.

### DIFF
--- a/numbat/src/tokenizer.rs
+++ b/numbat/src/tokenizer.rs
@@ -171,6 +171,13 @@ fn is_other_allowed_identifier_char(c: char) -> bool {
     c == '%'
 }
 
+fn is_subscript_char(c: char) -> bool {
+    let c_u32 = c as u32;
+
+    // See https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts#Superscripts_and_subscripts_block
+    (0x2080..=0x209CF).contains(&c_u32)
+}
+
 fn is_identifier_start(c: char) -> bool {
     unicode_ident::is_xid_start(c)
         || is_numerical_fraction_char(c)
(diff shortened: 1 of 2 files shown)


numbat/src/tokenizer.rs
```diff
@@ -171,8 +171,16 @@ fn is_other_allowed_identifier_char(c: char) -> bool {
...
+fn is_subscript_char(c: char) -> bool {
+    let c_u32 = c as u32;
+
+    // See https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts#Superscripts_and_subscripts_block
+    (0x2080..=0x209CF).contains(&c_u32)
+}
+
 fn is_identifier_start(c: char) -> bool {
     unicode_ident::is_xid_start(c)
+        || is_subscript_char(c)
```
sharkdp, line 183: Do we want to allow subscripts at the beginning of an identifier?

mhmdanas: Good question. I personally think subscripts at the start of identifiers look a bit unaesthetic, but technically speaking I see no reason not to allow them there.

sharkdp: I think it's not a question of aesthetics, but rather about being future-proof. If we want to use subscript characters in the syntax at some point in the future, it might be better to only allow them as identifier-continue characters.

It's subtle and might never be required, but if we don't see a use case for identifiers that start with a subscript, I would only allow them as identifier-continue characters.

sharkdp:
Thank you. Can we maybe add a parser regression test and make sure that `Mass_H₂O` is parsed as an identifier?

mhmdanas:
Done, I hope I correctly understood what you mean by "regression test."

sharkdp:
Thank you!

PR #288, opened 2024-01-02
Add some units

mhmdanas:


numbat/modules/units/misc.nbt
```diff
@@ -100,6 +110,11 @@ unit footballfield: Area = 105 m × 68 m # Standard FIFA football pitch
 @url("https://en.wikipedia.org/wiki/Olympic-size_swimming_pool")
 unit swimmingpool: Volume = 50 m × 25 m × 2 m # Olympic-size swimming pool (FR3)
 
+@name("Rack unit")
+@url("https://en.wikipedia.org/wiki/Rack_unit")
+@aliases(rackunits, RU: short, U: short)
+unit rackunit: Length = 0.04445 meter
```
mhmdanas, line 116: I want to use `1.75 inch`, but this causes `cargo run --example=inspect` to fail (even when I have `use si::imperial`):
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/examples/inspect`
thread 'main' panicked at numbat/examples/inspect.rs:11:66:
called `Result::unwrap()` on an `Err` value: TypeCheckError(UnknownIdentifier(Span { start: SourceCodePositition { byte: 3266, line: 117, position: 30 }, end: SourceCodePositition { byte: 3270, line: 117, position: 34 }, code_source_id: 16 }, "inch", None))
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: core::result::Result<T,E>::unwrap
             at /builddir/rust-1.74.0/library/core/src/result.rs:1077:23
   4: inspect::main
             at ./numbat/examples/inspect.rs:11:19
   5: core::ops::function::FnOnce::call_once
             at /builddir/rust-1.74.0/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

eminence: Maybe there's a problem with a circular dependency:  imperial.nbt depends on misc.nbt, so adding a dependency from misc.nbt back to imperial might cause a problem?

sharkdp: > Maybe there's a problem with a circular dependency

That's it. We currently 'tolerate' cyclic dependencies:

https://github.com/sharkdp/numbat/blob/cb1c6be6dc02a7f44608b4774b20cf979606c051/numbat/src/resolver.rs#L263-L275

But we should probably show an error instead.

sharkdp: For this PR, let's go with what you have, @mhmdanas. Another option would be to split this module into multiple pieces. But we can do this in a follow-up.

sharkdp (approved):
Thank you

PR #294, opened 2024-01-22
Add desktop file and Numbat icons to assets

mhmdanas:


assets/numbat.desktop
```diff
@@ -0,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=Numbat
+GenericName=Scientific calculator
```
mhmdanas, line 4: Suggestions for a different generic name maybe?

book/src/cli-installation.md
```diff
@@ -126,3 +126,5 @@ If this variable is set during compilation, the specified path will be compiled
 
 In order to test that everything is working as intended, you can open `numbat` and type `let meter = 2`. The
 path in the error message should point to the specified location (and *not* to `<builtin>/…`).
+
+Additionally, it's a good idea to install `assets/numbat.desktop` and `assets/numbat-*x*.png` to their appropriate locations (typically `/usr/share/applications` and `/usr/share/icons` respectively) so that users can e.g. launch Numbat from GNOME's applications search.
```
mhmdanas, line 130: This paragraph (or at least the desktop file part) is specific to operating systems that deal with desktop files (e.g. desktop Linux and FreeBSD), so I should note that as it does not apply to Windows, macOS, and possibly others. Suggestions on how to rephrase?

assets/numbat-16x16.png
mhmdanas, line 1: I just copied these files from `numbat-wasm/www`, it would be nice to get rid of the duplication somehow.

book/src/cli-installation.md
eminence: Is there a way to install `numbat.desktop` to a user-local location that doesn't require root permissions?  Like in `~/.local/share/applications/` ? If so, let's document that here, too

mhmdanas: Installing it to `~/.local/share/applications` works, but this is a packaging guide so I don't think it's the appropriate place.

mhmdanas: Also, I need to check what's the difference between `/usr/share/icons` and `/usr/share/pixmaps`...

eminence: Ah, I hadn't noticed that your new text was added into the packaging section.  ok 👍

assets/numbat.desktop
sharkdp: Maybe check what similar applications use?

book/src/cli-installation.md
sharkdp: > This paragraph (or at least the desktop file part) is specific to operating systems that deal with desktop files (e.g. desktop Linux and FreeBSD), so I should note that as it does not apply to Windows, macOS, and possibly others. Suggestions on how to rephrase?

Maybe simply

```suggestion
If your OS uses `.desktop` files, you might also want to install `assets/numbat.desktop` and `assets/numbat-*x*.png` to their appropriate locations (typically `/usr/share/applications` and `/usr/share/icons` respectively) so that users can e.g. launch Numbat from GNOME's applications search.
```

assets/numbat-16x16.png
sharkdp: I'm okay with the duplication. If we really want to get rid of it, we could probably modify the build script for the web version to copy the ones in `/assets`.

assets/numbat.desktop
mhmdanas: Qalculate uses just "Calculator," we could go with that?

sharkdp (approved):
Thank you

PR #307, opened 2024-01-31
Improve link from units-definitions to dimension-definitions (Fixes #306)

### LINKED ISSUE #306: Doc Request: add example on how to define a new unit called "pages"
I need to do some math thinking about pages-per-hour, etc. How can I define a new unit called `page`, so I can do math like: `10 pages / (50 pages/hour)`

I think that it would be great if this example could be added to this documentation page: https://numbat.dev/doc/unit-definitions.html

DeflateAwning:
Fixes #306

### DIFF
--- a/book/src/unit-definitions.md
+++ b/book/src/unit-definitions.md
@@ -7,7 +7,7 @@ A new **base unit** can be defined by specifying the *physical dimension* it rep
 ``` numbat
 unit second: Time
 ```
-Here, `Time` denotes the physical dimension. To learn more, you can read the [corresponding chapter](dimension-definitions.md). But for now, we can just assume that they are already given.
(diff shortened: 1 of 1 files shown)


sharkdp:
Thanks. We should maybe move the "Custom dimensions" chapter to "Unit definitions" page and call it something like "Ad-hoc units". It's arguablby the most common use case for creating new units, so it should probably be explained on that page directly.

DeflateAwning:
I'd support that; will let someone more involved in the project make such a change though. I think that this change is a reasonable option until then.

I'd also support moving Custom Dimensions above Custom Units in the sidebar ordering.

sharkdp:
Closing in favor of #380

Issue #133, opened 2023-07-27
Pretty print 'long' units with long prefixes


sharkdp:
If we do not have a short alias available, we should use long prefixes with the 'canonical' unit name:
```
>>> @metric_prefixes
unit points

  @metric_prefixes
  unit points

>>> megapoints

  megapoints

    = 1 Mpoints
```

So here we should print "megapoints", not "Mpoints". We want to make sure not to print anything that can not be parsed again (see also "roundtrip" tests -> this example should be added there)

do-you-dare:
Hello! I'd like to work on this issue. Is it ok to add a new `CanonicalName` struct with the name and `AcceotsPrefix`?

sharkdp:
Sure, sounds like a good idea

PR #309, opened 2024-02-04
Pretty print 'long' units with long prefixes

do-you-dare:
This work is related to https://github.com/sharkdp/numbat/issues/133. Any feedbacks are welcome.

I've run the test case on the issue description and it seems to work as expected. `cargo test` passes with no errors and I've added a roundtrip check as required as well.

If you want, I can squash or change the commits to better fit the project as well

### DIFF
--- a/numbat/src/decorator.rs
+++ b/numbat/src/decorator.rs
@@ -49,6 +49,19 @@ pub fn get_canonical_unit_name(unit_name: &str, decorators: &[Decorator]) -> Str
     unit_name.into()
 }
 
+pub fn get_canonical_accepts_prefix(decorators: &[Decorator]) -> AcceptsPrefix {
+    for decorator in decorators {
+        if let Decorator::Aliases(aliases) = decorator {
+            for (_alias, accepts_prefix) in aliases {
+                if accepts_prefix.map(|ap| ap.short).unwrap_or(false) {
+                    return accepts_prefix.unwrap();
+                }
+            }
+        }
+    }
+    AcceptsPrefix::only_long()
+}
+
 pub fn name(decorators: &[Decorator]) -> Option<String> {
     for decorator in decorators {
         if let Decorator::Name(name) = decorator {
--- a/numbat/src/decorator.rs
+++ b/numbat/src/decorator.rs
@@ -1,4 +1,4 @@
-use crate::prefix_parser::AcceptsPrefix;
+use crate::{prefix_parser::AcceptsPrefix, unit::CanonicalName};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Decorator {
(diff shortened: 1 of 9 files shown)


numbat/src/unit.rs
```diff
@@ -315,7 +343,7 @@ impl Unit {
 
     #[cfg(test)]
     pub fn kelvin() -> Self {
-        Self::new_base("kelvin", "K")
+        Self::new_base("kelvin", "K", AcceptsPrefix::none())
```
sharkdp, line 346: It's just for tests, but I'd go with 
```suggestion
        Self::new_base("kelvin", "K", AcceptsPrefix::only_short())
```
here as well. Prefixes with `K` are definitely used, e.g. `µK`.

numbat/src/decorator.rs
```diff
@@ -49,6 +49,19 @@ pub fn get_canonical_unit_name(unit_name: &str, decorators: &[Decorator]) -> Str
     unit_name.into()
 }
 
+pub fn get_canonical_accepts_prefix(decorators: &[Decorator]) -> AcceptsPrefix {
+    for decorator in decorators {
+        if let Decorator::Aliases(aliases) = decorator {
+            for (_alias, accepts_prefix) in aliases {
```
mhmdanas, line 55:
```suggestion
            for (_, accepts_prefix) in aliases {
```

numbat/src/decorator.rs
```diff
@@ -49,6 +49,19 @@ pub fn get_canonical_unit_name(unit_name: &str, decorators: &[Decorator]) -> Str
...
+                if accepts_prefix.map(|ap| ap.short).unwrap_or(false) {
+                    return accepts_prefix.unwrap();
+                }
```
mhmdanas, line 58: I feel like this bit could be cleaned up somehow, but idk

numbat/src/unit.rs
```diff
@@ -209,12 +230,15 @@ impl Unit {
         self == &Self::scalar()
     }
 
-    pub fn new_base(name: &str, canonical_name: &str) -> Self {
+    pub fn new_base(name: &str, canonical_name: &str, accepts_prefix: AcceptsPrefix) -> Self {
```
mhmdanas, line 233: This could directly accept a `CanonicalName`, but it'd make the tests a tad more verbose so it might be fine as-is.

mhmdanas (changes requested):

numbat/src/decorator.rs
sharkdp: Hm, right. It could potentially also be combined with `get_canonical_unit_name` to reduce the code duplication?

The `if` statement could be written as
```rs
                match accepts_prefix {
                    &Some(ap) if ap.short => {
                        return ap;
                    }
                    _ => {}
                }
```
which is maybe a bit clearer, but quite a bit longer :man_shrugging:

numbat/src/unit.rs
do-you-dare: Thanks, it was a bit verbose, but not a big deal. Changed.

sharkdp (approved):
Thank you

sharkdp:
Oh, hold on. This isn't actually fixed completely.

Try typing `megabarn`, and it will result in `Mbarn`. I'm currently looking into it.

mhmdanas:
Thanks for the good work!

PR #312, opened 2024-02-05
Extended number formatting

### LINKED ISSUE #264: Extended number formatting
It would be great to extend the way we can print numbers using string interpolations. We could start simple and add support for precision-control:
```
print("{pi:.3}")
```

We should stay compatible with what Python/Rust do for numbers.

mhmdanas:
Fixes #264.

TODO:
- [x] Add format specifiers for strings (and possibly other types?)
- [x] Fix todos
- [x] Check that format specifiers are valid (currently anything invalid will crash Numbat)
- [x] Fix units not being printed in quantity interpolations with format specifiers
- [ ] Research possible choices other than `strfmt`
- [x] Add tests
- [x] Document
- [ ] Improve error messages

### DIFF
--- a/book/src/example-numbat_syntax.md
+++ b/book/src/example-numbat_syntax.md
@@ -103,6 +103,7 @@ print(2 kilowarhol)              # Print the value of an expression
 print("hello world")             # Print a message
 print("value of pi = {pi}")      # String interpolation
 print("sqrt(10) = {sqrt(10)}")   # Expressions in string interpolation
+print("value of π ≈ {π:.3}")     # Format specifiers
 
 assert(1 yard < 1 meter)         # Assertion
 
--- a/book/src/procedures.md
+++ b/book/src/procedures.md
@@ -26,6 +26,15 @@ let speed = 25 km/h
 print("Speed of the bicycle: {speed} ({speed -> mph})")
 ```
 
+Format specifiers are also supported in interpolations. For instance:
+
+```nbt
+print("{pi:0.2f}") // Prints "3.14"
+```
+
+For more information on supported format specifiers, please see
+[this page](https://doc.rust-lang.org/std/fmt/#formatting-parameters).
+
 ## Testing
 
 The `assert_eq` procedure can be used to test for (approximate) equality of two quantities.
--- a/examples/numbat_syntax.nbt
+++ b/examples/numbat_syntax.nbt
@@ -98,6 +98,7 @@ print(2 kilowarhol)              # Print the value of an expression
 print("hello world")             # Print a message
 print("value of pi = {pi}")      # String interpolation
 print("sqrt(10) = {sqrt(10)}")   # Expressions in string interpolation
+print("value of π ≈ {π:.3}")     # Format specifiers
 
 assert(1 yard < 1 meter)         # Assertion
 
--- a/numbat/src/ast.rs
+++ b/numbat/src/ast.rs
@@ -58,7 +58,11 @@ impl PrettyPrint for BinaryOperator {
 #[derive(Debug, Clone, PartialEq)]
 pub enum StringPart {
     Fixed(String),
-    Interpolation(Span, Box<Expression>),
+    Interpolation {
+        span: Span,
+        expr: Box<Expression>,
+        format_specifiers: Option<String>,
+    },
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -426,9 +430,15 @@ impl ReplaceSpans for StringPart {
     fn replace_spans(&self) -> Self {
         match self {
             f @ StringPart::Fixed(_) => f.clone(),
-            StringPart::Interpolation(_, expr) => {
-                StringPart::Interpolation(Span::dummy(), Box::new(expr.replace_spans()))
-            }
+            StringPart::Interpolation {
+                expr,
+                format_specifiers,
+                span: _,
+            } => StringPart::Interpolation {
+                span: Span::dummy(),
+                expr: Box::new(expr.replace_spans()),
+                format_specifiers: format_specifiers.clone(),
+            },
         }
     }
 }
--- a/numbat/src/bytecode_interpreter.rs
+++ b/numbat/src/bytecode_interpreter.rs
@@ -187,8 +187,16 @@ impl BytecodeInterpreter {
                             let index = self.vm.add_constant(Constant::String(s.clone()));
                             self.vm.add_op1(Op::LoadConstant, index)
                         }
-                        StringPart::Interpolation(_, expr) => {
+                        StringPart::Interpolation {
+                            expr,
+                            span: _,
+                            format_specifiers,
+                        } => {
                             self.compile_expression_with_simplify(expr)?;
+                            let index = self.vm.add_constant(Constant::FormatSpecifiers(
+                                format_specifiers.clone(),
+                            ));
+                            self.vm.add_op1(Op::LoadConstant, index)
                         }
                     }
                 }
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -1140,10 +1140,9 @@ impl<'a> Parser<'a> {
                 vec![StringPart::Fixed(strip_first_and_last(&token.lexeme))],
             ))
         } else if let Some(token) = self.match_exact(TokenKind::StringInterpolationStart) {
-            let mut parts = vec![StringPart::Fixed(strip_first_and_last(&token.lexeme))];
+            let mut parts = Vec::new();
 
-            let expr = self.expression()?;
-            parts.push(StringPart::Interpolation(expr.full_span(), Box::new(expr)));
+            self.interpolation(&mut parts, &token)?;
 
             let mut span_full_string = token.span;
             let mut has_end = false;
@@ -1154,10 +1153,7 @@ impl<'a> Parser<'a> {
                 span_full_string = span_full_string.extend(&inner_token.span);
                 match inner_token.kind {
                     TokenKind::StringInterpolationMiddle => {
-                        parts.push(StringPart::Fixed(strip_first_and_last(&inner_token.lexeme)));
-
-                        let expr = self.expression()?;
-                        parts.push(StringPart::Interpolation(expr.full_span(), Box::new(expr)));
+                        self.interpolation(&mut parts, &inner_token)?;
                     }
                     TokenKind::StringInterpolationEnd => {
                         parts.push(StringPart::Fixed(strip_first_and_last(&inner_token.lexeme)));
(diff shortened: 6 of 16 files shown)


numbat/src/ast.rs
```diff
@@ -57,7 +57,11 @@ impl PrettyPrint for BinaryOperator {
 #[derive(Debug, Clone, PartialEq)]
 pub enum StringPart {
     Fixed(String),
-    Interpolation(Span, Box<Expression>),
+    Interpolation {
+        span: Span,
+        expr: Box<Expression>,
+        precision: Option<u64>,
```
mhmdanas, line 63: Instead of a `precision` field, we could have a `specifiers: Vec<Specifier>` field here.

numbat/src/parser.rs
```diff
@@ -1152,7 +1152,28 @@ impl<'a> Parser<'a> {
...
+                        .unwrap(),
+                )
+            } else {
+                None
+            };
+            parts.push(StringPart::Interpolation {
+                span: expr.full_span(),
+                expr: Box::new(expr),
+                precision,
+            });
```
mhmdanas, line 1176: This code is very obviously horrible (also only handles `:0.3` and not `:.3`), but I need to think some more about how to improve it.

numbat/src/tokenizer.rs
```diff
@@ -539,6 +539,37 @@ impl Tokenizer {
...
+                    return Err(TokenizerError {
+                        kind: TokenizerErrorKind::UnterminatedString,
+                        span: Span {
+                            start: self.token_start,
+                            end: self.current,
+                            code_source_id: self.code_source_id,
+                        },
+                    });
+                }
+            }
```
mhmdanas, line 572: The lexer does little more than just have the whole specifier as a token, would be nice to make the tokens more specific. For instance, if we add support for `:0.3f`, maybe it should be tokenized to "colon, specifer(0.3f)" or maybe "colon, number(0.3), text("f")` or something like that? idk

Also should we allow whitespace between the specifier parts?

numbat/src/value.rs
```diff
@@ -28,6 +28,7 @@ pub enum Value {
     /// A DateTime with an associated offset used when pretty printing
     DateTime(chrono::DateTime<chrono::Utc>, chrono::FixedOffset),
     FunctionReference(FunctionReference),
+    FormatSpecifier(String),
```
sharkdp, line 31: Hm, this feels wrong(?). A format specifier can't live as a standalone value, right? You probably added it because you needed to implement `Constant::to_value`? We might need to rethink that.

mhmdanas: > Hm, this feels wrong(?). A format specifier can't live as a standalone value, right?

Agreed.

> You probably added it because you needed to implement `Constant::to_value`? We might need to rethink that.

Indeed.

numbat/src/typechecker.rs
```diff
@@ -1019,12 +1019,15 @@ impl TypeChecker {
...
-                        }
+                        StringPart::Interpolation {
+                            span,
+                            expr,
+                            specifiers,
+                        } => Ok(typed_ast::StringPart::Interpolation {
+                            span: *span,
+                            specifiers: specifiers.clone(),
+                            expr: Box::new(self.check_expression(expr)?),
+                        }),
```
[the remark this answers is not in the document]
mhmdanas: Would be nice to be able to format them to be honest, for example to add whitespace padding or similar things. Not sure if `strfmt` supports that though.

numbat/src/ast.rs
```diff
@@ -58,7 +58,11 @@ impl PrettyPrint for BinaryOperator {
 #[derive(Debug, Clone, PartialEq)]
 pub enum StringPart {
     Fixed(String),
-    Interpolation(Span, Box<Expression>),
+    Interpolation {
+        span: Span,
+        expr: Box<Expression>,
+        specifiers: String,
```
sharkdp, line 64: I think I would prefer `Option<String>` to clearly differentiate between an empty and a missing format specifier.

numbat/src/bytecode_interpreter.rs
```diff
@@ -181,18 +181,29 @@ impl BytecodeInterpreter {
...
                             let index = self.vm.add_constant(Constant::String(s.clone()));
                             self.vm.add_op1(Op::LoadConstant, index)
                         }
-                        StringPart::Interpolation(_, expr) => {
+                        StringPart::Interpolation {
+                            expr,
+                            span: _,
+                            specifiers,
+                        } => {
+                            len += 1;
```
sharkdp, line 197: I don't understand. If we add a `Constant::FormatSpecifier` unconditionally for every interpolation (which is good), why complicate the logic by incrementing "len" by 1 here *and* in the VM?

numbat/src/parser.rs
```diff
@@ -1157,7 +1167,17 @@ impl<'a> Parser<'a> {
...
+                        let specifiers = self
+                            .match_exact(TokenKind::StringInterpolationSpecifiers)
+                            .map(|token| token.lexeme.clone())
+                            .unwrap_or("".to_string());
+
+                        parts.push(StringPart::Interpolation {
+                            span: expr.full_span(),
+                            expr: Box::new(expr),
+                            specifiers,
+                        });
```
sharkdp, line 1180: Move this to a function to remove the code duplication?

numbat/src/tokenizer.rs
sharkdp: I think we decided against this (for now)?

numbat/src/typechecker.rs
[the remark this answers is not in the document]
sharkdp: Right, it should work great for strings already. And things that are not number or a string... we can maybe convert them to a string for now.

numbat/src/tokenizer.rs
mhmdanas: Yeah, we're just passing it all to `strfmt` anyway.

numbat/src/typechecker.rs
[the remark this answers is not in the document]
sharkdp: For dates, we could even think about supporting things like `"{now():%Y-%m-%d}"`. That shouldn't be too hard. We could use the type checker to distringuish those cases from everything else. And then call `format_datetime` (in the VM) instead of `strfmt::strfmt`.

numbat/src/value.rs
sharkdp: Let's maybe revert the changes in value.rs for now and implement `to_value` using `unreachable!(…)`? See if that works out?

numbat/src/bytecode_interpreter.rs
mhmdanas: No, you're totally right, I got rid of that.

numbat/src/value.rs
sharkdp: Oh right. Let's leave things as they are and clean this up later. Maybe a constant can be a `Value`. Or a `Constant(Value)` or similar.

book/src/example-numbat_syntax.md
```diff
@@ -103,6 +103,7 @@ print(2 kilowarhol)              # Print the value of an expression
 print("hello world")             # Print a message
 print("value of pi = {pi}")      # String interpolation
 print("sqrt(10) = {sqrt(10)}")   # Expressions in string interpolation
+print("value of e ~= {e:.3}")    # Format specifiers
```
sharkdp, line 106: Minor: maybe
```suggestion
print("value of e ≈ {e:.3}")     # Format specifiers
```

sharkdp: Also, let's use `pi` instead of `e`. I think `e` can be confusing here because it's also used in scientific notation.
```suggestion
print("value of π ≈ {π:.3}")     # Format specifiers
```

book/src/procedures.md
```diff
@@ -26,6 +26,15 @@ let speed = 25 km/h
 print("Speed of the bicycle: {speed} ({speed -> mph})")
 ```
 
+Rust-style format specifiers are also supported in interpolations. For instance:
```
sharkdp, line 29: Let's not put too much emphasis on rust maybe?
```suggestion
Format specifiers are also supported in interpolations. For instance:
```

book/src/procedures.md
```diff
@@ -26,6 +26,15 @@ let speed = 25 km/h
...
+
+```nbt
+print("{pi:0.2f}") // Prints "3.14"
+```
+
+For more information on supported format specifiers, please see
+https://doc.rust-lang.org/std/fmt/#formatting-parameters.
```
sharkdp, line 36:
```suggestion
[this page](https://doc.rust-lang.org/std/fmt/#formatting-parameters).
```

examples/numbat_syntax.nbt
```diff
@@ -98,6 +98,7 @@ print(2 kilowarhol)              # Print the value of an expression
 print("hello world")             # Print a message
 print("value of pi = {pi}")      # String interpolation
 print("sqrt(10) = {sqrt(10)}")   # Expressions in string interpolation
+print("value of e ~= {e:.3}")    # Format specifiers
```
sharkdp, line 101: Should have commented here. See above:
```suggestion
print("value of π ≈ {π:.3}")     # Format specifiers
```

numbat/src/vm.rs
```diff
@@ -216,6 +219,7 @@ impl Display for Constant {
             Constant::Boolean(val) => write!(f, "{}", val),
             Constant::String(val) => write!(f, "\"{}\"", val),
             Constant::FunctionReference(inner) => write!(f, "{}", inner),
+            Constant::FormatSpecifiers(_) => unreachable!(),
```
sharkdp, line 222: This will be called in debug mode: Try
```
numbat --no-prelude --debug -e '"{2:.2}"'
```

numbat/src/value.rs
```diff
@@ -85,6 +86,7 @@ impl std::fmt::Display for Value {
             Value::String(s) => write!(f, "\"{}\"", s),
             Value::DateTime(dt) => write!(f, "datetime(\"{}\")", dt),
             Value::FunctionReference(r) => write!(f, "{}", r),
+            Value::FormatSpecifiers(_) => unreachable!(),
```
sharkdp, line 89: Are you sure this isn't called anywhere? Maybe it's best just to implement it... e.g. for debugging purposes
```suggestion
            Value::FormatSpecifiers(_) => write(f, "<format specifier>"),
```

numbat/src/vm.rs
```diff
@@ -861,13 +865,60 @@ impl Vm {
...
+                                        strfmt::strfmt(&format!("{{value{}}}", specifiers), &vars)
+                                            .map_err(|err| match err {
+                                                strfmt::FmtError::Invalid(s) => {
+                                                    RuntimeError::InvalidFormatSpecifiers(s)
+                                                }
+                                                strfmt::FmtError::KeyError(_) => unreachable!(),
+                                                strfmt::FmtError::TypeError(s) => {
+                                                    RuntimeError::InvalidTypeForFormatSpecifiers(s)
+                                                }
+                                            })?;
```
sharkdp, line 893: Can we please extract this into a function to avoid the duplication?

sharkdp (changes requested):
Very nice! Just a couple of minor suggestions.

numbat/src/vm.rs
mhmdanas: Neat, didn't know about `--debug`. :D

numbat/src/vm.rs
mhmdanas: The best I could do was extracting the function passed to `map_err`. I couldn't get the `strfmt::strfmt(...)` portion to be shared between the two branches because `vars` has a different type in each branch (in one its values are of type `f64`, in the other of type `String`. I could maybe do some generic shenanigans or, god forbid, use a macro, but doesn't seem worth it to me.

sharkdp (approved):
Thank you!

mhmdanas:
Thanks!

sharkdp:
Changes are live on https://numbat.dev/

Issue #202, opened 2023-10-10
Conversion functions


sharkdp:
There are a lot of features that would naturally fit into the conversion operator (`->`) framework, if it was a little bit more powerful. One idea to support various use cases would be to register certain functions `f` that could be used as the (right hand side) target of a conversion.

When we would see a conversion like
```
x -> f
```
we would simplify call `f(x)`.

Things that could be solved with this approach:
* Hex/oct/binary representation of numbers. We would register functions like `fn hex(x: Scalar) -> String`, allowing us to call `42 to hex`
* HH:MM:SS representations. We would register `fn human_time(t: Time) -> String`, allowing us to call `5000 second -> human_time`
* Degree-minute-second representations for angles. We would register `fn DMS(a: Angle) -> String`, allowing something like `32.123° -> DMS == 32° 7′ 23″`.
* Degree/Fahrenheit conversions might also be more convenient with this functionality
* Approximate conversion to fractions (0.428571 => 3/7)
* DateTime -> Unix Timestamps
* scalar -> string (`chr`)

For °C and °F, what we really want is to be able to enter something like `30°C -> K`, `30°C -> °F` or `300 K -> °C`. This would require us to have overloads for the conversion functions. Also, it wouldn't solve the problem of defining how to parse `30°C`.

sharkdp:
This is now implemented. See https://numbat.dev/doc/conversion-functions.html for details.

Issue #181, opened 2023-09-26
Add support for parsing and displaying times in HH:MM:SS format?


sharkdp:
See previous discussions here:

- https://github.com/sharkdp/insect/issues/331
- https://github.com/sharkdp/insect/issues/263
- https://github.com/sharkdp/insect/issues/349

See also:

https://github.com/sharkdp/numbat/blob/a114a078b874a4078148541173be692b879adefc/examples/format_time.nbt#L1-L9

eminence:
A related enhancement would be to support queries like `now() + 5 days + 4 hours` and get an answer like `Nov 2nd 03:36:20`

mhmdanas:
@eminence that would involve datetime handling, and to be frank, I'm not sure if we want to open that _big_ can of worms.

eminence:
I agree that adding datetime handling would be hard (lots of input formats to parse, tricky bits about how to convert between timezones, how to control output formats, etc), but I actually think there is a minimal version of date/time handling that is not so difficult that we can't do it :)

septatrix:
I closed my related issue #274 in favor of discussing this here in one place as it mostly touches the same area though I also mentioned support for dates there.

My use case would be simple mathematical operations involving mostly dates (without time) though datetime support would also be very much appreciated. I also do not think that this is that big of a can of worms. Timezone handling with `chrono` in rust is quite doable. If that turns out too annoying though one could also implement this step by step. This could be split into the following:

- [ ] Support for alternative formatting of time durations (the original topic of this issue)
- [ ] Support for date arithmetic (Gregorian calendar)
- [ ] Support for datetime arithmetic (UTC or naive local time, i.e. timezone less time)
- [ ] Support for timezone arithmetic

sharkdp:
@septatrix I agree with everything you said. I also think we should do this step by step.

Let's focus on the first point: "alternative formatting of time durations". Do we want HH:MM:SS or something like `HH hours + MM minutes + SS seconds`. The latter has the advantage that it's valid Numbat syntax. Qalculate does something similar. For example: "3.54 hours" can be displayed as `3 h + 32 min + 24 s`.

Qyriad:
Hm, would `<number>:<number>…` be that bad? Colon is only used for type annotation right now, right? So a digit to the left of the colon would be unambiguous?

That being said, some kind of delimiter could allow for other kinds of datetime literals too.

sharkdp:
> Colon is only used for type annotation right now, right?

[...]

PR #327, opened 2024-02-09
Human-readable time durations

sharkdp:
This allows us to do things like

    >>> 1 sidereal_day // human

      = 23 hours + 56 minutes + 4.090500 seconds    [String]

Or

    >>> ten million seconds // human

      = 115 days + 17 hours + 46 minutes + 40 seconds    [String]

Once we implement #202, we can also write those as conversions (`1 siderreal_day -> human`)

related: #181


mhmdanas:
I know it'd be a bit more verbose, but it would be nice if the code output e.g. `1 hour` instead of `1 hours`.

PR #352, opened 2024-02-15
refactor: remove manual impl. of `Option::map`

hamirmahal:


eminence:
I think all of your PRs have been fixes suggested by clippy, right?  If so, that's great.   Please feel free to include multiple fixes in a single PR, you don't need a separate PR for each issue

hamirmahal:
Okay, I'll keep that in mind.

Issue #354, opened 2024-02-16
Command line arguments: Use both script and expression


sharkdp:
It makes sense to use both a `FILE` argument as well as a `-e <CODE>`/`--expression <CODE>`. However, I think we don't want the argument order to play a role, so I suggest we always execute all `FILE`s first, then all `CODE` parts, in sequence — and everything in the same context.

### Discussed in https://github.com/sharkdp/numbat/discussions/353

<div type='discussions-op-text'>

<sup>Originally posted by **ginabythebay** February 15, 2024</sup>
It would be handy to be able to reference a script from the command line that would, for instance, define a function, and then call that function, also from the same command line.

See also #258 </div>

PR #356, opened 2024-02-17
Enable using script and expression cli arguments at the same time

Bzero:
This pull request should resolve issue #354.

With this modification it becomes possible to use the `FILE` and `-e <CODE>` arguments in the same command, first executing `FILE` and then `CODE`, regardless of the argument order.

### DIFF
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -52,7 +52,6 @@ struct Args {
         short,
         long,
         value_name = "CODE",
-        conflicts_with = "file",
         action = clap::ArgAction::Append
     )]
     expression: Option<Vec<String>>,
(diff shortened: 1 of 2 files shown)


numbat-cli/src/main.rs
```diff
@@ -201,35 +200,45 @@ impl Cli {
...
+                let result = self.parse_and_evaluate(
+                    &code,
+                    code_source,
+                    ExecutionMode::Normal,
+                    self.config.pretty_print,
+                );
+
+                let result_status = match result {
+                    std::ops::ControlFlow::Continue(()) => Ok(()),
+                    std::ops::ControlFlow::Break(ExitStatus::Success) => Ok(()),
```
Bzero, line 232: I was not quite sure if this case should result in an error/warning for `FILE` if both `FILE` and `CODE` are given?

sharkdp: Good catch. Is `ControlFlow::Break(ExitStatus::Success)` even a possible? :-)

Bzero: I don't think so.  `Cli.parse_and_evaluate()` can only return `ControlFlow::Continue(())` or `ControlFlow::Break(ExitStatus::Error)`.

Actually I don't see  `ControlFlow::Break(ExitStatus::Success)`  being used anywhere, in which cases was initially intended to be used?

sharkdp: I guess I intended it to handle things like a hypothetical `exit(0)` call from a Numbat program. Let's remove it for now!

sharkdp (approved):
Thank you!

Issue #371, opened 2024-02-22
Deliver as a PWA on the web


bsidhom:
It would be great to make Numbat installable as a [Progressive Web App (PWA)](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/What_is_a_progressive_web_app). The key benefit of this is the ability to use it _offline_ on any device with a modern web browser.

Here's the minimum set of changes needed to get core PWA functionality:

- [ ] [Web manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest). See also [web.dev](https://web.dev/learn/pwa/web-app-manifest).
- [ ] Icon assets. These will be associated with the installed PWA, either on desktop or home screen.
- [ ] Service worker with full offline pre-caching of _all_ mandatory static assets. At this point, this is likely the index page, WASM bundle, and any JS/CSS/fonts used. The ECB data is dynamic and should probably be fetched in a [network-first fashion](https://web.dev/articles/offline-cookbook#network-falling-back-to-cache) and only fetched on demand. 

Some great follow-up features which could be done down the road but would really enhance the experience:
- [ ] A first-class, web-forward UI as described in [#218](https://github.com/sharkdp/numbat/issues/218#issuecomment-1957084709). Depending on how this materializes, it could be the default (or only) UI for all web users or we could have separate "desktop" and "mobile" options.
- [ ] Persistent, offline user configs. As I'm currently envisioning it, this would basically just consist of user-defined modules saved across sessions (_outside_ of URL params). For maximum compatibility, this would probably live in the [origin-private file system](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system). In truth, I would prefer full-fledged [FS access](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API) (see [the spec](https://wicg.github.io/file-system-access/)) so that users could literally share the same config files with their locally-installed CLI app, but it looks like Safari has no plans to support this (i.e., it looks like we'll be stuck with [OPFS](https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/)).

In any case, I think it would be a good exercise to start with the current web implementation and see how far we can go with the offline experience.

bsidhom:
NOTE: At the moment, I'm only considering the core Numbat calculator app as being "in scope" for the PWA, but I assume we could easily cache the docs for offline use as well.

bsidhom:
[...]

PR #372, opened 2024-02-22
Implement service worker and add PWA manifest

bsidhom:
This is the first step in making Numbat fully offline-capable and installable as a PWA (see #371). It's a bit brittle right now because it relies on static assets being manually tracked and added to the service worker lists.

There's also a hole where the Google Fonts are not explicitly cached by the service worker; however the browser should be able to fall back to the system fonts or browser-level cache in this case. I tested this on my machine and it rendered properly when I was fully offline.

I was not able to test the ECB exchange rate fetching due to the lack of CORS policy headers on the Numbat proxy, since I was running this at localhost.

### DIFF
--- a/numbat-wasm/www/index.js
+++ b/numbat-wasm/www/index.js
@@ -110,11 +110,31 @@ function setup() {
   });
 }
 
+// TODO: I would prefer to use const lambdas rather than `function` in order to
+// avoid unexpected `this` resolution. Not sure if there's a reason this is
+// currently being used along with `var`. Those "old school" techniques do
+// hoist, but this can easily be worked around by reordering declarations.
+async function registerSw() {
+  if ("serviceWorker" in navigator) {
+    try {
+      // NOTE: The service worker URL should be stable between releases or this
+      // could lead to unexpected behavior.
+      await navigator.serviceWorker.register("./service-worker.js");
+    } catch (error) {
+      // NOTE: We don't allow the error to propagate because the app should
+      // still work without offline capabilities.
+      console.error("Failed to register Service Worker:", error);
+    }
+  } else {
+    console.warn("Service Workers not available in this browser");
+  }
+}
+
 var numbat;
 var combined_input = "";
 
 async function main() {
-  await init();
+  await Promise.all([init(), registerSw()]);
 
   setup_panic_hook();
 
--- a/numbat-wasm/www/service-worker.js
+++ b/numbat-wasm/www/service-worker.js
@@ -1,0 +1,181 @@
+// The prefix is necessary if multiple web apps will be hosted at the same
+// origin, e.g., as would be done on GitHub Pages. This prevents separate apps
+// from clashing when it comes to origin-scoped resources such as Caches, OPFS,
+// etc.
+const APP_PREFIX = "numbat_";
+// This should be updated any time cached assets change. It allows the entire
+// set of precached files to be refreshed atomically on SW update.
+// TODO: It would be nice if we did this automatically whenever a commit affects
+// cached files. Note that this could also be a sequence number or similar.
+// Random UUIDs just make it easier to avoid collisions without coordination.
+const VERSION_UUID = "819e71eb-c3df-46b8-885e-5650feaea241";
+const CACHE_NAME = `${APP_PREFIX}${VERSION_UUID}`;
+
+// NOTE: If these lists get very large, they should be moved into Sets for
+// membership tests.
+
+// Files atomically pre-cached at installation/update.
+// NOTE: We do _not_ cache the service worker itself.
+// TODO: Consider self-hosting Google fonts and caching. Alternatively, we could
+// cache the remote fonts directly.
+const PRECACHED_FILES = [
+  "./",
+  "index.html",
+  "index.js",
+  "jquery.min.js",
+  "jquery.mousewheel-min.js",
+  "jquery.terminal.min.js",
+  "keyboardevent-key-polyfill.js",
+  "main.css",
+  "opensearch.xml",
+  "pkg/numbat_wasm.js",
+  "pkg/numbat_wasm_bg.wasm",
+  "pkg/package.json", // Does this need to be cached?
+  "terminal.css",
+];
+// Files cached on-demand (at first load), but not refreshed. This should mostly
+// include platform/device-specific resources such as icons which take up a lot
+// of space and not required across the board. These should also not typically
+// contain resources which would lead to semantic breakages if out of sync with
+// other files.
+const CACHED_FILES = [
+  "icon.svg",
+  "numbat-16x16.png",
+  "numbat-196x196.png",
+  "numbat-32x32.png",
+  "numbat.png",
+  "numbat.svg",
+  // TODO: Google Fonts? Self-host and pre-cache those?
+];
+// URLs for which we hit the internet on every request, only falling back to the
+// cache when offline.
+const DYNAMIC_URLS = ["https://numbat.dev/ecb-exchange-rates.php"];
+
+const getAbsoluteUrl = (file) => new URL(file, self.location.href).href;
+
+const PRECACHED_URLS = PRECACHED_FILES.map(getAbsoluteUrl);
+const CACHED_URLS = CACHED_FILES.map(getAbsoluteUrl);
+
+const handleInstall = async (e) => {
+  e.waitUntil(
+    (async () => {
+      // Atomically pre-cache common assets. This is technically racy because
+      // there's no coordination between SW version id and cached file
+      // contents (that would require something like asset hashes). However,
+      // the time window for such a race is pretty narrow and unlikely to
+      // arise in practice.
+      const cache = await caches.open(CACHE_NAME);
+      await cache.addAll(PRECACHED_FILES);
+      console.log(`[Service Worker] installed version ${VERSION_UUID}`);
+    })()
+  );
+};
+
+const handleActivate = async (e) => {
+  e.waitUntil(
+    (async () => {
+      // Drop old cache versions so storage doesn't explode. By doing this
+      // only at activation time, we ensure that older versions of the app
+      // continue to work until the new one takes over.
+      const keys = await caches.keys();
+      const deletions = [];
+      for (const key of keys) {
+        if (!key.startsWith(APP_PREFIX)) {
+          // Only touch caches related to this app, even if they're at the
+          // same origin.
+          continue;
+        }
+        if (key != CACHE_NAME) {
+          // Only keep the most recent cache version.
+          deletions.push(caches.delete(key));
+        }
+      }
+      // NOTE: It might be excessive to fail activation on cleanup of old
+      // caches, but something has clearly gone wrong in this case.
+      await Promise.all(deletions);
+    })()
+  );
+};
+
+const handleFetch = async (e) => {
+  const url = new URL(e.request.url).href;
+
+  const servePrecachedFile = async () => {
+    const cache = await caches.open(CACHE_NAME);
+    // NOTE: It may be possible for our caches to be evicted, in which case
+    // we'd need to re-fetch. On the other hand, this also breaks the atomic
+    // update semantics and could lead to mismatched asset versions and
+    // subtle breakage.
+
+    // NOTE: The await isn't strictly necessary here thanks to Promise
+    // coalescing, but this is less error-prone if we add more logic below.
+    // Same applies elsewhere.
+    return await cache.match(e.request);
+  };
+
+  const serveCachedFile = async () => {
+    const cache = await caches.open(CACHE_NAME);
+    const cachedResponse = await cache.match(e.request);
+    if (cachedResponse) {
+      // Serve the cached resource and don't attempt to refresh.
+      return cachedResponse;
+    }
+    // The requested icon was not in the cache. Fetch it and return it, but
+    // also add it to the cache for later.
+    let fetchedResponse;
+    try {
+      fetchedResponse = await fetch(e.request);
+    } catch (error) {
+      console.error(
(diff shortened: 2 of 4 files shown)


numbat-wasm/www/index.js
```diff
@@ -110,11 +110,31 @@ function setup() {
   });
 }
 
+// TODO: I would prefer to use const lambdas rather than `function` in order to
```
bsidhom, line 113: I don't plan to merge this comment as-is but to remove it or update this file based on feedback. I'm not confident enough in my JavaScript abilities to use `function` in general unless I'm implementing generators, where it's strictly required. 🙂

numbat-wasm/www/service-worker.js
```diff
@@ -0,0 +1,181 @@
+// The prefix is necessary if multiple web apps will be hosted at the same
+// origin, e.g., as would be done on GitHub Pages. This prevents separate apps
+// from clashing when it comes to origin-scoped resources such as Caches, OPFS,
+// etc.
+const APP_PREFIX = "numbat_";
+// This should be updated any time cached assets change. It allows the entire
+// set of precached files to be refreshed atomically on SW update.
+// TODO: It would be nice if we did this automatically whenever a commit affects
```
bsidhom, line 8: I guess this could be done in the deploy or build script, but it's not clear under what conditions exactly we should generate a new version. I think in that case we would want a single source of truth for all self-hosted artifacts (and remote URLs), such as a local JSON file. Then the local script could check the file hashes for changes and the service worker could fetch that JSON to determine the lists of files.

bsidhom:
I highly recommend [this article about the ServiceWorker lifecycle](https://web.dev/articles/service-worker-lifecycle) if you aren’t familiar with it.

numbat-wasm/www/index.js
```diff
@@ -110,11 +110,31 @@ function setup() {
...
+    console.warn("Service Workers not available in this browser");
+  }
+}
+
 var numbat;
 var combined_input = "";
 
 async function main() {
-  await init();
+  await Promise.all([init(), registerSw()]);
```
bsidhom, line 137: I went back and forth as to whether this should be done in-line at `init()` time or during the `load` event [as is often suggested](https://web.dev/articles/service-workers-registration). My initial thought was to `register()` and `claim()` _first_ in order to take advantage of ServiceWorker-cached artifacts. However, it looks like the SW "prefetching" can actually [fall back to the browser cache](https://web.dev/articles/service-workers-registration#testing_things_out) if we allow the page to load first. This makes me think we should maybe wait until both the `load` event has fired _and_ the Numbat terminal has initialized before registering. On the other hand, this is a pretty minor detail given how large the WASM payload is at the moment and the fact that most users will be on modern devices with a good connection.

bsidhom:
Note that the staged version is using my own Cloudflare Worker-based CORS proxy, but is otherwise identical to this PR.

sharkdp:

> This is staged at https://bsidhom.github.io/numbat/. You should be able to install the app and use it offline.

It works great!

Before I take a closer look: My web development knowledge is a bit outdated, and I have never worked with PWAs. I'm also not a frequent user. Can you think of any downsides or side effects for site visitors (that do not want to install the PWA) resulting from this change?

bsidhom:
The only downside that I see is extra retained storage space since the service worker auto-registers. As a result, it will pre-cache all static assets for the site and keep them "indefinitely". By "indefinitely", I mean the following:

- Cached files do not obey cache-control headers because we're _manually_ saving them via the [Caches API](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage).
- The current policy in the service worker is to retain these files until the next service worker update (signaled by any byte-level changes in the source code or its inputs (_not_ by changes to the referenced cache files). This is why I included the version UUID, which allows us to signal an explicit update to the SW itself _and_ serves as a cache key. We explicitly drop _all_ outdated caches on update.
- Each browser has its own storage policy for PWAs (actually, "offline" site storage in general; this includes CacheStorage, as well as IndexedDB, local storage, and [others](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria)). All offline data gets lumped into a bucket based on origin and browsers garbage collect based on LRU or other policies. Some browsers allow explicit user overrides to _retain_ this data, but otherwise everything is eligible for eviction. As far as I'm aware, Safari has the [strictest eviction policy](https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/) and storage limits, but IIRC, it's still around 500 MiB--1 GiB, and the eviction policy is waived (in a confusing way) for "installed" apps.

This should not affect loading time appreciably because any assets that were fetched by the page before the SW will hit the browser cache when fetched. I did leave a note about potentially _delaying_ SW registration to address this, but in my experience, the WASM loading is so slow that there's no noticeable difference in page load time.

Beyond the page load behavior, this should result in no differences when using the site. It _does_, however, give us flexibility and additional features when installed. For example, keyboard shortcuts that are normally intercepted by the browser can be handled by the PWA if it's "installed". This should help, for example, provide a full-featured terminal emulator experience if we end up switching to xterm.js. It also allows other stuff such as notifications, etc., which are less relevant to Numbat.

numbat-wasm/www/index.js
sharkdp: I am mainly concerned about further delaying the time until a user can type something into the terminal. Due to the large WASM size, this is already a long wait time, unfortunately — you're right. If `registerSw` is negligible in comparison, I'm okay with either way.

numbat-wasm/www/service-worker.js
sharkdp: I mean.. unless we change something completely unrelated like the docs, which are currently not part of this PWA, almost everything would require a VERSION_UUID change, right?

Could we update every time we use the `deploy.sh` script to push a new version/state to numbat.dev?

numbat-wasm/www/service-worker.js
```diff
@@ -0,0 +1,181 @@
...
+  "jquery.terminal.min.js",
+  "keyboardevent-key-polyfill.js",
+  "main.css",
+  "opensearch.xml",
+  "pkg/numbat_wasm.js",
+  "pkg/numbat_wasm_bg.wasm",
+  "pkg/package.json", // Does this need to be cached?
+  "terminal.css",
+];
+// Files cached on-demand (at first load), but not refreshed. This should mostly
```
sharkdp, line 36: Thanks for writing these comments, but I'm still not sure I understand the difference between PRECACHED and CACHED.

sharkdp:
> The only downside that I see is extra retained storage space since the service worker auto-registers. As a result, it will pre-cache all static assets for the site and keep them "indefinitely". By "indefinitely", I mean the following:
> 
>  Cached files do not obey cache-control headers because we're _manually_ saving them via the [Caches API](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage).

This scares me a bit, to be honest :-).

Probably completely unrelated question: why do I see two service workers? (Firefox)

image

numbat-wasm/www/service-worker.js
bsidhom: `PRECACHED` files are downloaded deterministically before the first request to the service worker, since we know they will be used. `CACHED` files are cached on-demand at first request but then not updated thereafter (until the next app release, via service worker update). This helps save space because certain user agents will download different subset of optional assets, such as icons. This would be more pronounced if we have more icon sizes, but I've tested it in different browsers and they do in fact request different icon sizes and formats. There's no way to tell in advance which subset a given client might want.

I wasn't sure how to name this. It's not the same as "dynamically cached", since that refers to files whose contents are expected to change with every request (currently the exchange rate data).

bsidhom:
I'm not sure when you first loaded the page, but it's possible that I updated the SW between you first and second load and the new (second) version hasn't transitioned to the "activated" state yet. In that case, it will wait until you've closed all browser tabs with the PWA or navigated those tabs to different pages. This ensures that the SW update happens atomically and while there are no live clients. I haven't updated it in a few days; IIRC, the last change I made was to update the CORS proxy. If you've had tabs open since then, that could explain it. Otherwise, it's possible one of those SWs is just a dummy. That behavior might vary by browser. See [this](https://web.dev/articles/service-worker-lifecycle) for an explanation of the lifecycle. It can be a bit difficult to understand exactly what's going on until you've played around with updates a bit, but I can try to explain any specific questions you have.

Can you be more specific about what scares you? If I didn't make it clear above, it's _not_ actually indefinite: old caches are cleared when we upload new SW versions (which is trivially done any time the SW contents change). Similarly, browsers will delete storage data based on their own policies or based on user-specified policies. This varies by browser, so I can't make any blanket statements. But if you can describe a pathological situation you're envisioning we can try to walk through what might happen in that case.

sharkdp:
> Can you be more specific about what scares you? If I didn't make it clear above, it's _not_ actually indefinite: old caches are cleared when we upload new SW versions (which is trivially done any time the SW contents change). Similarly, browsers will delete storage data based on their own policies or based on user-specified policies. This varies by browser, so I can't make any blanket statements. But if you can describe a pathological situation you're envisioning we can try to walk through what might happen in that case.

I'm mainly thinking about the future. What happens if we add new assets that we forgot to add to this list? What happens if we delete assets and they are still on this list? Do we need to enforce some synchronization? If so, how? Also, how can we automate the process of updating the UUID?

The main thing that concerns me is that we're also changing things for users that do not install the PWA. And for them, I don't see too many benefits (their browers would have cached the resources anyway — unless they're offline). But I do see potential problems.

But you know what... I'm also okay with trying this out. I don't want to be overly pessimistic. I really like your contribution and it's definitely cool to provide Numbat as a PWA for users who are into that.

bsidhom:
> I'm mainly thinking about the future. What happens if we add new assets that we forgot to add to this list? What happens if we delete assets and they are still on this list? Do we need to enforce some synchronization? If so, how? Also, how can we automate the process of updating the UUID?

Yeah, the UUID and asset automation are my main concerns as well. Those are both addressable, but I'm not sure how that's best done. Random UUID generation itself is trivial, but we'd need to somehow use templating or a JS build system to get it into the SW itself. It would also be wasteful to generate this on _every_ commit since it really only needs to change if any of the static resources have changed. That basically ties the UUID _and_ asset list problems into the same thing. We'd want some automation to basically scrape the set of built artifacts, hash all the contents, and then decide from there how/whether to update. In fact, since we're computing this deterministically from the set of assets, it probably makes more sense to just make the version identifier a hash value. Then the updater script can be stateless.

If you want to avoid bringing in new build-time dependencies (such as Node/NPM, etc.), which I'd prefer to do as well, then this can probably be done in something like a `build.rs` file (or just a separate Rust binary that we add). I would have included the skeleton for this stuff in this change, but I wasn't sure how to do it robustly. Specifically, I don't want to change all of the JS (even just in the service worker) into a template which gets filled out by the Rust build, mostly because this is pretty hostile to editors. The asset discovery and hashing is (in principle) straightforward. I guess we could split the file out into separate parts, each of which are valid JS: the base SW logic itself (without the version id or asset list) and the generated content (version hash and asset list). Then the build binary can generate the asset-related contents and form the output SW file by concatenating or something. Or the base SW file can make an explicit fetch against the generated data.

Note that even if we go the route of auto-generation, it still makes sense to give careful thought to which resources need to be dynamically cached (at runtime) vs precached. If we don't care about downloading unnecessary stuff, then we can just precache _all_ discovered files.

numbat-wasm/www/service-worker.js
bsidhom: I was originally thinking that we could put this into the deploy script, but the more I think about it the more sense it makes to use deterministic hashing of all built artifacts (i.e., anything that would get pushed via the deploy script). See my other comment in the main thread.

numbat-wasm/www/service-worker.js
bsidhom: Agreed. Especially in the case of auto-generation, this is simpler and more robust if everything is precached.

numbat-wasm/www/index.js
```diff
@@ -110,11 +110,31 @@ function setup() {
...
+    } catch (error) {
+      // NOTE: We don't allow the error to propagate because the app should
+      // still work without offline capabilities.
+      console.error("Failed to register Service Worker:", error);
+    }
+  } else {
+    console.warn("Service Workers not available in this browser");
+  }
+}
+
```
hamirmahal, line 132:
```suggestion
const registerSw = async () => {
  if ("serviceWorker" in navigator) {
    try {
      // NOTE: The service worker URL should be stable between releases or this
      // could lead to unexpected behavior.
      await navigator.serviceWorker.register("./service-worker.js");
    } catch (error) {
      // NOTE: We don't allow the error to propagate because the app should
      // still work without offline capabilities.
      console.error("Failed to register Service Worker:", error);
    }
  } else {
    console.warn("Service Workers are not available in this browser.");
  }
}
```

numbat-wasm/www/index.js
hamirmahal: I offered a suggestion in my review.

sharkdp:
> If you want to avoid bringing in new build-time dependencies (such as Node/NPM, etc.), which I'd prefer to do as well, then this can probably be done in something like a `build.rs` file (or just a separate Rust binary that we add). I would have included the skeleton for this stuff in this change, but I wasn't sure how to do it robustly. Specifically, I don't want to change all of the JS (even just in the service worker) into a template which gets filled out by the Rust build, mostly because this is pretty hostile to editors

Valid concern, yes.

How about the following: could we use a dummy hash like `<replace this>` for the development version inside the repository, and then simply search & replace that with the actual hash across all assets when we deploy to "production"? The disadvantage would be that developers would have to properly reset all cache when working on the web version(?), because the hash would never change.

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

PR #378, opened 2024-02-23
Store datetimes with fixed offset

### LINKED ISSUE #376: Support for RFC 3339 in target time zone
I'm not able to get times to print in RFC 3339 using the `format_datetime` function:

```
>>> get_local_timezone()

  get_local_timezone()

    = America/Los_Angeles    [String]

>>> format_datetime("%Y-%m-%dT%H:%M:%S%:z", local(now()))

  format_datetime("%Y-%m-%dT%H:%M:%S%:z", local(now()))

    = 2024-02-23T22:01:52+00:00    [String]

>>> now()

  now()

    = Fri, 23 Feb 2024 14:01:54 -0800    [DateTime]
```

According to [the chrono docs](https://docs.rs/chrono/latest/chrono/format/strftime/index.html#specifiers), `%z` and friends should give the _local offset_ relative to UTC. However, it looks like the TZ info is getting dropped before being sent to the formatter and it's always displaying in UTC. (I would also like to add formatters/parsers for the reference format to the prelude, but I'm not sure if others would find that useful.)

sharkdp:
This changes a couple of things regarding datetime handling:

- We don't store utc datetime and offset separately, but as a single DateTime<FixedOffset>. This not just fixes a bug, but also feels much cleaner overall.
- Do not ignore offsets in format_datetime, fixes #376
- Fix crashes for out-of-range values in `from_unixtime`
- datetime(…) will now return datetimes with the specified UTC offset (not the local UTC offset)
- from_unixtime(…) will now return datetimes in the local timezone, not in UTC

### DIFF
--- a/numbat/src/value.rs
+++ b/numbat/src/value.rs
@@ -26,7 +26,7 @@ pub enum Value {
     Boolean(bool),
     String(String),
     /// A DateTime with an associated offset used when pretty printing
-    DateTime(chrono::DateTime<chrono::Utc>, chrono::FixedOffset),
+    DateTime(chrono::DateTime<chrono::FixedOffset>),
     FunctionReference(FunctionReference),
 }
 
@@ -59,8 +59,8 @@ impl Value {
     }
 
     #[track_caller]
-    pub fn unsafe_as_datetime(&self) -> &chrono::DateTime<chrono::Utc> {
-        if let Value::DateTime(dt, _) = self {
+    pub fn unsafe_as_datetime(&self) -> &chrono::DateTime<chrono::FixedOffset> {
+        if let Value::DateTime(dt) = self {
             dt
         } else {
             panic!("Expected value to be a string");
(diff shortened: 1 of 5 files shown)


numbat/src/value.rs
```diff
@@ -26,7 +28,7 @@ pub enum Value {
     Boolean(bool),
     String(String),
     /// A DateTime with an associated offset used when pretty printing
-    DateTime(chrono::DateTime<chrono::Utc>, chrono::FixedOffset),
+    DateTime(chrono::DateTime<chrono::FixedOffset>),
```
sharkdp, line 31: @eminence Do you remember why you chose to store datetime in UTC + offset separately?

eminence: No, not really.   I think the initial version didn't support any datetime conversions, so storing a `DateTime<Utc>` was a natural fit, and it didn't evolve in the best way when Tz conversions were added.  I may have also been operating under [a potentially faulty] assumption that it would be easier to reason about datetime operations is everything was done in Utc (and then only converted to something else for printing). 

However skimming your new version here, it does look less complicated 👍

sharkdp: Thank you!

PR #386, opened 2024-03-01
feat: create directories for history if they DNE

### LINKED ISSUE #385: try to create directories for history if they don't already exist

hamirmahal:
fixes #385.

### DIFF
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -315,8 +315,6 @@ impl Cli {
         let result = self.repl_loop(&mut rl, interactive);
 
         if interactive {
-            // Create the prerequisite directories required to save history, if they don't exist already.
-            fs::create_dir_all(history_path.parent().unwrap())?;
             rl.save_history(&history_path).context(format!(
                 "Error while saving history to '{}'",
                 history_path.to_string_lossy()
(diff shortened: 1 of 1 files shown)


numbat-cli/src/main.rs
```diff
@@ -315,10 +315,21 @@ impl Cli {
         let result = self.repl_loop(&mut rl, interactive);
 
         if interactive {
-            rl.save_history(&history_path).context(format!(
-                "Error while saving history to '{}'",
-                history_path.to_string_lossy()
-            ))?;
+            // Try to save the history, and if it fails because prerequisite directories don't exist, create them.
+            if let Err(e) = rl.save_history(&history_path) {
```
sharkdp, line 319: Instead of checking whether it fails, we could always call `create_dir_all`.

hamirmahal: Let me know if the latest changes are more or less what you had in mind.

hamirmahal: Let me know if the latest changes are more or less what you had in mind.

sharkdp:
We actually have a call to `fs::create_dir` in `get_history_path`. The problem seems to be that you didn't have a `~/.local/share` folder at all? Is that possible?

I changed the fix to use `fs::create_dir_all` instead.

sharkdp (approved):

PR #370, opened 2024-02-21
Add '-i' option to enter interactive session after executing script or expression

Bzero:
Fixes #369.

PR #388, opened 2024-03-05
Skip serializing 'enter_repl'

Bzero:
This pull request fixes a small error that I made in #370 which accidentally made the 'enter_repl' variable accessible in the config file.


sharkdp (approved):
Thank you!

Issue #314, opened 2024-02-05
Non-ideal error message when both string and interpolation are unterminated


mhmdanas:
```console
>>> print("test {1)
error: while parsing
  ┌─ <input:5>:1:7
  │
1 │ print("test {1)
  │       ^^^^^^^ Unterminated string interpolation
```

(example taken from [this test](https://github.com/sharkdp/numbat/blob/8a4816048ea8d8f258d8f8d719d933423da66c08/numbat/src/parser.rs#L2461))

I would expect the error to either be "Unterminated string," or for the error span to point at just `{1`.

mhmdanas:
This issue was resolved by #389.

PR #389, opened 2024-03-07
Change error message for unterminated strings with interpolation.

Zengor:
I was exploring the codebase when I stumbled into what seemed to be the cause of #314.

The parser section handling interpolated strings would error out as UnterminatedStringInterpolation but it was effectively checking for unterminated strings. I couldn't really find a case where the tokenizer would be unable to catch an unterminated interpolation where it wasn't also a case of an unterminated string (let me know if there is one.)

For reference, here's the error message behavior with this PR:

image
image
---

There might be more edge cases with newlines. The error messages seems mostly fine as is, but the last one is maybe a bit weird:
image
image
image

### DIFF
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -192,8 +192,8 @@ pub enum ParseErrorKind {
     #[error("Expected 'else' in if-then-else condition")]
     ExpectedElse,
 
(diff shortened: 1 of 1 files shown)


sharkdp:
Thank you very much!

PR #396, opened 2024-03-12
Add "darcy" unit

mhmdanas:


numbat/modules/units/misc.nbt
```diff
@@ -115,6 +115,11 @@ unit swimmingpool: Volume = 50 m × 25 m × 2 m # Olympic-size swimming pool (FR
 @aliases(rackunits, RU: short, U: short)
 unit rackunit: Length = 0.04445 meter
 
+@metric_prefixes
+@aliases(darcys, darcies, D: short)
```
sharkdp, line 119: I would prefer if we keep the one-letter constant `D` free for now. Users can still set `unit D = darcy` (potentially with `@metric_prefixes`) if they want.
```suggestion
@aliases(darcys, darcies)
```

numbat/modules/units/misc.nbt
```diff
@@ -115,6 +115,11 @@ unit swimmingpool: Volume = 50 m × 25 m × 2 m # Olympic-size swimming pool (FR
...
+@url("https://en.wikipedia.org/wiki/Darcy_(unit)")
+unit darcy: Length^2 = (1 bar / 1 atmosphere) × micrometer²
```
sharkdp, line 121: I was wondering if we should introduce `dimension Permeability = Length^2`, but (1) there is also *electric* permeability with a different physical dimension, which might be confusing. And (2) this new dimension would show up as an alternative whenever the result is an area — which might also be confusing. So let's leave it at `Length^2` for now.

sharkdp (approved):
Thanks

Issue #390, opened 2024-03-08
'human' doesn't work great for negative durations


sharkdp:
```
>>> -3.5 days -> human

    = -4 days + 12 hours    [String]
```

MugheesQasim:
@sharkdp the reason for the wrong result of 'human' for negative duration is the usage of 'floor' command. 'floor' returns the largest integer less than or equal to the given number. In our case, we are getting '-4' when we provide '-3.5' as input as '-4' is the largest integer less than or equal to '-3.5'. 

Example case:
`human (-3.5 days)`

There can be two possible solutions for this issue:
1-) The one suggested in PR #400. To use ago instead in case of negative duration, the expected output would be:
`3 days 12 hours ago`
2-) Use abs() function to floor the duration to a positive integer first and then showing the output in this format:
`- (3 days + 12 hours)`

Let me know which one seems more reasonable or if you have anything else in mind, I would be happy to contribute.

sharkdp:
@MugheesQasim Did you also see https://github.com/sharkdp/numbat/pull/425?

MugheesQasim:
@sharkdp yes, the solution in PR #425 is using 'ago' in case of negative duration. If Numbat is ready to have two variants of human, `human_date` and `human_time`, then merging the PR will resolve this issue also.

sharkdp:
closed via #425

PR #400, opened 2024-03-15
Issue #390

DamienBradleyDSP:
For #390

Small rework of human.nbt

-1 day  ->    1 day ago
-2 hours -> 2 hours ago

### DIFF
--- a/examples/datetime_human_tests.nbt
+++ b/examples/datetime_human_tests.nbt
@@ -1,7 +1,13 @@
 assert((0 second         -> human) == "0 seconds")
+assert((0.1 second       -> human) == "0.100 seconds")
+assert((0.01 second      -> human) == "0.010 seconds")
+assert((0.001 second     -> human) == "0.001 seconds")
+
 assert((1 second         -> human) == "1 second")
 assert((5 second         -> human) == "5 seconds")
 assert((1.5 second       -> human) == "1.500 seconds")
+assert((2.55 second      -> human) == "2.550 seconds")
+assert((5.99999 second   -> human) == "5.999990 seconds")
 
 assert((60 seconds       -> human) == "1 minute")
 assert((73 seconds       -> human) == "1 minute + 13 seconds")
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you. From a quick glance, this seems to do more than just solve the problem with negative times. If so, can we please add regression tests for these other features/issues?

mhmdanas:
> I assume you mean just expand the number of unit tests that are already there to something more comprehensive?

Yes, that's generally good practice and what we expect people to do when contributing features to Numbat. :)

examples/datetime_human_tests.nbt
```diff
@@ -32,7 +51,46 @@ assert((1e12 days        -> human) == "1_000_000_000_000 days")
...
+assert((-10.5 decades    -> human) == "38350 days ago")
+
+assert((-25 ms           -> human) == "0.025 seconds ago")
+assert((-6400 ms         -> human) == "6.400 seconds ago")
+
+assert((-2500 µs         -> human) == "0.002500 seconds ago")
+assert((-40525 µs        -> human) == "0.040525 seconds ago")
+
+assert((-5147 ns         -> human) == "0.000005147 seconds ago")
+assert((-23s + 1234 ns   -> human) == "22.999998766 seconds ago")
```
sharkdp, line 96: The logic for "… ago" is relatively simple, right? In that case, we don't need to repeat all these tests here.

sharkdp: > True but does protect against changes if someone switches the logic in some way in the future, and with the test base's overall speed I'd say it doesn't hurt anyone to keep these in

Speed is not the matter, but humans are going to read those tests and wonder what they are for. I would really prefer if we strip down the negative-time-duration tests to one or two.

numbat/modules/datetime/human.nbt
```diff
@@ -3,34 +3,51 @@ use core::strings
...
-fn _human_readable_duration(time: Time, dt: DateTime, num_days: Scalar) -> String =
-    _human_join(_human_join(_human_join(_human_days(_human_num_days(time)), _human_hours(dt)), _human_minutes(dt)), _human_seconds(dt))
+fn _human_readable_duration(time: Time) -> String =
+    _human_join(_human_join(_human_join(_human_formatted_days(time), _human_formatted_hours(time)), _human_formatted_minutes(time)), _human_formatted_seconds(time))
 
 # Implementation details:
 # we skip hours/minutes/seconds for durations larger than 1000 days because:
 #   (a) we run into floating point precision problems at the nanosecond level at this point
 #   (b) for much larger numbers, we can't convert to DateTimes anymore
+fn human_time_check_overflow(time: Time) =
```
sharkdp, line 45: can we please `_`-prefix this as well?

sharkdp:
I am still not sure what this PR brings in addition to the "… ago" handling of negative durations. For now, I'm going to close this in favor of #425. But please feel free to comment here and we can reopen it.

PR #401, opened 2024-03-16
Added random function

mrkbac:
This PR add a `random()` function.

Example output:

```nbt
>>> 10 * random() // floor

  floor(10 × random())

    = 6 
```

Let me know if I should change anything!

### DIFF
--- a/numbat/src/ffi.rs
+++ b/numbat/src/ffi.rs
@@ -916,7 +916,7 @@ fn from_unixtime(args: &[Value]) -> Result<Value> {
 fn random(args: &[Value]) -> Result<Value> {
     assert!(args.len() == 0);
 
-    let output = fastrand::f64();
+    let output = rand::random::<f64>();
 
     Ok(Value::Quantity(Quantity::from_scalar(output)))
 }
--- a/numbat/src/ffi.rs
+++ b/numbat/src/ffi.rs
@@ -411,6 +411,15 @@ pub(crate) fn functions() -> &'static HashMap<String, ForeignFunction> {
             },
         );
 
+        m.insert(
+            "random".to_string(),
+            ForeignFunction {
+                name: "random".into(),
+                arity: 0..=0,
+                callable: Callable::Function(Box::new(random)),
+            },
+        );
+
         m
     })
 }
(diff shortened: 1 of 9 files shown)


numbat/tests/interpreter.rs
```diff
@@ -362,6 +362,7 @@ fn test_other_functions() {
     expect_output("floor(3.9)", "3");
     expect_output("round(3.9)", "4");
     expect_output("round(3.1)", "3");
+    expect_output("random() == random()", "false");
```
mhmdanas, line 365: This test could fail in theory (though very unlikely), so not sure we should have it.

mhmdanas: Actually this is probably so unlikely that we can probably disregard it entirely, /shrug

numbat/src/ffi.rs
```diff
@@ -903,3 +912,11 @@ fn from_unixtime(args: &[Value]) -> Result<Value> {
 
     Ok(Value::DateTime(dt))
 }
+
+fn random(args: &[Value]) -> Result<Value> {
+    assert!(args.len() == 0);
+
+    let output = fastrand::f64();
```
mhmdanas, line 919: We could use the `rand` crate instead (it's already a transitive dependency of Numbat).

mhmdanas (changes requested):

numbat/src/ffi.rs
mrkbac: I've changed it to `rand` and removed the test.

sharkdp (approved):
Thank you!

Bzero:
I am just a minute too late with my comment but I think it might be worth to think about if the `rand()` function should take `min` and `max` arguments. 

For many applications one will need to scale the default [0, 1) interval and to me `rand(a, b)` looks more convenient than `rand() * (b - a) + a`.

mhmdanas:
We don't have optional parameters at the moment so we might want to leave it until that's implemented.

sharkdp:
@Bzero Thank you for your comment. I had similar thoughts before merging. But I was hoping that we could now implement a function like rand(a, b) or randint(from, to) using that one source of randomness `random()`. If that turns out to be tricky (e.g. because of subtle biases) , let's add a more FFI functions or rethink this one.

Bzero:
Yes that sounds reasonable, it would probably make sense to implement these features in a module in numbat itself.

sharkdp:
> to implement these features in a module in numbat itself

Yes, absolutely. That's what I meant. Should have mentioned it. In general, I try to keep the FFI as small as possible.

PR #402, opened 2024-03-18
Add yuan as renminbi alias

P1n3appl3:
I tried to use [yuan](https://en.wikipedia.org/wiki/Yuan_(currency)) in a currency conversion and it was missing, I think this should be enough to add it. Yuan is both the singular and plural form, but I see there's also "yens" defined in numbat as an alias for the japanese yen so maybe I should add "yuans" for convenience?


sharkdp:
> Yuan is both the singular and plural form, but I see there's also "yens" define

In this case, let's only add `yuan` :+1:

sharkdp (approved):
Thank you

mhmdanas:
> Yuan is both the singular and plural form, but I see there's also "yens" defined in numbat as an alias for the japanese yen so maybe I should add "yuans" for convenience?

I removed the `yens` alias in d41a9943b739393c129c22e976d98b0e7badb6e0 for consistency.

mhmdanas:
Actually the same applies to South Korean won too, I think I'll research this a bit more later...

mhmdanas:
According to https://www.merriam-webster.com/dictionary/won, "wons" can also be the plural form, so let's keep it.

Issue #315, opened 2024-02-05
How should Numbat handle exchange currency fetching over long-lived sessions?


mhmdanas:
Relevant IRC discussion: https://github.com/sharkdp/numbat/issues/139#issuecomment-1795026339

After Numbat fetches currency exchange rates, they are stored for the duration of the session without ever being refreshed. This is problematic for long-running sessions because exchange rates fluctuate over time, and so in a Numbat session running over e.g. a week, the rates will become outdated.

One solution is to fetch them regularly, say once a day (maybe make the frequency configurable?). We could also make it configurable whether to prefetch them at that frequency or to only fetch them after they're outdated and a currency unit was used. This solution raises the question of how to handle network errors when refetching the exchange rates; we should probably let the user know about the failure lest they use outdated exchange rate information.

Thoughts?

bsidhom:
On the note of ECB data: is there a reason this request is proxied through a PHP server rather than queried directly? I came across this while looking into what it would take to make Numbat into a PWA with full offline support. The proxying doesn’t really change that at all, but it does make the deployment of Numbat less self-contained. Is this just for caching purposes in order to be nicer to the ECB service?

sharkdp:
> On the note of ECB data: is there a reason this request is proxied through a PHP server rather than queried directly?

Yes. I was running into CORS issues when attempting to let the client access the ECB server directly. I'm not an expert on this topic, so please let me know if you think there is a better solution. I'd love to get rid of the proxy step.

> caching purposes in order to be nicer to the ECB service?

That's another potential reason, but not the main one. And with the current number of Numbat users, I think we're good :-)

bsidhom:
Ah, unfortunately I’m not sure of any ways around that. I thought maybe service workers could offer an escape hatch (since they proxy all requests under the page scope), but they’re bound by the same cross-origin rules and only get access to opaque responses.

Dietr1ch:
> caching purposes in order to be nicer to the ECB service?

More than caching to save the ECB some load, I thought this was about saving latency.


Also, so far this talks about the "current" exchange rate, but rates are fluctuating and become historical too. On some transactions the rates on a particular recent day may be useful (`100usd@2025-01-25 to eur`), and that would mean that there's no need to toss old cached rates to keep the "current" ones fresh.
Having dated rates would also help offer a fallback when there's availability problems found while getting fresh rates (network, service down).

[...]

PR #403, opened 2024-03-18
Cache exchange rates

P1n3appl3:
I use lots of short invocations of numbat with `fetching-policy = "on-first-use"` which means it re-fetches the rates every time I open it to do a quick currency conversion.  The `on-startup` fetching policy makes it a bit better, but the delay is still noticeable if the first thing I do uses currency, and it just feels wasteful to download them every time when they're only updated daily. This change reuses the fetched rates for 24 hours to alleviate that delay.

This could make the rates up to a day out of date (if you happened to fetch them the first time right before the European Central Bank releases the next day's rates). In my mind this isn't hugely important as anyone who needs super up-to-date rates is probably using a different source in the first place, but the max age could be changed to something shorter to make it fetch more aggressively. If users want even more control, `max-age` could be specified as a config option in the `[exchange-rates]` section.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -921,6 +921,7 @@ name = "numbat-exchange-rates"
 version = "0.5.0"
 dependencies = [
  "attohttpc",
+ "dirs",
  "quick-xml",
 ]
 
--- a/numbat-exchange-rates/Cargo.toml
+++ b/numbat-exchange-rates/Cargo.toml
@@ -11,7 +11,8 @@ rust-version = "1.70"
 
 [dependencies]
 attohttpc = { version = "0.27.0", default-features = false, features = ["tls-rustls-webpki-roots"], optional = true }
+dirs = { version = "5", optional = true }
 quick-xml = "0.31.0"
 
 [features]
-fetch-exchangerates = ["dep:attohttpc"]
+fetch-exchangerates = ["dep:attohttpc", "dep:dirs"]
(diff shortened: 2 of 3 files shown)


sharkdp:
Thank you for looking into this. I believe we also have a ticket for this.

I only had a quick glance at the code. I think the caching should rather be moved to the application layer, not to the numbat-exchange-rates library. I realize that this is a bit harder to implement (as we need to serialize/deserialize), but I think that would be cleaner.

P1n3appl3:
When you say it should be in the "application layer" do you mean that `numbat-cli` should specify how writing and loading the cache file works? For that to work with on-demand currency loading there would need to be some sort of callback when the rates get fetched. I've roughed out that change in [this commit](https://github.com/sharkdp/numbat/commit/0db6cf8111ee9b5ff90070a35f0aea9675c3ec3f) but it a bit of a mess. Let me know if that's the direction you were thinking of and if so I'll try to clean it up and add it here.

Also I see #315 but that's kind of the opposite issue, was there another ticket you were thinking of?

sharkdp:
I don't fully understand why we need this setup with two closures and additional locks.

Could we add a (possibly atomic?) `filesystem_caching_enabled` flag to the `ExchangeRates` struct, set that according to the config when starting the application, and then internally have two branches for the caching/non-caching case?

sharkdp:
I'm closing this due to inactivity. Please feel free to comment in case it should be re-opened.

PR #404, opened 2024-03-18
Fix clippy lints

P1n3appl3:
I use a global rustfmt config which cause spurious changes when formatting in this repo. Adding an empty `.rustfmt.toml` in the root will force rustfmt to use the default options for line length/comment reflow/etc.

I also wanted to run clippy while making some other changes and cleaned up the existing lints so I could see any new ones I added. The only interesting one was `result_large_err` in the typechecker where the `Ok` variants were around 32 bytes and some of the `Err`s were a couple hundred bytes. I just allowed it for that whole module but I could go allow the individual functions it fired on, or use boxing to make the `Result` smaller.

### DIFF
--- a/numbat-cli/src/completer.rs
+++ b/numbat-cli/src/completer.rs
@@ -1,10 +1,7 @@
 use std::sync::{Arc, Mutex};
 
 use numbat::{unicode_input::UNICODE_INPUT, Context};
-use rustyline::{
-    self,
-    completion::{extract_word, Completer, Pair},
-};
+use rustyline::completion::{extract_word, Completer, Pair};
 
 pub struct NumbatCompleter {
     pub context: Arc<Mutex<Context>>,
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -22,8 +22,7 @@ use anyhow::{bail, Context as AnyhowContext, Result};
 use clap::Parser;
 use rustyline::config::Configurer;
 use rustyline::{
-    self, error::ReadlineError, history::DefaultHistory, Completer, Editor, Helper, Hinter,
-    Validator,
+    error::ReadlineError, history::DefaultHistory, Completer, Editor, Helper, Hinter, Validator,
 };
 use rustyline::{EventHandler, Highlighter, KeyCode, KeyEvent, Modifiers};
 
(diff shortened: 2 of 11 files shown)


sharkdp:
> I use a global rustfmt config which cause spurious changes when formatting in this repo. Adding an empty `.rustfmt.toml` in the root will force rustfmt to use the default options for line length/comment reflow/etc.

I don't have this in any of my Rust projects, and I'd rather avoid it if possible.

P1n3appl3:
Removed it, would you be ok with adding it as an entry in `.gitignore` so I can have it locally without worrying about committing it? Without it I'd accidentally run `cargo fmt` and end up introducing a bunch of spurious changes that line-wrap comments because I use [`wrap_comments = true`](https://rust-lang.github.io/rustfmt#wrap_comments)

mhmdanas:
No opinion on whether to have `.rustfmt.toml` in our `.gitignore`, but you can put it in a global `.gitignore` file that applies to all repos on your system: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

P1n3appl3:
Good point, I'll just add it to my global ignore

mhmdanas:
Closing since people can use a global `.gitignore`.

mhmdanas:
Ah, sorry, I didn't pay attention and thought this PR just added `.rustfmt.toml` to `.gitignore`...

mhmdanas (approved):
LGTM

sharkdp:
> I also wanted to run clippy while making some other changes and cleaned up the existing lints so I could see any new ones I added. The only interesting one was `result_large_err` in the typechecker where the `Ok` variants were around 32 bytes and some of the `Err`s were a couple hundred bytes. I just allowed it for that whole module but I could go allow the individual functions it fired on, or use boxing to make the `Result` smaller.

I saw those clippy warnings before and wasn't yet sure how to handle these. I think we should more carefully look into whether or not those appear in performance critical code. And if yes, whether or not boxing the error type would be beneficial. Otherwise, I'm fine with `allow`ing it.

P1n3appl3:
Not sure why the doc checks failed, re-running `book/build.sh` doesn't introduce any changes. I rebased onto master in case it had to do with being out of date.

hamirmahal (commented):
It's possible to add an optional `cargo clippy` check, by making all other checks required, like `CICD / all-jobs` in https://github.com/sharkdp/bat/pull/2913.

Even if a new `cargo clippy` check is optional, it still might encourage contributors to commit fewer linting errors into the codebase when they see a failing clippy check for their PR. `clippy` usually says how to fix any errors, too.

image

hamirmahal:
The previous comment is in reference to https://github.com/sharkdp/numbat/pull/355#issuecomment-1951414596, but I thought I'd bring it up here, since it's relevant again.

PR #406, opened 2024-03-22
Tokenize three periods as ellipsis (not just unicode character …)

mhmdanas:


sharkdp:
Thank you.

Note that you can also do `\dots<tab>` in the CLI version.

Issue #144, opened 2023-09-08
Support more mathematical functions


sharkdp:
Basic math
- [ ] erf
- [ ] erfc (complementary error function 1 - erf, useful for better accuracy)
- [ ] beta function
- [ ] zeta function?

Float
- [x] is_infinite
- [x] is_nan
- [x] frac (fractional part)
- [ ] int (integer part)
- [ ] sgn
- [ ] …

Number theory

- [x] gcd
- [x] lcm

Combinatorics:
- [ ] nPr/perm: number of ways to arrange r objects out of a total of n objects
- [ ] nCr/comb: number of ways to choose r objects out of a total of n objects without regard to order
- [ ] idiv (integer division)
- [x] binomial coefficient
- [x] @rben01: factorial (there is the ! postfix operator but not a standalone function (that I could find))
- [ ] @rben01: various sequences like Fibonacci, Lucas, Catalan

Bitwise?
- [ ] and, or, …
- [ ] shl, shr

Simple statistics
- [x] median
- [x] stddev
- [ ] geomean (see interesting observations regarding geomean here: http://typesatwork.imm.dtu.dk/material/TaW_Paper_TypesAtWork_Kennedy.pdf)

Statistics
- [x] binomial distribution etc
- [ ] @rben01: z-score calculator and inverse z-score calculator (probability → n standard devs and n standard devs → probability)

Numerics:
- [x] root-finding algorithms
- [x] numerical diff./integration, see #59 
- [ ] numerical maximum/minimum?

Cryptography
- [ ] checksum / hash functions?

Benjamin-L:
I'm interested in working on bit-wise operations

sharkdp:
> I'm interested in working on bit-wise operations

Sounds great. Note that we do not have proper integers at the moment. We basically just show floating point numbers as integers if they "look like one". I would definitely like to add support for different numeric types (floats, big/arbitrary-precision integers, potentially rationals, potentially complex numbers).

Just as a hint. We could still look into getting bit-wise operations working with what we have at the moment.

PR #407, opened 2024-03-23
Add NaN, inf, is_nan and is_infinite

Bzero:
This pull request introduces the 'NaN' keyword such that such that `NaN` values may be used directly in numbat.

As far as I am aware, `NaN` can currently only be accessed by calling an FFI function that returns `NaN` (e.g. `ln(-1)` or `sqrt(-1))` but there is no direct way to declare a `NaN` value. With this change it becomes possible to use `NaN` directly, e.g:

```
>>> let foo = NaN

  let foo: Scalar = NaN
```

To check if a value is `NaN,` a `is_nan()` function is added (see also #144).

I stumbled over this limitation when trying to implement a function with a limited domain that can handle input outside of the domain in a meaningful way in numbat. Using the `NaN` keyword it is possible to write:

```
fn triangular_number(n: Scalar) -> Scalar=
	if n>=0
	then n*(n+1)/2
	else NaN
```


sharkdp:
Thank you. I thought about this in the past, but didn't get around to implementing it — thanks.

I'm not sure if we really need NaN as a special token in the language, though? Could we perhaps define it in the prelude? For now, we could use something like `ln(-1)` and later we could maybe replace that by something cleaner. Like a `from_bits(0xFFC00000)` call, where `from_bits` would act like a bitlevel cast from int=>float.




> I stumbled over this limitation when trying to implement a function with a limited domain that can handle input outside of the domain in a meaningful way in numbat.

Note that you can use `error("…")` for those cases:
```
fn triangular_number(n: Scalar) -> Scalar=
	if n>=0
	then n*(n+1)/2
	else error("n needs to be non-negative")
```

Bzero:
> I'm not sure if we really need NaN as a special token in the language, though? Could we perhaps define it in the prelude? 

I figured having 'NaN' as a reserved keyword would make it available even with the '--no-prelude' option (just as any other value) and it would protect it from being user-redefined which could lead to unexpected behavior. On the other hand also `π` and all other constants can be overwritten so I am happy to turn it into a constant defined in prelude.

> For now, we could use something like `ln(-1)` and later we could maybe replace that by something cleaner. Like a `from_bits(0xFFC00000)` call, where `from_bits` would act like a bitlevel cast from int=>float.

Sounds good to me!

Bzero:
> Note that you can use `error("…")` for those cases:
> 
> ```
> fn triangular_number(n: Scalar) -> Scalar=
> 	if n>=0
> 	then n*(n+1)/2
> 	else error("n needs to be non-negative")
> ```

If I understood it right that currently only works for functions with `Scalar` return values, right?

mhmdanas:
Do we also want to add "constants" for positive and negative infinity as well?

Bzero:
I think that could be handy. Positive infinity `inf` should be enough though as one can just use `-inf` for negative infinity.

On that occasion it would probably make sense to add `f64::is_finite()` to the FFI.

sharkdp:
> I figured having 'NaN' as a reserved keyword would make it available even with the '--no-prelude' option (just as any other value) and it would protect it from being user-redefined which could lead to unexpected behavior.

I thought about this some more and I think I actually agree with you on this. Would you mind reverting that change again and making NaN/inf a proper builtin? (Sorry)

> If I understood it correctly that currently only works for functions with `Scalar` return values, right?

Yes. I thought about adding something like
```
fn error_d<D>(x: D, msg: String) -> D = error(msg) * unit_of(x)
```
which could be used like `4.3 km + error_d(m, "message")`. That would only be a workaround solution, though. The proper solution would be to extend the type system to allow for `!` return types (functions that never return). That might actually not be too hard, since the typing rules are quite simple (everything combined with `!` yields `!`)

Bzero:
> I thought about this some more and I think I actually agree with you on this. Would you mind reverting that change again and making NaN/inf a proper builtin? (Sorry)

Of course not! That's on me, I was changing it a bit too quickly before waiting for your reply.

numbat/src/parser.rs
```diff
@@ -1724,6 +1730,12 @@ mod tests {
         should_fail(&["0x1.2", "0b1.0", "0o1.0", "0x.1", "0b.0", "0o.1"]);
     }
 
+    #[test]
```
Bzero, line 1733: I had some trouble writing a proper test for `NaN` since `NAN == NAN` is `false`.

mhmdanas: You could add a test for this somewhere in the interpreter tests, and check that `NaN == NaN` evaluates to false in Numbat.

Bzero: There is a [test](https://github.com/sharkdp/numbat/blob/32c3e075f14d5682addd45208d5e88a1b8b617fd/numbat/tests/interpreter.rs#L112)  in the interpreter tests already which works fine but I thought it might be good to have a parser test too?

mhmdanas: Oh I missed that, but we can have something like `NaN != NaN` as a test too I suppose.

sharkdp: I would think we're good with the `inf` test here. The code for NaN is the same.

sharkdp (approved):

Issue #413, opened 2024-04-05
Random distribution sampling


Bzero:
Following up on the discussion in #401 I started to implement functions for random sampling of various distributions in Numbat.

I thought the following distributions would be nice to have: 

- [Continuous uniform distribution](https://en.wikipedia.org/wiki/Continuous_uniform_distribution)
- [Integer uniform distribution](https://en.wikipedia.org/wiki/Discrete_uniform_distribution)
- [Normal distribution](https://en.wikipedia.org/wiki/Normal_distribution)
- [Bernoulli distribution](https://en.wikipedia.org/wiki/Bernoulli_distribution)
- [Binomial distribution](https://en.wikipedia.org/wiki/Binomial_distribution)
- [Geometric distribution](https://en.wikipedia.org/wiki/Geometric_distribution)
- [Poisson distribution](https://en.wikipedia.org/wiki/Poisson_distribution)
- [Exponential distribution](https://en.wikipedia.org/wiki/Exponential_distribution)
- [Log normal distribution](https://en.wikipedia.org/wiki/Log-normal_distribution)
- [Pareto distribution](https://en.wikipedia.org/wiki/Pareto_distribution)

Does this selection sound reasonable or would some of these be considered out of scope?

sharkdp:
> Does this selection sound reasonable or would some of these be considered out of scope?

I'd be happy to support all of these, but it would be great if you could open a PR for just one or two of them at first.. so we can see where and how to integrate those without duplicating a lot of code. Ideally, I would like most of the code to be in Numbat (instead of having all of them as custom FFI functions implemented in Rust). But I understand that the language is probably not ready for this yet?

Bzero:
> I'd be happy to support all of these, but it would be great if you could open a PR for just one or two of them at first.. so we can see where and how to integrate those without duplicating a lot of code.

Sure, that sounds sensible.

> But I understand that the language is probably not ready for this yet?

I think numbat is perfectly suitable for most of them, out of those I already had a look at none was particularly tough to implement in numbat itself. The ones which have an analytical cumulative distribution function can be easily sampled using [inversion sampling](https://en.wikipedia.org/wiki/Inverse_transform_sampling) and I think for many of the others similar transforms exist.

sharkdp:
> I think numbat is perfectly suitable for most of them

Awesome!

sharkdp:
Should we close this, and open a new ticket if we feel something is still missing?

Issue #37, opened 2022-09-28
Zero, NaN, inf should be polymorphic


sharkdp:
`3 meter + 0` should work. Similarly: something like `hypot(2m, 0)` should work.

PR #416, opened 2024-04-27
Random distribution sampling 

Bzero:
This PR adds the function `rand_uniform(a, b)` which samples the continuous uniform distribution using inversion sampling. The function takes care of argument reversal.

Partially addresses #413.

### DIFF
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -4,6 +4,9 @@ use core::scalar
 # description: Uniformly samples the interval [0,1).
 fn random() -> Scalar
 
+# name: Continuous uniform distribution sampling
+# url: https://en.wikipedia.org/wiki/Continuous_uniform_distribution
+# description: Uniformly samples the interval [a,b) if a<=b or [b,a) if b<a using inversion sampling.
 fn rand_uniform<T>(a: T, b: T) -> T =
     if a <= b
     then random() * (b - a) + a
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -1,5 +1,7 @@
 use core::scalar
 use core::quantities
+use core::error
+use core::functions
 use math::functions
 
 # name: Standard uniform distribution sampling
@@ -31,7 +33,7 @@ fn rand_bernoulli(p: Scalar) -> Scalar =
     then (if random() < p
         then 1
         else 0)
-    else error("p must be a probability (0 <= p <= 1).")
+    else error("Argument p must be a probability (0 <= p <= 1).")
 
 # name: Binomial distribution sampling
 # url: https://en.wikipedia.org/wiki/Binomial_distribution
@@ -42,7 +44,7 @@ fn rand_binom(n: Scalar, p: Scalar) -> Scalar =
     then rand_binom(n-1, p) + rand_bernoulli(p)
     else if n == 0
     then 0
-    else error("n must be a positive integer.")
+    else error("Argument n must be a positive integer.")
 
 # name: Normal distribution sampling
 # url: https://en.wikipedia.org/wiki/Normal_distribution
@@ -57,7 +59,7 @@ fn rand_norm<T>(μ: T, σ: T) -> T =
 fn rand_geom(p: Scalar) -> Scalar =
     if p>=0 && p<=1
     then ceil( ln(1-random()) / ln(1-p) )
-    else error("p must be a probability (0 <= p <= 1).")
+    else error("Argument p must be a probability (0 <= p <= 1).")
 
 # A helper function for rand_poisson, counts how many samples of the standard uniform distribution need to be multiplied to fall below lim.
 fn _poisson(lim: Scalar, prod: Scalar) -> Scalar =
@@ -73,7 +75,7 @@ fn _poisson(lim: Scalar, prod: Scalar) -> Scalar =
 fn rand_poisson(λ: Scalar) -> Scalar =
     if λ >= 0
     then _poisson(exp(-λ), 1)
-    else error("λ must not be negative.")
+    else error("Argument λ must not be negative.")
 
 # name: Exponential distribution sampling
 # url: https://en.wikipedia.org/wiki/Exponential_distribution
@@ -82,7 +84,7 @@ fn rand_poisson(λ: Scalar) -> Scalar =
 fn rand_expon<T>(λ: T) -> 1/T =
     if value_of(λ) > 0
     then - ln(1-random()) / λ
-    else NaN * unit_of(1/λ)
+    else error("Argument λ must be positive.")
 
 # name: Log-normal distribution sampling
 # url: https://en.wikipedia.org/wiki/Log-normal_distribution
@@ -97,4 +99,4 @@ fn rand_lognorm(μ: Scalar, σ: Scalar) -> Scalar =
 fn rand_pareto<T>(α: Scalar, min: T) -> T =
     if value_of(min) > 0 && α > 0
     then min / ((1-random())^(1/α))
-    else NaN * unit_of(min)
+    else error("Both arguments α and min must be positive.")
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -22,6 +22,17 @@ fn rand_int<T>(a: T, b: T) -> T =
     then floor( random() * (floor(b) - ceil(a) + 1 unit_of(a)) ) + ceil(a)
     else floor( random() * (floor(a) - ceil(b) + 1 unit_of(b)) ) + ceil(b)
 
+# name: Bernoulli distribution sampling
+# url: https://en.wikipedia.org/wiki/Bernoulli_distribution
+# description: Samples a Bernoulli random variable, that is, 1 with probability p, 0 with probability 1-p.
+#              Parameter p must be a probability (0 <= p <= 1).
+fn rand_bernoulli(p: Scalar) -> Scalar =
+    if p>=0 && p<=1
+    then (if random() < p
+        then 1
+        else 0)
+    else error("p must be a probability (0 <= p <= 1).")
+
 # name: Normal distribution sampling
 # url: https://en.wikipedia.org/wiki/Normal_distribution
 # description: Samples a normal distribution with mean μ and standard deviation σ using the Box-Muller transform.
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -27,3 +27,12 @@ fn rand_int<T>(a: T, b: T) -> T =
 # description: Samples a normal distribution with mean μ and standard deviation σ using the Box-Muller transform.
 fn rand_norm<T>(μ: T, σ: T) -> T =
     μ + sqrt(-2 σ² × ln(random())) × sin(2π × random())
+
+# name: Geometric distribution sampling
+# url: https://en.wikipedia.org/wiki/Geometric_distribution
+# description: Samples a gemoetric distribution(the distribution of the number of Bernoulli trials with probability p needed to get one success) by inversion sampling.
+#              Parameter p must be a probability (0 <= p <= 1).
+fn rand_geom(p: Scalar) -> Scalar =
+    if p>=0 && p<=1
+    then ceil( ln(1-random()) / ln(1-p) )
+    else error("p must be a probability (0 <= p <= 1).")
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -45,3 +45,9 @@ fn rand_expon<T>(λ: T) -> 1/T =
     if value_of(λ) > 0
     then - ln(1-random()) / λ
     else NaN * unit_of(1/λ)
+
+# name: Log-normal distribution sampling
+# url: https://en.wikipedia.org/wiki/Log-normal_distribution
+# description: Sampling a log-normal distribution, that is, a distribution whose log is a normal distribution with mean μ and standard deviation σ.
+fn rand_lognorm(μ: Scalar, σ: Scalar) -> Scalar =
+    exp( μ + σ × rand_norm(0, 1) )
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -33,6 +33,17 @@ fn rand_bernoulli(p: Scalar) -> Scalar =
         else 0)
     else error("p must be a probability (0 <= p <= 1).")
 
+# name: Binomial distribution sampling
+# url: https://en.wikipedia.org/wiki/Binomial_distribution
+# description: Samples a binomial distribution by doing n Bernoulli tirals with probability p.
+#              Parameter n must be a positive integer, parameter p must be a probability (0 <= p <= 1).
+fn rand_binom(n: Scalar, p: Scalar) -> Scalar =
+    if n >= 1
+    then rand_binom(n-1, p) + rand_bernoulli(p)
+    else if n == 0
+    then 0
+    else error("n must be a positive integer.")
+
 # name: Normal distribution sampling
 # url: https://en.wikipedia.org/wiki/Normal_distribution
 # description: Samples a normal distribution with mean μ and standard deviation σ using the Box-Muller transform.
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -1,4 +1,6 @@
 use core::scalar
+use core::quantities
+use math::functions
 
 # name: Standard uniform distribution sampling
 # description: Uniformly samples the interval [0,1).
@@ -11,3 +13,11 @@ fn rand_uniform<T>(a: T, b: T) -> T =
     if a <= b
     then random() * (b - a) + a
     else random() * (a - b) + b
+
+# name: Discrete uniform distribution sampling
+# url: https://en.wikipedia.org/wiki/Discrete_uniform_distribution
+# description: Uniformly samples the integers in the interval [a, b].
+fn rand_int<T>(a: T, b: T) -> T =
+    if a <= b
+    then floor( random() * (floor(b) - ceil(a) + 1 unit_of(a)) ) + ceil(a)
+    else floor( random() * (floor(a) - ceil(b) + 1 unit_of(b)) ) + ceil(b)
--- a/numbat/modules/core/random.nbt
+++ b/numbat/modules/core/random.nbt
@@ -51,3 +51,12 @@ fn rand_expon<T>(λ: T) -> 1/T =
 # description: Sampling a log-normal distribution, that is, a distribution whose log is a normal distribution with mean μ and standard deviation σ.
 fn rand_lognorm(μ: Scalar, σ: Scalar) -> Scalar =
     exp( μ + σ × rand_norm(0, 1) )
+
+# name: Pareto distribution sampling
+# url: https://en.wikipedia.org/wiki/Pareto_distribution
+# description: Sampling a pareto distrobution with minimum value min and shape parameter α using inversion sampling.
+#              Both parameters α and min must be posive.
+fn rand_pareto<T>(α: Scalar, min: T) -> T =
+    if value_of(min) > 0 && α > 0
+    then min / ((1-random())^(1/α))
+    else NaN * unit_of(min)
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you!

I'd be very glad to accept more functions like these, especially if they can be implemented in Numbat. Before we do so, let's maybe discuss a minimum standard:

- Names: it would be great if we can choose a naming scheme that is close to something existing (e.g. numpy, scipy, matlab, R, …). This will make it easier for users to find what they want.
- Documentation: We do not have `@name`/`@url` etc. for functions, but I think it would be good to add a tiny bit of documentation for each function, and maybe a URL as a reference if someone wants to know more. I know that we don't have this for other functions in the prelude, but I very much would like to add this in the future, so this seems like a good place to start. Especially since some of those functions will not be universally known like `sin` or `sqrt`. This will make it easier for us to add proper "doc comments" for functions in the future.
- Discoverability: it would be great if you could list those functions in `book/src/list-functions.md`. Feel free to add a new section. `random` is also missing at the moment.
- Tests: This is hard to do for non-deterministic functions. I'd be okay with leaving out tests. We could add simple dummy tests like `x = rand_uniform(5, 7); assert(x >= 5 && x <= 7)` just to make sure that those functions will continue to work in the future. If you do so, please add them in a new `examples/random_tests.nbt` file (similar to what we do for `datetime`). For distributions without had boundary conditions, there's really not a lot we could test, but maybe it still makes sense to at least call them. Not sure.

Please feel free to add more of these in a single PR if that is easier for you.

Bzero:
Thanks a lot for your feedback!

- Names: Yes, I agree that this would be convenient. I would suggest to base the names on [scipy.stats](https://docs.scipy.org/doc/scipy/reference/stats.html) and to prefix them with `rand_` (or `random_` if you prefer) to make clear that these are non-deterministic functions.
- Documentation: Sounds good, I will add documentation as if there were `@name`, `@url`, maybe `@usage` function decorators,  so it will be easy to convert them if that feature will be added in the future.
- Discoverability: Agreed.
- Tests: So, I had actually written a small script which samples a large number of values and compares them to the distribution function (using scipy.stats) to check that the functions behave as intended. I think this method could serve as a test, I will see if I can do something similar with rust tests.

> Please feel free to add more of these in a single PR if that is easier for you.

Yes, that might be better. I will bundle them into a single PR.

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+    if prod > lim
+    then _poisson(lim,  prod × random()) + 1
+    else -1
+
+# name: Poisson distribution sampling
+# url: https://en.wikipedia.org/wiki/Poisson_distribution
+# description: Sampling a poisson distribution with rate λ, that is, the distribution of the number of events occurring in a fixed interval if these events occur with mean rate λ.
+#              The rate parameter λ must not be negative.
+# This implementation is based on the exponential distribution of inter-arrival times. For details see L. Devroye, Non-Uniform Random Variate Generation, p. 504, Lemma 3.3.
+fn rand_poisson<T>(λ: T) -> T =
```
sharkdp, line 73: Can you give a short justification for this type signature? At first, I thought it was wrong because it should return a dimensionless *count* of events. But getting an answer like 3/min for `rand_poisson(2/min)` seems reasonable. I'm just a bit worried because 2/min = 120/hour (for example), and `rand_poisson(120/hour)` would yield different results. Yes, the sum of poisson-distributed random variables is poisson-distributed, so in that sense
```
rand_poisson(120/hour)  ~  rand_poisson(2/min) + rand_poisson(2/min) + … another 58 times
```
but it still feels a bit strange to me?

Bzero: I chose this signature because the unit of the rate conveys information about the process, it matters if the events are per minute or per hour, especially if one continues to do computations with the results.  This could also be useful when using it with different dimensions, e.g or trees/m^2.

I agree however that it is a real issue that λ is ambiguous depending on the unit, especially if the unit is not immediately clear, for example `rand_poisson(100/hour - 1/min)`. 

One way to deal with this could be to split λ up into a scalar number of events and a normalization containing units.

Probably it is safest and easiest however to convert the signature to `fn rand_poisson(λ: Scalar) -> Scalar` so it only accepts and returns scalar values like the other discrete distributions and to force the user to take care of the unit explicitly if they need it.

Bzero:
Since you are already reviewing, here are a few more notes on the implementations that might be relevant which I had not  written up yet because I have not yet had time to look into the testing question:

* If a function receives an invalid parameter it returns an error if it is a scalar function and NAN otherwise. I hope that is okay this way, if there is a way to return errors for non-scalar functions in the future we can change that.

* The normal distribution is implemented slightly inefficiently as it has to discard every second sample.

* The method used to sample the Bernoulli distribution has suboptimal linear runtime in n. Unless people start doing heavy number crunching this should not be a problem and for large n the approximation by the normal distribution can be used.

* Similarly, the implementation of the Poisson distribution has linear expected runtime in λ. Again, in practice this should not be an issue and for large λ the normal distribution can be used too.

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+fn rand_bernoulli(p: Scalar) -> Scalar =
+    if p>=0 && p<=1
+    then (if random() < p
+        then 1
+        else 0)
+    else error("p must be a probability (0 <= p <= 1).")
+
+# name: Binomial distribution sampling
+# url: https://en.wikipedia.org/wiki/Binomial_distribution
+# description: Samples a binomial distribution by doing n Bernoulli tirals with probability p.
```
sharkdp, line 38: minor: typo
```suggestion
# description: Samples a binomial distribution by doing n Bernoulli trials with probability p.
```

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+    then rand_binom(n-1, p) + rand_bernoulli(p)
+    else if n == 0
+    then 0
+    else error("n must be a positive integer.")
+
+# name: Normal distribution sampling
+# url: https://en.wikipedia.org/wiki/Normal_distribution
+# description: Samples a normal distribution with mean μ and standard deviation σ using the Box-Muller transform.
+fn rand_norm<T>(μ: T, σ: T) -> T =
+    μ + sqrt(-2 σ² × ln(random())) × sin(2π × random())
```
sharkdp, line 51: :heart_eyes:

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+
+# name: Geometric distribution sampling
+# url: https://en.wikipedia.org/wiki/Geometric_distribution
+# description: Samples a gemoetric distribution(the distribution of the number of Bernoulli trials with probability p needed to get one success) by inversion sampling.
```
sharkdp, line 55: minor: typo
```suggestion
# description: Samples a geometric distribution (the distribution of the number of Bernoulli trials with probability p needed to get one success) by inversion sampling.
```

numbat/modules/core/random.nbt
sharkdp: > especially if the unit is not immediately clear, for example `rand_poisson(100/hour - 1/min)`.

:+1:

> Probably it is safest and easiest however to convert the signature to `fn rand_poisson(λ: Scalar) -> Scalar` so it only accepts and returns scalar values like the other discrete distributions and to force the user to take care of the unit explicitly if they need it.

It's a bit sad, but yes. I think it would be best to restrict it to scalars.


I think what might be useful is for `rand_poisson(40 trees / hectare)` to give an answer like `35 trees` (not `35 trees / hectare`), because we're asking for one particular outcome of that random experiment. So maybe the count is the more descriptive result compared to a rate, because the latter would indicate some kind of mean value. But that might be even more confusing.

And of course it's impossible to implement in this form, because there is no way of knowing what is the count and what is the period/interval.

We *could* use a two-argument function like you proposed, but instead of splitting up magnitude and unit, we could maybe split up the rate into count (40 trees) and interval (hectare). Like so:
```
fn rand_poisson_unitful<C, I>(count: C, interval: I) -> C = …
```
But that's also a bit weird because we would basically just throw away the second argument :smile:.

I think we can just go with the Scalar->Scalar version for now, unless we still come up with a good solution.

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+#              The rate parameter λ must not be negative.
+# This implementation is based on the exponential distribution of inter-arrival times. For details see L. Devroye, Non-Uniform Random Variate Generation, p. 504, Lemma 3.3.
+fn rand_poisson<T>(λ: T) -> T =
+    if value_of(λ) >= 0
+    then _poisson(exp(-value_of(λ)), 1) * unit_of(λ)
+    else NaN * unit_of(λ)
+
+# name: Exponential distribution sampling
+# url: https://en.wikipedia.org/wiki/Exponential_distribution
+# description: Sampling an exponential distribution(the distribution of the distance between events in a Poisson process with rate λ) using inversion sampling.
```
sharkdp, line 80:
```suggestion
# description: Sampling an exponential distribution (the distribution of the distance between events in a Poisson process with rate λ) using inversion sampling.
```

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+#              The rate parameter λ must be positive.
+fn rand_expon<T>(λ: T) -> 1/T =
+    if value_of(λ) > 0
+    then - ln(1-random()) / λ
+    else NaN * unit_of(1/λ)
```
sharkdp, line 85: Here, the generic T->1/T signature comes out naturally and makes a lot of sense. Nice!

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+
+# name: Log-normal distribution sampling
+# url: https://en.wikipedia.org/wiki/Log-normal_distribution
+# description: Sampling a log-normal distribution, that is, a distribution whose log is a normal distribution with mean μ and standard deviation σ.
+fn rand_lognorm(μ: Scalar, σ: Scalar) -> Scalar =
+    exp( μ + σ × rand_norm(0, 1) )
+
+# name: Pareto distribution sampling
+# url: https://en.wikipedia.org/wiki/Pareto_distribution
+# description: Sampling a pareto distrobution with minimum value min and shape parameter α using inversion sampling.
```
sharkdp, line 95:
```suggestion
# description: Sampling a Pareto distribution with minimum value min and shape parameter α using inversion sampling.
```

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+#              Both parameters α and min must be posive.
```
sharkdp, line 96:
```suggestion
#              Both parameters α and min must be positive.
```

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+fn rand_pareto<T>(α: Scalar, min: T) -> T =
+    if value_of(min) > 0 && α > 0
+    then min / ((1-random())^(1/α))
+    else NaN * unit_of(min)
```
sharkdp, line 100: We should really make 0, NaN and inf polymorphic (#37).

sharkdp:
Thank you very much — this looks fantastic!

I wrote a small example program (in Rust... we really need a list data type in Numbat!) to sample large amounts of numbers from those distributions and compared the normalized histograms with the PDFs of those distributions. Everything looked correct :+1: 

```rs
use numbat::{
    module_importer::FileSystemImporter, resolver::CodeSource, Context, InterpreterResult,
};
use std::path::Path;

fn get_value(ctx: &mut Context, code: &str) -> f64 {
    match ctx.interpret(code, CodeSource::Internal).unwrap().1 {
        InterpreterResult::Value(v) => v.unsafe_as_quantity().unsafe_value().0,
        _ => todo!(),
    }
}

fn main() {
    let module_path = Path::new(&std::env::var_os("CARGO_MANIFEST_DIR").unwrap()).join("modules");

    let mut importer = FileSystemImporter::default();
    importer.add_path(module_path);
    let mut ctx = Context::new(importer);
    let _result = ctx.interpret("use all", CodeSource::Internal).unwrap();

    let code = std::env::args().nth(1).unwrap();

    let results = (0..10000)
        .map(|_| get_value(&mut ctx, code.as_str()))
        .collect::<Vec<_>>();
    for r in results {
        println!("{}", r);
    }
}
```

And a Python script to plot a histogram:
```
import matplotlib.pyplot as plt
import numpy as np
import sys

data = np.loadtxt(sys.argv[1])

weights = np.ones_like(data) / len(data)
plt.hist(data, bins=50, weights=weights)
plt.show()
```

> Since you are already reviewing, here are a few more notes on the implementations that might be relevant which I had not written up yet because I have not yet had time to look into the testing question:
> 
>     * If a function receives an invalid parameter it returns an error if it is a scalar function and NAN otherwise. I hope that is okay this way, if there is a way to return errors for non-scalar functions in the future we can change that.

:+1: 

>     * The normal distribution is implemented slightly inefficiently as it has to discard every second sample.

I think that is completely acceptable. Changing that would require us to keep state somewhere (like in a static variable), right?

>     * The method used to sample the Bernoulli distribution has suboptimal linear runtime in n. Unless people start doing heavy number crunching this should not be a problem and for large n the approximation by the normal distribution can be used.

:+1: 

>     * Similarly, the implementation of the Poisson distribution has linear expected runtime in λ. Again, in practice this should not be an issue and for large λ the normal distribution can be used too.

:+1:

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+# name: Geometric distribution sampling
+# url: https://en.wikipedia.org/wiki/Geometric_distribution
+# description: Samples a gemoetric distribution(the distribution of the number of Bernoulli trials with probability p needed to get one success) by inversion sampling.
+#              Parameter p must be a probability (0 <= p <= 1).
+fn rand_geom(p: Scalar) -> Scalar =
+    if p>=0 && p<=1
+    then ceil( ln(1-random()) / ln(1-p) )
+    else error("p must be a probability (0 <= p <= 1).")
+
+# A helper function for rand_poisson, counts how many samples of the standard uniform distribution need to be multiplied to deceed lim.
```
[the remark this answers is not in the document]
Bzero: The opposite of exceed. But I suppose it is a neologism, I can change it.

numbat/modules/core/random.nbt
Bzero: I agree, a scalar function is the cleanest way and also the standard one people will be familiar with, unlike any two argument Poisson functions we might make up.

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,100 @@
...
+    then random() * (b - a) + a
+    else random() * (a - b) + b
+
+# name: Discrete uniform distribution sampling
+# url: https://en.wikipedia.org/wiki/Discrete_uniform_distribution
+# description: Uniformly samples the integers in the interval [a, b].
+fn rand_int<T>(a: T, b: T) -> T =
+    if a <= b
+    then floor( random() * (floor(b) - ceil(a) + 1 unit_of(a)) ) + ceil(a)
+    else floor( random() * (floor(a) - ceil(b) + 1 unit_of(b)) ) + ceil(b)
```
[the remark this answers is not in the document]
Bzero: Yes, looks like `unit_of` and `value_of` are indeed more dangerous than I was aware of 😅.

Lets change it to `Scalar` too, it was probably never a good idea to add units to discrete distributions in the first place.

numbat/modules/core/random.nbt
Bzero: Yes, I think for continuous distributions units work without any problems.

sharkdp:
> * If a function receives an invalid parameter it returns an error if it is a scalar function and NAN otherwise. I hope that is okay this way, if there is a way to return errors for non-scalar functions in the future we can change that.

I just implemented a proper version of `error(…)` here: https://github.com/sharkdp/numbat/pull/420. It would be great if you could rebase/merge with master and use `error(…)` instead of NaN.

Bzero:
Thanks a lot for all your comments and catching my mistakes!

> I wrote a small example program (in Rust... we really need a list data type in Numbat!) to sample large amounts of numbers from those distributions and compared the normalized histograms with the PDFs of those distributions. Everything looked correct 👍
> 
> ```rust
> use numbat::{
>     module_importer::FileSystemImporter, resolver::CodeSource, Context, InterpreterResult,
> };
> use std::path::Path;
> 
> fn get_value(ctx: &mut Context, code: &str) -> f64 {
>     match ctx.interpret(code, CodeSource::Internal).unwrap().1 {
>         InterpreterResult::Value(v) => v.unsafe_as_quantity().unsafe_value().0,
>         _ => todo!(),
>     }
> }
> 
> fn main() {
>     let module_path = Path::new(&std::env::var_os("CARGO_MANIFEST_DIR").unwrap()).join("modules");
> 
>     let mut importer = FileSystemImporter::default();
>     importer.add_path(module_path);
>     let mut ctx = Context::new(importer);
>     let _result = ctx.interpret("use all", CodeSource::Internal).unwrap();
> 
>     let code = std::env::args().nth(1).unwrap();
> 
>     let results = (0..10000)
>         .map(|_| get_value(&mut ctx, code.as_str()))
>         .collect::<Vec<_>>();
>     for r in results {
>         println!("{}", r);
>     }
> }
> ```
> 
> And a Python script to plot a histogram:
> 
> ```python
> import matplotlib.pyplot as plt
> import numpy as np
> import sys
> 
> data = np.loadtxt(sys.argv[1])
> 
> weights = np.ones_like(data) / len(data)
> plt.hist(data, bins=50, weights=weights)
> plt.show()
> ```

Thanks, this is really helpful! I think if we wanted to integrate this in the normal testing one should be able to do a Kolmogorov-Smirnov test to automatically judge whether the expected pdf fits the sampled data. I will have a look if there is a rust library that would let us do this easily.


> > The normal distribution is implemented slightly inefficiently as it has to discard every second sample.
> 
> I think that is completely acceptable. Changing that would require us to keep state somewhere (like in a static variable), right?

Yes, the Box-Muller transform provides two independent random numbers at a time and we only use one of them.

Bzero:
> I just implemented a proper version of `error(…)` here: #420. It would be great if you could rebase/merge with master and use `error(…)` instead of NaN.

Oh perfect, thanks a lot, this will make it a lot more user friendly!

sharkdp:
> Thanks, this is really helpful! I think if we wanted to integrate this in the normal testing one should be able to do a Kolmogorov-Smirnov test to automatically judge whether the expected pdf fits the sampled data. I will have a look if there is a rust library that would let us do this easily.

I'd be okay with merging this without any additional tests. If you want to work on that, please feel free to go ahead. But let's also keep test execution time in mind. Drawing 10000 samples in debug mode (which is what we usually have for `cargo test`) took ~seconds. Not sure how many samples would be needed for a proper test, but I'd rather not have any tests that take >1s, if possible.

Bzero:
> I'd be okay with merging this without any additional tests. If you want to work on that, please feel free to go ahead. ~~But let's also keep test execution time in mind. Drawing 10000 samples in debug mode (which is what we usually have for `cargo test`) took ~seconds. Not sure how many samples would be needed for a proper test, but I'd rather not have any tests that take >1s, if possible.~~
> 
> Edit: nevermind. if you want to implement this, let's just do it. Worst case, we can deactivate those tests locally and only run them in CI, if they take too long.

Well, I am not totally sure it is worth the effort then. Implementation would not be straight forward for me as I am not familiar with any of the rust libraries that can do K-S tests and these functions are not likely going to be modified a lot or rely on parts that are likely change a lot so it may be good enough to manually compare the histograms instead of doing it automatically.

numbat/modules/core/random.nbt
```diff
@@ -1,3 +1,102 @@
...
+    else error("Argument λ must not be negative.")
+
+# name: Exponential distribution sampling
+# url: https://en.wikipedia.org/wiki/Exponential_distribution
+# description: Sampling an exponential distribution (the distribution of the distance between events in a Poisson process with rate λ) using inversion sampling.
+#              The rate parameter λ must be positive.
+fn rand_expon<T>(λ: T) -> 1/T =
+    if value_of(λ) > 0
+    then - ln(1-random()) / λ
+    else error("Argument λ must not be positive.")
```
sharkdp, line 87:
```suggestion
    else error("Argument λ must be positive.")
```

Bzero: Thanks for catching this, that was a silly mistake.

sharkdp:
Thank you!

Bzero:
Welcome, thanks a lot for your help!

Issue #166, opened 2023-09-20
Can not infer type parameter when calling generic function from generic function


sharkdp:
```
>>> fn f<T>(y: T, x: T) = atan2(y, x)
error: while type checking
   ┌─ <input:1>:1:23
   │
 1 │ fn f<T>(y: T, x: T) = atan2(y, x)
   │                       ^^^^^ … could not be infered for this function call
   │
   ┌─ Module 'math::functions', File /home/ped1st/.config/numbat/modules/math/functions.nbt:30:4
   │
30 │ fn atan2<T>(y: T, x: T) -> Scalar
   │    ----- The type parameter(s)  in this generic function
```

Weirdly, it works if we give that type parameter another name:
```
fn f<S>(y: S, x: S) = atan2(y, x)
```

So it looks like we need to properly separate namespaces here.

PR #418, opened 2024-04-27
Add support for struct/record types

simmsb:
Hey, I've been playing with numbat recently and I've found the type system to work perfectly for my use case (calculating inductor, resistor, etc parameters for a boost converter flashlight driver), however the lack of record types is causing me to repeat myself a bit too much.

Here I've added an implementation of structural record types that are used like so:

```nbt
>>> ${foo: 1A, bar: 3V}

  ${foo: 1 ampere, bar: 3 volt}

    = ${foo: 1 A, bar: 3 V}    [${foo: Current, bar: Length² × Mass / (Current × Time³)}]

>>> ${foo: 1A, bar: 3V}.foo

  ${foo: 1 ampere, bar: 3 volt}.foo

    = 1 A    [Current]
```

```nbt
# A value level struct
${x: Length, y: Length}

# A function with a struct as a parameter
fn euclidian_distance(a: ${x: Length, y: Length}, b: ${x: Length, y: Length}) =
  sqrt(sqr(a.x - b.x) + sqr(a.y - b.y))
  
assert_eq(
  euclidian_distance(${x: 0m, y: 0m}, ${x: 6m, y: 8m}),
  10m)
  
let r = ${foo: 1}

# Struct fields can be accessed using `.field` notation
let x = r.foo
```

The choice of the `${` syntax for records was because it looked fairly easy to add unambiguously.

Records are implemented in the interpreter with a new value type, and two new opcodes:
- `Value::Struct(Arc<[(String, usize)]>, Vec<Value>)`
Contains the field metadata of the struct (the list of field names and mapped value locations), and the list of values of the struct. The field metadata is only used when printing the fields of the struct.
- `Op::BuildStruct`
Takes two parameters: a 'struct metadata index' and the number of fields, uses the metadata index to retrieve the struct field metadata from the VM, then pops N values from the stack and places them and the metadata in a struct value
- `Op::DestructureStruct`
Takes one parameter: the index into the struct to retrieve. Pops a struct from the stack and pushes back on the value at the given index.

Please feel free to decline this PR, or request significant changes if you intended to go with a different direction for record types :)

### DIFF
--- a/book/src/structs.md
+++ b/book/src/structs.md
@@ -1,21 +1,24 @@
 # Struct types
 
-Numbat supports structurally typed records:
+Numbat supports nominally typed records:
 
 ```nbt
-# A value level struct
-${x: Length, y: Length}
+# Define a struct
+struct Vector {
+  x: Length,
+  y: Length,
+}
+
+let v = Vector {x: 6m, y: 8m}
 
 # A function with a struct as a parameter
-fn euclidian_distance(a: ${x: Length, y: Length}, b: ${x: Length, y: Length}) =
+fn euclidian_distance(a: Vector, b: Vector) =
   sqrt(sqr(a.x - b.x) + sqr(a.y - b.y))
   
 assert_eq(
-  euclidian_distance(${x: 0m, y: 0m}, ${x: 6m, y: 8m}),
+  euclidian_distance(Vector {x: 0m, y: 0m}, v),
   10m)
   
-let r = ${foo: 1}
-
 # Struct fields can be accessed using `.field` notation
-let x = r.foo
+let x = v.x
 ```
--- a/book/src/structs.md
+++ b/book/src/structs.md
@@ -13,7 +13,7 @@ let v = Vector {x: 6m, y: 8m}
 
 # A function with a struct as a parameter
 fn euclidian_distance(a: Vector, b: Vector) =
-  sqrt(sqr(a.x - b.x) + sqr(a.y - b.y))
+  sqrt((a.x - b.x)² + (a.y - b.y)²)
   
 assert_eq(
   euclidian_distance(Vector {x: 0m, y: 0m}, v),
--- a/book/src/structs.md
+++ b/book/src/structs.md
@@ -1,0 +1,21 @@
+# Struct types
+
+Numbat supports structurally typed records:
+
+```nbt
+# A value level struct
+${x: Length, y: Length}
+
+# A function with a struct as a parameter
+fn euclidian_distance(a: ${x: Length, y: Length}, b: ${x: Length, y: Length}) =
+  sqrt(sqr(a.x - b.x) + sqr(a.y - b.y))
+  
+assert_eq(
+  euclidian_distance(${x: 0m, y: 0m}, ${x: 6m, y: 8m}),
+  10m)
+  
+let r = ${foo: 1}
+
+# Struct fields can be accessed using `.field` notation
+let x = r.foo
+```
--- a/book/src/structs.md
+++ b/book/src/structs.md
@@ -1,6 +1,6 @@
 # Struct types
 
-Numbat supports nominally typed records:
+Numbat supports structs as compound data structures:
 
 ```nbt
 # Define a struct
--- a/book/src/structs.md
+++ b/book/src/structs.md
@@ -1,6 +1,6 @@
-# Struct types
+# Structs
 
-Numbat supports structs as compound data structures:
+Numbat compound data structures in the form of structs:
 
 ```nbt
 # Define a struct
@@ -9,16 +9,15 @@ struct Vector {
   y: Length,
 }
 
-let v = Vector {x: 6m, y: 8m}
+let origin = Vector { x: 0 m, y: 0 m }
+let position = Vector { x: 6 m, y: 8 m }
 
 # A function with a struct as a parameter
-fn euclidian_distance(a: Vector, b: Vector) =
+fn euclidean_distance(a: Vector, b: Vector) =
   sqrt((a.x - b.x)² + (a.y - b.y)²)
   
-assert_eq(
-  euclidian_distance(Vector {x: 0m, y: 0m}, v),
-  10m)
+assert_eq(euclidean_distance(origin, position), 10 m)
   
 # Struct fields can be accessed using `.field` notation
-let x = v.x
+let x = position.x
 ```
--- a/examples/numbat_syntax.nbt
+++ b/examples/numbat_syntax.nbt
@@ -111,14 +111,15 @@ assert_eq(1 yd, 1 m, 10 cm)      # Assert that two quantities are equal, up to
 type(2 m/s)                      # Print the type of an expression
 
 
-# 9. Structs
+# 10. Structs
 
-# defining a struct
-struct Foo { foo: Current, bar: Scalar }
+struct Foo {                        # defining a struct
+  foo: Current,
+  bar: Scalar
+}
 
-# Constructing a struct
-Foo {foo: 1A, bar: pi}
+Foo {foo: 1A, bar: pi}              # constructing a struct
 
 let r: Foo = Foo {foo: 1A, bar: pi}
 
-r.foo  # Field access is performed with `.field` notation
+r.foo                               # Field access is performed with `.field` notation
--- a/examples/numbat_syntax.nbt
+++ b/examples/numbat_syntax.nbt
@@ -113,9 +113,12 @@ type(2 m/s)                      # Print the type of an expression
 
 # 9. Structs
 
-${foo: 1A, bar: pi}  # Constructing a record
+# defining a struct
+struct Foo { foo: Current, bar: Scalar }
 
-# Records are structurally typed and field order is not fixed
-let r: ${bar: Scalar, foo: Current} = ${foo: 1A, bar: pi}
+# Constructing a struct
+Foo {foo: 1A, bar: pi}
+
+let r: Foo = Foo {foo: 1A, bar: pi}
 
 r.foo  # Field access is performed with `.field` notation
--- a/examples/numbat_syntax.nbt
+++ b/examples/numbat_syntax.nbt
@@ -113,13 +113,16 @@ type(2 m/s)                      # Print the type of an expression
 
 # 10. Structs
 
-struct Foo {                        # defining a struct
-  foo: Current,
-  bar: Scalar
+struct Element {                 # Define a struct
+    name: String,
+    atomic_number: Scalar,
+    density: MassDensity,
 }
 
-Foo {foo: 1A, bar: pi}              # constructing a struct
+let hydrogen = Element {         # Instantiate it
+    name: "Hydrogen",
+    atomic_number: 1,
+    density: 0.08988 g/L,
+}
 
-let r: Foo = Foo {foo: 1A, bar: pi}
-
-r.foo                               # Field access is performed with `.field` notation
+hydrogen.density                 # Access field of a struct
--- a/examples/numbat_syntax.nbt
+++ b/examples/numbat_syntax.nbt
@@ -125,4 +125,4 @@ let hydrogen = Element {         # Instantiate it
     density: 0.08988 g/L,
 }
 
-hydrogen.density                 # Access field of a struct
+hydrogen.density                 # Access the field of a struct
--- a/examples/numbat_syntax.nbt
+++ b/examples/numbat_syntax.nbt
@@ -109,3 +109,13 @@ assert_eq(1 ft, 12 in)           # Assert that two quantities are equal
 assert_eq(1 yd, 1 m, 10 cm)      # Assert that two quantities are equal, up to
                                  # the given precision
 type(2 m/s)                      # Print the type of an expression
+
+
+# 9. Structs
+
+${foo: 1A, bar: pi}  # Constructing a record
+
+# Records are structurally typed and field order is not fixed
+let r: ${bar: Scalar, foo: Current} = ${foo: 1A, bar: pi}
+
+r.foo  # Field access is performed with `.field` notation
--- a/numbat/src/ast.rs
+++ b/numbat/src/ast.rs
@@ -203,6 +203,18 @@ macro_rules! conditional {
 }
 
 #[cfg(test)]
+macro_rules! struct_ {
+    ( $( $field:ident : $val:expr ),* ) => {{
+        crate::ast::Expression::MakeStruct(
+            Span::dummy(),
+            vec![
+                $((Span::dummy(), stringify!($field).to_owned(), $val)),*
+            ]
+        )
+    }};
+}
+
+#[cfg(test)]
 pub(crate) use binop;
 #[cfg(test)]
 pub(crate) use boolean;
@@ -218,6 +230,8 @@ pub(crate) use logical_neg;
 pub(crate) use negate;
 #[cfg(test)]
 pub(crate) use scalar;
+#[cfg(test)]
+pub(crate) use struct_;
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum TypeAnnotation {
--- a/numbat/src/ast.rs
+++ b/numbat/src/ast.rs
@@ -85,6 +85,8 @@ pub enum Expression {
     Boolean(Span, bool),
     String(Span, Vec<StringPart>),
     Condition(Span, Box<Expression>, Box<Expression>, Box<Expression>),
+    MakeStruct(Span, Vec<(Span, String, Expression)>),
+    AccessStruct(Span, Span, Box<Expression>, String),
 }
 
 impl Expression {
@@ -116,6 +118,8 @@ impl Expression {
                 span_if.extend(&then_expr.full_span())
             }
             Expression::String(span, _) => *span,
+            Expression::MakeStruct(full_span, _) => *full_span,
+            Expression::AccessStruct(full_span, _ident_span, _, _) => *full_span,
         }
     }
 }
@@ -223,6 +227,7 @@ pub enum TypeAnnotation {
     String(Span),
     DateTime(Span),
     Fn(Span, Vec<TypeAnnotation>, Box<TypeAnnotation>),
+    Struct(Span, Vec<(Span, String, TypeAnnotation)>),
 }
 
 impl TypeAnnotation {
@@ -234,6 +239,7 @@ impl TypeAnnotation {
             TypeAnnotation::String(span) => *span,
             TypeAnnotation::DateTime(span) => *span,
             TypeAnnotation::Fn(span, _, _) => *span,
+            TypeAnnotation::Struct(span, _) => *span,
         }
     }
 }
@@ -260,6 +266,17 @@ impl PrettyPrint for TypeAnnotation {
                     + m::space()
                     + return_type.pretty_print()
                     + m::operator("]")
+            }
+            TypeAnnotation::Struct(_, fields) => {
+                m::operator("${")
+                    + Itertools::intersperse(
+                        fields.iter().map(|(_, n, t)| {
+                            m::identifier(n) + m::operator(":") + m::space() + t.pretty_print()
+                        }),
+                        m::operator(",") + m::space(),
+                    )
+                    .sum()
+                    + m::operator("}")
             }
         }
     }
@@ -397,6 +414,13 @@ impl ReplaceSpans for TypeAnnotation {
             TypeAnnotation::Fn(_, pt, rt) => {
                 TypeAnnotation::Fn(Span::dummy(), pt.clone(), rt.clone())
             }
+            TypeAnnotation::Struct(_, fields) => TypeAnnotation::Struct(
+                Span::dummy(),
+                fields
+                    .iter()
+                    .map(|(_, n, v)| (Span::dummy(), n.clone(), v.replace_spans()))
+                    .collect(),
+            ),
         }
     }
 }
@@ -492,6 +516,19 @@ impl ReplaceSpans for Expression {
             Expression::String(_, parts) => Expression::String(
                 Span::dummy(),
                 parts.iter().map(|p| p.replace_spans()).collect(),
+            ),
+            Expression::MakeStruct(_, fields) => Expression::MakeStruct(
+                Span::dummy(),
+                fields
+                    .iter()
+                    .map(|(_, n, v)| (Span::dummy(), n.clone(), v.replace_spans()))
+                    .collect(),
+            ),
+            Expression::AccessStruct(_, _, expr, attr) => Expression::AccessStruct(
+                Span::dummy(),
+                Span::dummy(),
+                Box::new(expr.replace_spans()),
+                attr.clone(),
             ),
         }
     }
--- a/numbat/src/ast.rs
+++ b/numbat/src/ast.rs
@@ -85,7 +85,12 @@ pub enum Expression {
     Boolean(Span, bool),
     String(Span, Vec<StringPart>),
     Condition(Span, Box<Expression>, Box<Expression>, Box<Expression>),
-    MakeStruct(Span, Vec<(Span, String, Expression)>),
+    MakeStruct {
+        full_span: Span,
+        ident_span: Span,
+        name: String,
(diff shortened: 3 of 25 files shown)


sharkdp:
Hi. I have not yet done a detailed review, but let start by saying that I am extremely excited about this PR! This is definitely something I was planning to add to the language.

> Hey, I've been playing with numbat recently and I've found the type system to work perfectly for my use case (calculating inductor, resistor, etc parameters for a boost converter flashlight driver), however the lack of record types is causing me to repeat myself a bit too much.

:+1: 

> Here I've added an implementation of structural record types

I feel a bit bad to go back to the design level here because you already added this complete implementation including documentation, error handling, etc.

But before we go ahead with this, I would really like to understand if *structural* record types are the best choice for Numbat, or if we should rather go with a nominative system.

I am familiar with structural record types from languages like PureScript. If you don't have an additional (strong) newtype mechanism on top, there's the obvious problem of record types colliding with other (structurally same) record types of completely different semantic meaning. For example, you might want to have `${x: Length, y: Length}` for two different coordinate systems which you don't want to confuse. 

I think my first impulse would have been to add more "usual" struct/record types to Numbat that would be based on the name of the type.. with a Rust-like syntax:
``` rs
struct WorldCoordinate {
  x: Length,
  y: Length,
}

struct LocalCoordinate {
  x: Length,
  y: Length,
}

fn world_to_local(pos: WorldCoordinate) -> LocalCoordinate = …
```

I'm curious why you decided to go the other way? Would a nominative record type be harder to implement?

simmsb:
> I'm curious why you decided to go the other way? Would a nominative record type be harder to implement?

This was pretty much the reason, with named structs the parser would either need to become context sensitive, or we add an extra step after parsing or during typechecking to resolve whether an identifier is a dimension or a struct name.

I think the implementation of structs in the evaluator can stay the same, I'm happy to work on changing this to a nominal system.

Do you have any thoughts on the best way to go about parsing and disambiguating struct and dimension names?

sharkdp:
> I think the implementation of structs in the evaluator can stay the same, I'm happy to work on changing this to a nominal system.

It would be great if we could explore this. If it turns out to be more complex than we think now, please feel free to discuss it first before spending a lot of effort.

> Do you have any thoughts on the best way to go about parsing and disambiguating struct and dimension names?

So before we talk about implementation, maybe we should consider the language design first. Is it confusing for users to have both struct types (and later maybe enums, type aliases, …) in the same namespace as the dimension types? I mean, we have the same problem with the other builtin types (String, Bool, DateTime, …), but it might be nice to know from reading the code whether something less familiar like `Area` is a dimension type or a struct (`struct Area { width: Length, height: Length }`). But that would require us to either work with a prefix (like `$Area` for struct types) or some kind of case convention (Upper camel case for dimension types, lower snake case for other types). That would solve the problem immediately (in the parser), but somehow I don't like those syntax/name restrictions. So unless we find a(nother) good reason to separate them by prefix/naming-convention, let's assume we don't put any restrictions on the struct names.

Which leaves us with the problem you mentioned. In the parser, we don't know what we are parsing when we see `let x: Area …`. I would probably leave the parser context free and solve that in a later stage. We do something similar for prefixed units like `km` where the parser doesn't know if that is a normal identifier, or a unit `m` which is prefixed by `k` for kilo. We resolve that later in the "prefix transformer" stage, which comes right after the parser, before the type checker.

We could generalize this to a "name resolution" stage that would solve both the prefix and the dimension-vs-normal-type ambiguity. It would rely on seeing the `dimension` or `struct` definition first.

Another option which might be more reasonable is to do it in the type checker. We already add all dimension types to `TypeChecker::registry`. And all function definitions that we see to `TypeChecker::function_signatures`. Similarly, we would probably have a place to store all user-defined non-dimension types, together with the information we need for type checking (field names with corresponding field types). So whenever we see a type name like `Area`, we could check which class it belongs to. Why do you think it requires an additional stage/pass? I guess that would certainly be required if types could be declared *after their usage*, but I would postpone that problem for now (?).

In both cases, we would have to modify the parser.. at least naming-wise. What is now called `DimensionExpression` would probably be renamed to `TypeExpression`. And `DimensionExpression::Dimension` would be renamed to `TypeExpression::TypeIdentifier`. And then we would simply allow the *parsing* of type annotations like `Area^2`, even if `Area` would be a struct type. We would only throw an error later in the type checker when we see a non-dimension type being used in a `DimensionExpression::Multiply`/`Divide`/`Power` type-level operation.

Do you think that could work, or do you see additional problems?

sharkdp:
Just a side note regarding your current implementation: the record types do not work with generics.
```
>>> fn norm<T>(p: ${x: T, y: T}) -> T = sqrt(p.x^2 + p.y^2)

  fn norm<T>(p: ${x: T, y: T}) -> T = sqrt(p.x² + p.y²)

>>> norm(${x: 3, y: 4})
error: while type checking
  ┌─ <input:15>:1:12
  │
1 │ fn norm<T>(p: ${x: T, y: T}) -> T = sqrt(p.x^2 + p.y^2)
  │            - ${x: T, y: T}
  │
  ┌─ <input:16>:1:6
  │
1 │ norm(${x: 3, y: 4})
  │      ^^^^^^^^^^^^^ ${x: Scalar, y: Scalar}
  │
  = Incompatible types in function call: expected '${x: T, y: T}', got '${x: Scalar, y: Scalar}' instead
```

This is not a problem of your change, but more a problem with my buggy generics implementation… which does not properly distinguish between type parameters and concrete types in some cases (see #166 and a non-ready attempt to resolve it: #253). 

Just wanted to mention it. I would probably skip generics for structs in the first iteration. Later, something like
```
struct Coordinate<T> {
  x: T,
  y: T,
}
```
might be nice to have. But there are *a lot* of other things that are more important :smile:.

simmsb:
> We could generalize this to a "name resolution" stage that would solve both the prefix and the dimension-vs-normal-type ambiguity. It would rely on seeing the `dimension` or `struct` definition first.

This makes sense and I think it's the best route to take, I'll have a play around with implementing this.

> We would only throw an error later in the type checker when we see a non-dimension type being used in a `DimensionExpression::Multiply`/`Divide`/`Power` type-level operation.

Yeah, I suppose I'll keep structs and dimensions separate in the typed AST, and do the renaming of `DimensionExpression` you suggest in the untyped AST.

> record types do not work with generics.

Ah, I actually made this work for return types, are parameter generics significantly different?

simmsb:
I've made structs nominally typed using the syntax you've proposed, I also refactored a small amount how name collisions are handled so that checking for name collisions between items in the 'type' namespace (and also 'value' namespace) are a bit simpler (however I didn't refactor the collision checking that occurs within the prefix transformer)

As for disambiguating dimensions and struct names, I simply check if a type expression is only a type identifier with a name matching a struct, and take it to be a struct type here: https://github.com/simmsb/numbat/blob/simmsb/struct-types/numbat/src/typechecker.rs#L1835-L1842.  I think this is sound as a dimension name cannot clash with a struct.

sharkdp:
Thank you very much for the update — I tried a few things and it looks great!

Some remarks before I dive deeper into the review:
- Some tests are currently failing. It looks like newlines are currently not supported in struct definitions (but of course it would be great if we could allow them)
- It should not be allowed to define a `struct Meter` (or similar) with a name that is already taken by a dimension type. And vice versa.
- I think it would be great to have some basic tests (1) in the parser, (2) for name resolving but most importantly (3) for the type checker.
- The new `struct` keyword should be added to `numbat/src/keywords.rs` (see comment in `tokenizer.rs` to keep the lists in sync)

simmsb:
I've fixed all the tests and CI, there's now tests for parsing around structs and some more tests for type checking of structs, and name collisions.

Struct names do collide properly with dimension names, but not with unit, variable, and function names.
```
>>> struct Current {}
error: identifier clash in definition
   ┌─ Module 'core::dimensions', File <builtin>/modules/core/dimensions.nbt:32:11
   │
32 │ dimension Current
   │           ------- Previously defined dimension here
   │
   ┌─ <input:3>:1:8
   │
 1 │ struct Current {}
   │        ^^^^^^^ identifier is already in use
```

numbat/src/parser.rs
```diff
@@ -38,9 +39,10 @@
 //! power           ::=   factorial ( "^" "-" ? power ) ?
 //! factorial       ::=   unicode_power "!" *
 //! unicode_power   ::=   call ( "⁻" ? ( "¹" | "²" | "³" | "⁴" | "⁵" | "⁶" | "⁷" | "⁸" | "⁹" ) ) ?
-//! call            ::=   primary ( "(" arguments? ")" ) *
+//! call            ::=   primary ( ( "(" arguments? ")" ) | "." identifier ) *
 //! arguments       ::=   expression ( "," expression ) *
-//! primary         ::=   boolean | string | hex_number | oct_number | bin_number | number | identifier | "(" expression ")"
+//! primary         ::=   boolean | string | hex_number | oct_number | bin_number | number | identifier ( struct_expr ? ) | "(" expression ")" | struct_expr
```
sharkdp, line 44: Minor: can we keep the recursive case at the end? (for aesthetic reasons only)
```suggestion
//! primary         ::=   boolean | string | hex_number | oct_number | bin_number | number | identifier ( struct_expr ? ) | struct_expr | "(" expression ")"
```

numbat/src/ast.rs
```diff
@@ -85,6 +85,13 @@ pub enum Expression {
     Boolean(Span, bool),
     String(Span, Vec<StringPart>),
     Condition(Span, Box<Expression>, Box<Expression>, Box<Expression>),
+    MakeStruct {
```
sharkdp, line 88: This is not super important, but I think it could be useful to change the naming a bit to make it easier to understand the code. I would distinguish between a *struct*, which is the type-level construct, and an *instance* of the struct, which is the actual object constructed at runtime. 

I think I would rename:
 * `MakeStruct` => `InstantiateStruct`
 * `AccessStruct` => `AccessField`

numbat/tests/interpreter.rs
```diff
@@ -52,9 +52,12 @@ fn expect_failure(code: &str, msg_part: &str) {
...
         let error_message = e.to_string();
         println!("{}", error_message);
-        assert!(error_message.contains(msg_part));
+        assert!(
+            error_message.contains(msg_part),
+            "Expected {msg_part} but got {error_message}"
+        );
     } else {
-        panic!();
+        panic!("Expected an error but but instead {code} did not fail");
```
sharkdp, line 60:
```suggestion
        panic!("Expected an error but the code '{code}' did not fail");
```

numbat/tests/interpreter.rs
```diff
@@ -52,9 +52,12 @@ fn expect_failure(code: &str, msg_part: &str) {
     if let Err(e) = ctx.interpret(code, CodeSource::Internal) {
         let error_message = e.to_string();
         println!("{}", error_message);
-        assert!(error_message.contains(msg_part));
+        assert!(
+            error_message.contains(msg_part),
+            "Expected {msg_part} but got {error_message}"
```
sharkdp, line 57:
```suggestion
            "Expected '{msg_part}' but got '{error_message}'"
```

book/src/structs.md
```diff
@@ -0,0 +1,24 @@
...
+struct Vector {
+  x: Length,
+  y: Length,
+}
+
+let v = Vector {x: 6m, y: 8m}
+
+# A function with a struct as a parameter
+fn euclidian_distance(a: Vector, b: Vector) =
+  sqrt(sqr(a.x - b.x) + sqr(a.y - b.y))
```
sharkdp, line 16: Let's maybe change this to 
```suggestion
  sqrt((a.x - b.x)² + (a.y - b.y)²)
```
to make it easier to read.

numbat/tests/interpreter.rs
```diff
@@ -478,7 +480,7 @@ fn test_prefixes() {
 fn test_parse_errors() {
     expect_failure(
         "3kg+",
-        "Expected one of: number, identifier, parenthesized expression",
+        "Expected one of: number, identifier, parenthesized expression, struct construction",
```
sharkdp, line 483:
```suggestion
        "Expected one of: number, identifier, parenthesized expression, struct instantiation",
```

book/src/structs.md
```diff
@@ -0,0 +1,24 @@
+# Struct types
+
+Numbat supports nominally typed records:
```
sharkdp, line 3: I think I would skip the "nominally typed" part, as I'm sure most users will not know what it means:
```suggestion
Numbat supports structs as compound data structures:
```

examples/numbat_syntax.nbt
```diff
@@ -109,3 +109,17 @@ assert_eq(1 ft, 12 in)           # Assert that two quantities are equal
...
+struct Foo {                        # defining a struct
+  foo: Current,
+  bar: Scalar
+}
+
+Foo {foo: 1A, bar: pi}              # constructing a struct
+
+let r: Foo = Foo {foo: 1A, bar: pi}
+
+r.foo                               # Field access is performed with `.field` notation
```
sharkdp, line 125: I would suggest a more concrete example, like:
```suggestion
# 10. Structs

struct Element {                 # Define a struct
    name: String,
    atomic_number: Scalar,
    density: MassDensity,
}

let hydrogen = Element {         # Instantiate it
    name: "Hydrogen",
    atomic_number: 1,
    density: 0.08988 g/L,
}

hydrogen.density                 # Access field of a struct
```

numbat/src/bytecode_interpreter.rs
```diff
@@ -165,6 +167,31 @@ impl BytecodeInterpreter {
...
+                    .sorted_by_key(|(n, _)| struct_info.fields.get_index_of(n).unwrap());
+
+                for (_, expr) in sorted_exprs.rev() {
+                    self.compile_expression_with_simplify(expr)?;
+                }
+
+                let struct_info_idx = self.vm.get_structinfo_idx(&struct_info.name).unwrap() as u16;
+
+                self.vm
+                    .add_op2(Op::BuildStruct, struct_info_idx, exprs.len() as u16);
```
sharkdp, line 186: Similar here: maybe rename the operation to `BuildStructInstance` or just `BuildInstance`

numbat/src/vm.rs
```diff
@@ -871,6 +901,8 @@ impl Vm {
                         Value::String(s) => s,
                         Value::DateTime(dt) => crate::datetime::to_rfc2822_save(&dt),
                         Value::FunctionReference(r) => r.to_string(),
+                        // TODO: better
```
sharkdp, line 904: Can we resolve this to-do?

numbat/src/vm.rs
```diff
@@ -946,6 +978,34 @@ impl Vm {
...
+
+                    self.stack.push(Value::Struct(struct_info, content));
+                }
+                Op::DestructureStruct => {
+                    let field_idx = self.read_u16();
+                    let content = self.pop();
+
+                    let Value::Struct(_meta, mut content) = content else {
+                        panic!("tried to destructure something which was not a struct (the typechecker should prevent this)");
+                    };
```
sharkdp, line 1004: For other `Value` types, we have `.unsafe_as_xxx()` methods. Let's maybe follow the same pattern here (see `impl Value` in `value.rs`)

numbat/src/value.rs
```diff
@@ -29,6 +33,7 @@ pub enum Value {
     DateTime(chrono::DateTime<chrono::FixedOffset>),
     FunctionReference(FunctionReference),
     FormatSpecifiers(Option<String>),
+    Struct(Arc<StructInfo>, Vec<Value>),
```
sharkdp, line 36: Again, naming: should this be called `StructInstance`?

numbat/src/name_resolution.rs
```diff
@@ -1,18 +1,76 @@
...
+    pub fn add(
+        &mut self,
+        name: String,
+        span: Span,
+        item_type: String,
+    ) -> Result<(), NameResolutionError> {
+        self.add_inner(name, span, item_type, false)
+    }
+
+    fn add_inner(
```
sharkdp, line 48: What does "inner" mean in this context? A small doc comment (or different name) might help here.

numbat/src/tokenizer.rs
```diff
@@ -621,7 +630,12 @@ impl Tokenizer {
                     self.advance();
                 }
 
-                if self.peek().map(|c| c == '.').unwrap_or(false) {
+                if self.peek().map(|c| c == '.').unwrap_or(false)
+                    && self
+                        .peek2()
+                        .map(|c| !is_identifier_start(c))
+                        .unwrap_or(true)
```
sharkdp, line 637: This logic seems non-trivial. It would be great if we could add tokenizer tests to make sure this works correctly in the future.

numbat/src/typed_ast.rs
```diff
@@ -848,6 +922,7 @@ mod tests {
         roundtrip_check("-3!");
         roundtrip_check("(-3)!");
         roundtrip_check("megapoints");
+        roundtrip_check("Foo {foo: 1 meter, bar: 1 second}");
```
sharkdp, line 925: Minor: I would personally go for a style with whitespace  (similar to struct definition):

```suggestion
        roundtrip_check("Foo { foo: 1 meter, bar: 1 second }");
```

numbat/src/value.rs
sharkdp, line 36: Why is the `StructInfo` behind an `Arc`?

numbat/src/typechecker.rs
```diff
@@ -297,6 +301,30 @@ pub enum TypeCheckError {
 
     #[error("Base units can not be dimensionless.")]
     NoDimensionlessBaseUnit(Span, String),
+
+    #[error("Unknown struct '{1}")]
+    UnknownStruct(Span, String),
+
+    #[error("Unknown field {2} of struct {3}")]
```
sharkdp, line 308: I prefer to always add quotes around variable parts of error messages, because they can otherwise lead to confusing sentences with field/struct names that do not immediately stand out:
```suggestion
    #[error("Unknown field '{2}' of struct '{3}'")]
```

numbat/src/typechecker.rs
```diff
@@ -297,6 +301,30 @@ pub enum TypeCheckError {
...
+    UnknownFieldOfStruct(Span, Span, String, String),
+
+    #[error("Duplicate field {2} in struct definition")]
```
sharkdp, line 311:
```suggestion
    #[error("Duplicate field '{2}' in struct definition")]
```

numbat/src/typechecker.rs
```diff
@@ -297,6 +301,30 @@ pub enum TypeCheckError {
...
+    DuplicateFieldInStructDefinition(Span, Span, String),
+
+    #[error("Duplicate field {2} in struct construction")]
```
sharkdp, line 314:
```suggestion
    #[error("Duplicate field '{2}' in struct instantiation")]
```

numbat/src/typechecker.rs
```diff
@@ -297,6 +301,30 @@ pub enum TypeCheckError {
...
+    DuplicateFieldInStructConstruction(Span, Span, String),
+
+    #[error("Accessing field {2} of non struct type {3}")]
```
sharkdp, line 317:
```suggestion
    #[error("Accessing field '{2}' of non struct type '{3}'")]
```

numbat/src/typechecker.rs
```diff
@@ -297,6 +301,30 @@ pub enum TypeCheckError {
...
+    AccessingFieldOfNonStruct(Span, Span, String, Type),
+
+    #[error("Accessing unknown field {2} of struct {3}")]
```
sharkdp, line 320:
```suggestion
    #[error("Accessing unknown field '{2}' of struct '{3}'")]
```

numbat/src/typechecker.rs
```diff
@@ -297,6 +301,30 @@ pub enum TypeCheckError {
...
+    AccessingUnknownFieldOfStruct(Span, Span, String, Type),
+
+    #[error("Missing fields from struct construction")]
+    MissingFieldsFromStructConstruction(Span, Span, Vec<(String, Type)>),
```
sharkdp, line 324:
```suggestion
    #[error("Missing fields in struct instantiation")]
    MissingFieldsInStructInstantiation(Span, Span, Vec<(String, Type)>),
```

numbat/src/typechecker.rs
```diff
@@ -416,6 +444,12 @@ fn evaluate_const_expr(expr: &typed_ast::Expression) -> Result<Exponent> {
                 "binary operator for datetimes",
             ))
         }
+        e @ typed_ast::Expression::MakeStruct(_, _, _) => Err(
+            TypeCheckError::UnsupportedConstEvalExpression(e.full_span(), "make struct"),
```
sharkdp, line 448:
```suggestion
            TypeCheckError::UnsupportedConstEvalExpression(e.full_span(), "instantiate struct"),
```

numbat/src/typechecker.rs
```diff
@@ -416,6 +444,12 @@ fn evaluate_const_expr(expr: &typed_ast::Expression) -> Result<Exponent> {
...
+        ),
+        e @ typed_ast::Expression::AccessStruct(_, _, _, _, _, _) => Err(
+            TypeCheckError::UnsupportedConstEvalExpression(e.full_span(), "access struct"),
```
sharkdp, line 451:
```suggestion
            TypeCheckError::UnsupportedConstEvalExpression(e.full_span(), "access field of struct"),
```

numbat/src/typechecker.rs
```diff
@@ -661,10 +697,35 @@ impl TypeChecker {
...
+                        .into_iter()
+                        .map(|(n, (s, t))| {
+                            (
+                                n.to_owned(),
+                                (s.clone(), apply_substitutions(t, substitute, substitutions)),
+                            )
+                        })
+                        .collect(),
+                }),
+                type_ => type_.clone(),
```
sharkdp, line 724: Can this code even trigger? We can't have structs with generic parameters, can we?

numbat/src/typechecker.rs
```diff
@@ -1094,6 +1155,107 @@ impl TypeChecker {
...
+                fields,
+            } => {
+                let fields_checked = fields
+                    .iter()
+                    .map(|(_, n, v)| Ok((n.to_string(), self.check_expression(v)?)))
+                    .collect::<Result<Vec<_>>>()?;
+
+                let Some(struct_info) = self.structs.get(name) else {
+                    return Err(TypeCheckError::UnknownStruct(*ident_span, name.clone()));
+                };
```
sharkdp, line 1171: Wow, I never realized let-else was a thing. How could I live without it?

numbat/src/vm.rs
```diff
@@ -946,6 +978,34 @@ impl Vm {
...
+
+                    let value = content.remove(field_idx as usize);
```
sharkdp, line 1006: The `.remove` call is a bit confusing when reading this, but I assume it's because this gets us a value, not just a reference?

`.remove` is deprecated, apparently? https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.remove

sharkdp: .remove moves all elements after the accessed index, i.e. it's O(N) in the number of fields N. I don't think we need to worry about performance just yet, but it might be better to use `.swap_remove` or `content[field_idx as usize].clone()`.

sharkdp (changes requested):
Ok, I finished my review. Sorry for taking a bit longer. This looks really great! Most comments are just minor naming/clarification things. Let me know if you need any help with finishing this. I'm really looking forward to integrating this.

sharkdp:
Another small thing that I found:

This leads to an error:
```
dimension A
fn foo<A>(a: A) -> Scalar = 1
```
but this doesn't:
```
struct A {}
fn foo<A>(a: A) -> Scalar = 1
```

sharkdp (approved):
I made some changes myself. Hope that's okay. There are two remaining open comments/questions, which I do not see as blocking issues that would prevent me from merging this. Maybe we can come back to them at some point.

simmsb:
Nice, thanks for taking the time to make those changes :)

sharkdp:
FYI: we're already making use of structs in a new feature: #431

PR #419, opened 2024-04-29
Bump iana-time-zone to no longer use a yanked package

RossSmyth:
[iana-time-zone 0.1.59 is yanked](https://crates.io/crates/iana-time-zone/0.1.59) so this removes that as a dep.

Thanks for this awesome tool!


sharkdp:
Thank you

PR #420, opened 2024-04-29
Implement 'Never' type (!)

sharkdp:
This allow us to use `error(…)` in almost all contexts. In particular, it makes the following code compile for all `X`:

```hs
if <some precondition>
  then X
  else error("please make sure <some precondition> is met")
```

@Bzero This one is for you :smile:


Bzero:
Thanks @sharkdp 😄!

Issue #261, opened 2023-11-20
Add a list datatype


sharkdp:
- What kind of syntax do we want for lists?
- How do we want to add lists in the type checker? What should be the syntax for defining list types? `List<T>`?

irevoire:
What should be the pattern to work with it?

I see two big paths personally;
- Should it be iterative, and thus, should we add mutability and loop to the language
- Or should it be functional, and thus there are multiple ways to implement it in any case, we'll need a new syntax for anonymous functions:
  - Like Python, with global functions everywhere: `count(filter(map(list, fn (price) price * 50%), fn (price) price < 10))` (I hate it very much)
  - Like most iterative sane language with _methods_, but then, should we implement some kind of objects? `list.map(fn (price) price * 50%).filter(fn (price) price < 10).count()`
  - Like functional language with some kind of symbol to chains function call (here I used OCaml cool `|>` symbol); `list |> map fn (price) price * 50% |> filter fn (price) price < 10 |> count`

mhmdanas:
> * Like functional language with some kind of symbol to chains function call (here I used OCaml cool `|>` symbol); `list |> map fn (price) price * 50% |> filter fn (price) price < 10 |> count`

We could maybe do this with `//`, but extending it to accept partial applications of functions.

sharkdp:
> What should be the pattern to work with it?

Those are good questions, and I currently don't have an answer. I certainly see Numbat as a kind of functional programming language. As for the lambda syntax, I personally prefer the Rust style `|price| price × 50%` compared to Python (`lambda price: price × 50%`) or Haskell `\price -> price × 50%`. As for the chaining/filtering syntax, I need to think about it.

mhmdanas:
https://github.com/sharkdp/numbat/pull/398 also brings up the question of whether we should also add (heterogeneously-typed) tuples and not just lists. Thoughts?

> I personally prefer the Rust style |price| price × 50% compared to Python (lambda price: price × 50%) or Haskell \price -> price × 50%.

I really dislike the Python syntax personally (also doesn't really fit in with the rest of the Numbat syntax), but a bit more on the fence about Rust-style vs Haskell-style syntax. Would the Rust-style syntax possibly introduce parsing difficulties if we decide to add `|` as a bitwise OR later?

irevoire:
Since you can’t use the `|` as a unary prefix operator, I don’t think it’ll cause any issues.

sharkdp:
Some cool things we could do with lists:
[...]

PR #432, opened 2024-05-16
List datatype

sharkdp:
This is a basic implementation of lists for Numbat. We construct lists using `[…]` notation and the type of lists is specified as `List<T>` where `T` can be any type.

```
>>> [1, 2, 3]

    = [1, 2, 3]    [List<Scalar>]

>>> [1m, 30cm]

    = [1 m, 30 cm]    [List<Length>]

>>> [["foo", "bar"], ["baz"]]

    = [[foo, bar], [baz]]    [List<List<String>>]

```

Operationally, they work just fine. For example, we can add the functional-programming style primitives `head`, `tail` and `cons`:

```
fn head<A>(xs: List<A>) -> A
fn tail<A>(xs: List<A>) -> List<A>
fn cons<A>(x: A, xs: List<A>) -> List<A>
```

and they work as expected:

```
>>> cons(0, [1, 2, 3])

    = [0, 1, 2, 3]    [List<Scalar>]
```

We can even write some useful functions like
```
>>> fn generate(n: Scalar, f: Fn[() -> Scalar]) -> List<Scalar> =
  if n == 1
    then [f()]
    else cons(f(), generate(n - 1, f))
```

```
>>> generate(10, random)

    = [0.372604, 0.123614, 0.297907, 0.966904, 0.296039, 0.351666, 0.923613, 0.0140005, 0.254682, 0.518355]    [List<Scalar>]
```

Type checking of list construction also works fine:

```
>>> [1 m, 40 cm, 3 kg, 2 m]
error: while type checking
  ┌─ <input:3>:1:2
  │
1 │ [1 m, 40 cm, 3 kg, 2 m]
  │  ^^^         ---- Mass
  │  │            
  │  Length
  │
  = Incompatible types in list: expected 'Length', got 'Mass' instead
```

However, there are a number of open questions / issues regarding the handling of lists in the type checker:

- We don't have polymorphic *values* yet. So we can't type the empty list `[]`, which should have the type `∀D. List<D>`. We have a similar problem with typing `0`, `NaN` and `inf`, see #37.
- Almost all useful functions on lists will be generic over the element type. Very basic things (like len, head, tail, cons) work fine when implemented via the FFI. But if we want to use those functions to implement `map`, `filter`, etc., we are running into type inference problems (already reported as https://github.com/sharkdp/numbat/issues/166).
- Polymorphism in Numbat only works for *dimension types* (`Scalar`, `Length`, `Length^2`, `Length / Time`, etc). But so far, we can't be generic over *all* types, which means we can not give a type to a function that works on all lists, only on lists of arbitrary dimension types. So we can create things like `List<Bool>`,`List<String>` or `List<List<Length>>`, but we can't run `cons`, `head`, `tail` on them.

related ticket: #261

### DIFF
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -1,0 +1,3 @@
+fn head<A>(xs: List<A>) -> A
+fn tail<A>(xs: List<A>) -> List<A>
+fn cons<A>(x: A, xs: List<A>) -> List<A>
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -1,3 +1,55 @@
+use core::scalar
+
+fn len<A>(xs: List<A>) -> Scalar
 fn head<A>(xs: List<A>) -> A
 fn tail<A>(xs: List<A>) -> List<A>
 fn cons<A>(x: A, xs: List<A>) -> List<A>
+
+fn is_empty<A>(xs: List<A>) -> Bool = len(xs) == 0
+
+# We definitely want to make the following functions generic, but this is not yet possible:
+# Also, we want the base case to be the empty list, but this is also not yet possible.
+
+fn generate(n: Scalar, f: Fn[() -> Scalar]) -> List<Scalar> =
+  if n == 1
+    then [f()]
+    else cons(f(), generate(n - 1, f))
+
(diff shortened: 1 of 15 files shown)


sharkdp:
see follow up: #443

Issue #421, opened 2024-05-01
Add info for functions


eminence:
A while ago we added support for getting [info about units and variables](https://github.com/sharkdp/numbat/pull/267), 

Now that we're getting some [interesting functions](https://github.com/sharkdp/numbat/pull/416), let's extend the `info` mechanism (and related decorators) to support functions

sharkdp:
Yes — that would be great to have. Thank you for creating the ticket. It would also allow us to auto-generate the list of functions (https://numbat.dev/doc/list-functions.html) and to enrich it with more information.

PR #435, opened 2024-05-21
Function decorators

Bzero:
This PR addresses #421, enables decorators for functions and extends the `info` command to functions.

An additional `@description()` decorator is introduced which can be used on functions, as well as variables and units. If applied several times on the same function/variable/unit the descriptions are joined.

There are two caveats to the current implementation: 
- To add aliases of functions a copy of the function is made for each alias. This is not ideal as it wastes some resources and prohibits the use of aliases for FFI functions. A cleaner and more efficient way to handle aliases in the future  that would also enable using aliases for FFI functions could be to use a 'function pointer'.
- `info some_function` displays the signature of the function, however this currently only works for non-generic functions.

### DIFF
--- a/numbat/modules/datetime/functions.nbt
+++ b/numbat/modules/datetime/functions.nbt
@@ -6,7 +6,7 @@ use units::si
 fn now() -> DateTime
 
 @url("https://numbat.dev/doc/date-and-time.html")
-@description("Parses a string (date and time) into a DateTime object.")
+@description("Parses a string (date and time) into a DateTime object. See https://numbat.dev/doc/date-and-time.html#date-time-formats for an overview of the supported formats.")
 fn datetime(input: String) -> DateTime
 
 @url("https://numbat.dev/doc/date-and-time.html")
--- a/numbat/modules/datetime/functions.nbt
+++ b/numbat/modules/datetime/functions.nbt
@@ -1,26 +1,56 @@
 use core::strings
 use units::si
 
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Returns the current date and time.")
 fn now() -> DateTime
 
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Parses a string (date and time) into a DateTime object.")
 fn datetime(input: String) -> DateTime
+
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Formats a DateTime object as a string.")
 fn format_datetime(format: String, input: DateTime) -> String
 
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Returns the users local timezone.")
 fn get_local_timezone() -> String
+
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Returns a timezone conversion function, typically used with the conversion operator.")
 fn tz(tz: String) -> Fn[(DateTime) -> DateTime]
+
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Timezone conversion function targeting the users local timezone (datetime -> local).")
 let local: Fn[(DateTime) -> DateTime] = tz(get_local_timezone())
+
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Timezone conversion function to UTC.")
 let UTC: Fn[(DateTime) -> DateTime] = tz("UTC")
 
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Converts a DateTime to a UNIX timestamp.")
 fn unixtime(input: DateTime) -> Scalar
+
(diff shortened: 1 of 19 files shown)


numbat/modules/datetime/functions.nbt
```diff
@@ -1,26 +1,56 @@
 use core::strings
 use units::si
 
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Returns the current date and time.")
 fn now() -> DateTime
 
+@url("https://numbat.dev/doc/date-and-time.html")
+@description("Parses a string (date and time) into a DateTime object.")
```
sharkdp, line 9: Maybe add 
```suggestion
@description("Parses a string (date and time) into a DateTime object. See https://numbat.dev/doc/date-and-time.html#date-time-formats for an overview of the supported formats.")
```
(or use that URL inside `@url`).

numbat/src/lib.rs
```diff
@@ -394,6 +418,57 @@ impl Context {
...
+
+            if let Ok((statements, _)) = self.interpret(keyword, CodeSource::Internal) {
+                let signature = match Statement::as_expression(&statements[0]) {
+                    Some(expression) => expression
+                        .get_type()
+                        .to_readable_type(self.dimension_registry()),
+                    None => m::empty(),
+                };
+                help += m::text("Signature: ") + signature + m::nl();
+            }
```
sharkdp, line 443: This is a nice hack, but it might be better/easier to add a querying function to `TypeChecker` which has all the information we need (in `TypeChecker::function_signatures`). That could be called here using something like `self.typechecker.get_function_signature(keyword)`. But then you would need to turn `FunctionSignature` into something that you can render, so… let's maybe stick with the current solution for now. Things are about to change significantly in the type checker, which might also affect how function signatures are stored.

sharkdp:
> This PR addresses #421, enables decorators for functions and extends the `info` command to functions.
> 
> An additional `@description()` decorator is introduced which can be used on functions, as well as variables and units. If applied several times on the same function/variable/unit the descriptions are joined.

Very nice work!

> * To add aliases of functions a copy of the function is made for each alias.

Do we actually *want* to have aliases for functions? It's not a great (but somewhat useful) solution for units and constants, as it clutters the namespace. It is nice to be able to write `1 meter`, and also `2 meters`. But for functions, it might make sense to avoid the issues all together by disallowing `@aliases`. The only real use case I see right now would be `log` vs `ln` (which are two separate definitions for now). What do you think?

Bzero:
Thanks! 

> Do we actually _want_ to have aliases for functions? It's not a great (but somewhat useful) solution for units and constants, as it clutters the namespace. It is nice to be able to write `1 meter`, and also `2 meters`. But for functions, it might make sense to avoid the issues all together by disallowing `@aliases`. The only real use case I see right now would be `log` vs `ln` (which are two separate definitions for now). What do you think?

I am not too concerned about cluttering the namespace since I would not consider it a particularly good idea to call a function or variable `meters` anyways so I my eyes aliases are a perfectly fine solution for units.

For functions (and variables) I think aliases are not really important but can be handy, for example to have a long and a short name. I am however not really sure if it is useful to have an alias decorator since it is already possible to do

`let my_alias = some_function`

which has the advantage that it actually results in a reference and not in a copy of `some_function`.

numbat/src/lib.rs
Bzero: Yes, I was not really happy about this part of the code either. I will see if I can implement a method to query and print `TypeChecker::function_signatures`, thanks for that hint!

sharkdp:
In light of the upcoming changes to the type checker, I would appreciate if we could get rid of aliases for functions.

numbat/src/bytecode_interpreter.rs
```diff
@@ -40,7 +41,7 @@ pub struct BytecodeInterpreter {
     // Maps names of units to indices of the respective constants in the VM
     unit_name_to_constant_index: HashMap<String, u16>,
     /// List of functions
-    functions: HashMap<String, bool>,
+    functions: HashMap<String, (bool, DecoratorMetadata)>,
```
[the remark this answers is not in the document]
Bzero: Yes, without aliases there is little reason to store it here. I moved the function metadata into the existing hashmap in the type checker.

numbat/src/lib.rs
Bzero: I rewrote this part, the function signature is now directly extracted from `TypeChecker::FunctionSignature` and now also works for generic functions.

Bzero:
Aliases are now disallowed for functions and the function signature is directly extracted from the type checker which enables them for generic functions too.

sharkdp (approved):
Thank you!

Issue #145, opened 2023-09-08
Numerical derivatives, integrals


sharkdp:
It should actually be fairly easy to implement a simple version of this (if we disregard all of the real world numerical problems that can arise). The major blocker for this is the fact that we don't have function types and corresponding function objects (#159), so we cannot simply write something like `diff(sqrt, 1.0)`

With a hypothetical syntax for function types `Fn[(X1, X2, X3, …) -> X_return]`, we could even implement numerical diff in Numbat itself using the new builtin `unit_of` function:

```rs
let eps = 1e-10

fn diff<X, Y>(f: Fn[(X) -> Y], x: X) -> Y / X =
  (f(x + eps · unit_of(x)) - f(x)) / (eps · unit_of(x))
```

Obviously, we would like to have something a little bit more sophisticated for production though.

For a concrete function, this already works:
```rs
fn diff_sqrt<X>(x: X) -> X^(-1/2) = (sqrt(x + eps · unit_of(x)) - sqrt(x)) / (eps · unit_of(x))

assert_eq(diff_sqrt(1), 1/2, 1e-5)
assert_eq(diff_sqrt(2), 1/2 × 2^-0.5, 1e-5)
```

sharkdp:
With #159 merged, we can now do this:

```
fn diff(f: Fn[(Scalar) -> Scalar], x: Scalar) -> Scalar =
    (f(x + eps) - f(x)) / eps

assert_eq(diff(log, 2.0), 1 / 2, 1e-5)
assert_eq(diff(sin, 0.0), 1.0, 1e-5)
```

We still can't do the generic version though, because our generics handling in the type checker is not yet powerful enough.

sharkdp:
The generic version of this is now also supported (see #451), so we can do things like
```
use numerics::diff

fn dist(t: Time) -> Length = 0.5 g0 t^2
fn velocity(t: Time) -> Velocity = diff(dist, t)

assert_eq(velocity(2.0 s), 2.0 s × g0, 1e-3 m/s)
```

Issue #66, opened 2023-02-15
Add sums and products


sharkdp:
Add support for `sum` and `product`, just like in Insect. Potentially with a different syntax. Insect can do this:
```
>>> sum(x², x, 1, 10)

   = 385
```

It would be fantastic if we could implement this in the language itself. However, this would require a whole new set of features. The first argument is actually something like a lambda: `λx: x²`. But we would also need recursion (#32) [Edit: now implemented] and probably support for a whole new datatype (lists) [Edit: now also implemented] to properly support this. 

If we had this, we could also implement other feature requests easily:
- https://github.com/sharkdp/insect/issues/396
- https://github.com/sharkdp/purescript-quantities/pull/65
- Statistics functions from https://github.com/sharkdp/numbat/issues/144#issue-1888030344

sharkdp:
This is now possible, but will be much nicer once we have closures (and maybe lambdas), see #347. Example application from https://github.com/sharkdp/insect/issues/396:

![image](https://github.com/sharkdp/numbat/assets/4209276/61206f44-db60-4b1a-8b9b-1328b36160e2)

archisman-panigrahi:
How about having `parallel_resistors` as a default, inbuilt function?

sharkdp:
> How about having `parallel_resistors`, `serial_capacitors` and `parallel_inductors` as default, inbuilt functions? This is something which people may need regularly.

Sure, why not. We could have a `extra::electrostatics` (or similar) module for that.

PR #443, opened 2024-05-29
Full type inference, list datatype

### LINKED ISSUE #29: Functions: full type inference
It would be very cool if we could infer function parameter types and return types. For example, given hypothetical code like
```
fn g(t: Time) -> Speed = …

fn f(d, t) = (d + 1 meter) / g(t)
```
we could easily infer from `d + 1 meter` that `d: Length`, and from `g(t)` that `t: Time`. Furthermore, we could compute the return type to be `Length / Speed =Time`, leading to an inferred type of
```
fn f(d: Length, t: Time) -> Time
```

This also relates to generics (#25), as some (most?) functions would probably have inferred generic types. For example, in the absence of *any* annotations,
```
kineticEnergy(m, v) = 1/2 * m * v^2
```
would have an inferred type of
```
kineticEnergy<T0, T1>(m: T0, v: T1) -> T0 * T1^2 = 1/2 * m * v^2
```

## Possible constraints
* **Addition, Subtraction, Conversion**: these binary operators act as constraints on their operands. For example, an addition like
  ```
  expr1 + expr2
  ```
  imposes a simple `[expr1] == [expr2]` constraint. Similarly, `expr1 -> expr2` also forces `expr1` to have the same dimension as `expr2`.
* **Function application:** A function call like 
  ```
  f(expr1, expr1, …, exprN)
  ```
  imposes N constraints, one for each argument: `[exprK] == [parameter k of f]`. A simple example would be something like
  ```
  fn normalize_len(l: Length) -> Length = …
  fn f(d, alpha) = normalize_length(d) * sin(alpha)
  ```
  where we could use the function calls to easily get the types of `d` and `alpha`.
* **Exponentiation**: Since the exponent in an expression like
  ```
  expr1 ^ expr2
  ```
  always needs to be a scalar, we can infer `[expr2] == 1`. In general, we can *not* assume that `expr1` is also a scalar (`meter^2` is perfectly fine, for example). So there is also no constraint for `expr1`.

## Formalization

Given a (partially annotated) function definition like one of these (`Cx` is a concrete type):
```
f(x0, x1, x2, …) = …
f(x0, x1: C1, x2, …) = …
f(x0, x1, x2, …) -> C_ret = …
f(x0, x1: C1, x2, …) -> C_ret = …
…
```
We start by filling in blanks with free type variables `T0, T1, …` and `T_ret`:
```
f(x0: T0, x1: T1, x2: T2, …) -> T_ret = …
f(x0: T0, x1: C1, x2: T2, …) -> T_ret = …
f(x0: T0, x1: T1, x2: T2, …) -> C_ret = …
f(x0: T0, x1: C1, x2: T2, …) -> C_ret = …
…
```

Next, we walk the AST of the right hand side expression and record all constraints that we discover (see above). For example, given  a declaration like `f(x0, x1, x2) = sqrt((x0 · x2² + second) · second / x1²) -> meter`, we would identify:
* `T0 · T2^2 == Time` (from `x0 · x2² + second`)
* `((T0 · T2^2) · Time / T1²)^(1/2) == T0^(1/2) · T1 · T2 · Time^(1/2) == Length` (from the `->` conversion operator)

These constraints would then have to be converted to a system of linear equations on the exponents. 

```
  1 * exponents0 + 0 * exponents1 + 2 * exponents2 == (1, 0)
1/2 * exponents0 + 1 * exponents1 + 1 * exponents2 == (1, -1/2)
```
where `exponents_i` and the right hand sides would be vectors with one rational number entry for each of the *D* physical dimensions. In total, this is a set of N_constraints · D equations, where the left-hand side will be the same regardless of the physical dimension.

#### Example 1

```
g(t: Length) -> Length = …
f(x, y, z) = g(sqrt(x)/y) * (y + 2 * second) * 2^z
```

* From `2^z`, we infer `[z]: 1`.
* From `y + 2 * second`, we infer `[y]: Time`
* From `g(sqrt(x)/y)`, we infer `sqrt(x)/y: Length`, i.e. `sqrt(x): Length · Time` or `x: Length² · Time²`.
* Finally, there return type can be computed as `Length · Time`

```
f(x: Length² · Time², y: Time, z: 1) -> Length · Time = g(sqrt(x)/y) * (y + 2 * second) * 2^z
```

#### Example 2
```
f(x) = x + x^2
```
should be inferred as
```
f(x: Scalar) -> Scalar
```

#### Example 3
(from https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-391.pdf)

```
f(x,y,z) = x*x + y*y*y*y*y + z*z*z*z*z*z
```
should ideally be inferred as
```
f<T>(x: T^15, y: T^6, z: T^5) -> T^30
```

### Underdetermined

For a function definition like
```
f(x, y) = x + y^3
```
we could infer one of those (equivalent) types:
```
f(x: T, y: T^(1/3)) -> T = x + y^3
f(x: T^3, y: T) -> T^3 = x + y^3
```

### Overdetermined

A function like
```
f(x) = (x + meter)/(y + second)
```
should lead to a type check error.

### LINKED ISSUE #37: Zero, NaN, inf should be polymorphic

### LINKED ISSUE #166: Can not infer type parameter when calling generic function from generic function

### LINKED ISSUE #261: Add a list datatype

### LINKED ISSUE #324: assert_eq should work for all types (that are comparable)
We can do
```
assert("{1+2}" == "3")
```
but I'd like to be able to
```
assert_eq("{1+2}", 3)
```
in order to get better assertion messages.

sharkdp:
This PR adds proper type inference to Numbat. It also makes the type system more powerful (with respect to generics). A few highlights:

- Full Hindley-Milner-style type inference for arbitrary Numbat programs. Including inference for dimension types (see #29)
- Generics for arbitrary types, not just dimension types
- Proper support for `List<T>` for all types `T` (see #261). Together with proper generics handling (see below), this should allow us to implement #66
- Polymorphic values for `0`, `NaN` and `inf` (see #37)  
- Type inference works for recursive functions, too. We can get rid of the restriction that the return type needs to be specified.
- Fixes #166
- Allows us to implement generic numerical differentiation (#145) inside Numbat using:
  ```
  let eps = 1e-10

  fn diff<X, Y>(f: Fn[(X) -> Y], x: X) -> Y / X =
    (f(x + eps · unit_of(x)) - f(x)) / (eps · unit_of(x))
  ```
  Example: say you have a function from `Time -> Length`, then you will get a `Velocity` as a result from `diff`.

Other changes:
- Support for variadic functions has been removed. We can cover all use cases with lists
- The `Never`/`!` data type has been removed. Instead, the `error` function now has type `forall A. String -> A` 

Open challenges:
- [x] There is a severe regression concerning error messages for rather basic things like `kg m / s^2 + kg m^2` where we will now get a constraint solver error instead of our nice incompatible-dimensions error.
- [x] Properly integrate `struct`s. Field access is currently broken when accessing structs in functions. This could be fixed by requiring annotations for structs. What type should `fn f(x) = x.foo` have otherwise? The only reasonable option would be to go back to structurally typed records and do something similar to Haskell (or probably better, similar to PureScript).
- [x] `DateTime` arithmetic is currently broken. The problem is with the overloaded `+`/`-` operators. If we have something like `x - y` with yet-unknown types, we would like to say: either, `x` and `y` can be of the same dimension type OR `x` could be a `DateTime` and `y` a `Time` or a `DateTime`. This kind of either-or constraint can not be formulated currently. The way this is solved in other Hindley-Milner languages is through type classes / trait bounds. We would need to add (multi-parameter?) type classes to model things like `forall D: Dim. Sub(D, D)`, `Sub(DateTime, DateTime)` and `Sub(DateTime, Time)`.
- [x] Properly treat `Dim` constraints and add a new syntax for them, e.g. `fn sqrt<D: Dim>(x: D^2) -> D = …`.

To do:
- [x] Finish type inference experiments
- [x] Remove variadic functions
- [x] `error("hi") + true` should be a type check error
- [x] Fix inferred type for e.g. `fn test(f, xs) = f(head(xs))`
- [x] Completely remove Never type
- [x] Replace Never type with `forall D. D` or `forall D: Dim. D`
- [x] Add tests from #253
- [x] Does something like `1 m + 0 * 1 s` work?
- [x] Fix `==`/`!=` operator.
- [x] Addition/subtraction needs `Dim` constraints (and possibly other operators as well): `fn f(x)=x+true`, `fn f(x,y)=x+y`
- [x] Arity needs to be checked properly for callables
- [x] `fn f(x, y) = x^y`, fix `todo!(…)`
- [x] Test for `fn f(x) = x + true`
- [x] Add tests for type inference (see typechecker experiments)
- [x] Fix env polution by fn parameter identifiers
- [x] Bring back commented-out tests
- [x] Improve constraint solver error messages
- [x] Improve substitution error messages (how can they trigger?)
- [x] Fix all TODOs
- [x] Bring back `readable_type`s (see TODOs in bytecode_interpreter and type_scheme.rs)
- [x] Bring back printing of function signatures when using `info`
- [x] Remove to_concrete_type
- [x] Remove instantiate_for_printing?
- [x] Reset `deploy.sh`

Follow ups:
- [x] Removing `log`ging dependencies?
- [ ] Update documentation, update https://numbat.dev/articles/intro.html
- [ ] `try_trivial_resolution` can easily analyze the constraint structure further to discard constraints like `List<T0> ~ Bool`, which will yield better error messages in return
- [ ] Equality on function types should be forbidden. Maybe we need to introduce an `Eq` type class.
- [ ] Better error message for sth. like `true(2)` 
- [ ] Lists are embarrassingly slow. I'm sure there are some long hanging fruit...
- [ ] Check if we can reduce the number of constraints: (1) by cancelling out equal constraints and (2) by not adding as many constraints. for example, in `x + y`, we need `type(x) ~ type(y)` and `IsDim(type(x))`, but we could potentially save the `IsDim(type(y))`. On the other hand, that would result in worse error messages.
- [ ] Re-read http://typesatwork.imm.dtu.dk/material/TaW_Paper_TypesAtWork_Kennedy.pdf
- [ ] Implement `mean`, `stddev`, etc in Numbat
- [ ] For all (generic) functions that we have, check that we can (1) remove all type annotations and (2) get the same type or a more general type.
- [ ] Give `quadratic_equation` a `List<B/A>` return type.
- [ ] Something like `fn foo<D1, D2>(x: D1*D2) = 1` should yield a "missing Dim bound" constraint, because the type parameters are used type annotation dimension expression
- [ ] Add tests for type inference with partial annotations
- [ ] Better error message for `fn f<A: Dim>(x: A) -> A = 2x; f(true)`
- [ ] When quantifying a function type with annotations, use the user-specified identifiers
- [x] Disallow generic types in dimension / unit definitions (`unit magic = 0; 1 m -> magic`), allow them in variable definitions
- [ ] Improve error messages for callables, see `function_types_in_argument_position` test
- [ ] Fix regression concerning "readable types". For example: `let v = 1m/s` would previously print `let v: Velocity = …`. Similarly, it would be cool if we would annotate `v` in `fn f(v)=v+1 knot` with a `Velocity` type
- [ ] Bring numerical diff functionality to the prelude (#145)

Cool stuff:

* Generic list functions, see `numbat/modules/core/lists.nbt`.
* Infer type of `map` or `fold`. Type something like
   ```
   fn fld(f, acc, xs) = if xs == [] then acc else fld(f, f(acc, head(xs)), tail(xs))
   ```
   and see the right type of fold come out
* `diff`: Check out `examples/numerical_diff.nbt` on this branch
* Type something like `fn f(x) = m² / x + s² * x` or `fn f(x, y) = x + y^2` and see the correct type being inferred.

closes #29 
closes #261 
closes #324 
closes #37
closes #166


numbat/modules/extra/algebra.nbt
```diff
@@ -1,10 +1,9 @@
...
   (-b + sign × sqrt(b² - 4 a c)) / 2 a
 
-fn quadratic_equation<A2, B2>(a: A2, b: B2, c: B2²/A2) -> String =
-  # TODO: Replace `a == 0 a` check after implementing https://github.com/sharkdp/numbat/issues/37.
-  if a == 0 a
-  then if b == 0 b then if c == 0 c then "infinitely many solutions" else "no solution" else "x = {-c / b}"
-  else if b² < 4 a c then "no real-valued solution" else if b² == 4 a c then "x = {-b / 2 a}" else "x₁ = {_qe_solution(a, b, c, 1)}; x₂ = {_qe_solution(a, b, c, -1)}"
+fn quadratic_equation<A: Dim, B: Dim>(a: A, b: B, c: B²/A) -> String =
+  if a == 0
+    then if b == 0 then if c == 0 then "infinitely many solutions" else "no solution" else "x = {-c / b}"
```
hamirmahal, line 8: Nice!

Issue #452, opened 2024-06-04
Generic structs


sharkdp:
With the recent changes to generics in #443, it might not be too hard to support generic structs. As in:
```
struct Pair<A, B> {
  first: A,
  second: B,
}
```
or maybe
```
struct Vec<D> {
  x: D,
  y: D,
  z: D,
}
```

sharkdp:
For the implementation of [this function](https://github.com/sharkdp/numbat/blob/df1a6369c7bee76214c03f8a4b148f9ff0757380/numbat/modules/core/mixed_units.nbt#L6), I wanted to have something like
```
struct MixedUnitComponent<D: Dim> {
  unit: D,
  name: String,
}
```
in order to pass just one `List<MixedUnitComponent<D>>` argument around instead of two synchronized lists `List<D>` and `List<String>`.

sharkdp:
Early draft: https://github.com/sharkdp/numbat/pull/514

sharkdp:
Another example where this is needed: #529

mihirmanna:
Any updates on this? I ran into this issue while trying to set up something similar to the Vec<D> example at the very top:
```
struct Vec<D> {
  x: D,
  y: D,
  z: D,
}
```

sharkdp:
No update yet, no

PR #464, opened 2024-06-11
Basic plotting support

sharkdp:
This is a draft for adding basic plotting support to Numbat. Contrary to the first attempt in #147, we now implement the basic data preprocessing step in Numbat itself, using the new struct and `List` data types. We can now also implement this as a normal function, instead of having to introduce a new command.

We can now write something like:
```rs
let A0 = 3 cm
let ω = 20 Hz
let λ = 1.5 s

fn amplitude(t) = A0 exp(-t/λ) cos(ω t)

line_plot(amplitude, 0 s, 8 s) |>
  xlabel("Time") |>
  ylabel("Amplitude") |>
  show
```
and get the following plot. The `xlabel`/`ylabel` calls are optional. Note how the units here are inferred and added to the axis descriptions:

image

Similarly, we can create a simple bar chart:

```rs
let names = ["Iron", "Copper", "Silver", "Gold", "Platinum"]
let elements = map(element, names)

fn density(e: ChemicalElement) -> MassDensity = e.density

let densities = map(density, elements)

bar_chart(densities) |>
  xlabels(names) |>
  value_label("Density") |>
  show
```

image

Open points:

- This would be much more powerful if we had #452. This would allow us to rewrite `line_plot(amplitude, 0 s, 8 s)` into `line_plot(amplitude) |> xlim(0 s, 8 s)`, for example. It would also allow us to implement `ylim(…, …)`. We could also store the function inside the `LinePlot` struct instead of storing all the data points. We would only convert to data points before passing it to the FFI
- It would be nice to be able to move the `|>` reverse-apply operators in the example above to the next line. To have it look like
  ```
  line_plot(amplitude, 0 s, 8 s)
    |> xlabel("Time")
    |> ylabel("Amplitude")
    |> show
  ```


eminence:
For CLI usage, I wonder if we could hook it up to ratatui's canvas or [chart](https://github.com/ratatui-org/ratatui/tree/main/examples#chart) widgets.  This probably wouldn't be all that useful in practice, but might be fun anyway :)

sharkdp:
> For CLI usage, I wonder if we could hook it up to ratatui's canvas or [chart](https://github.com/ratatui-org/ratatui/tree/main/examples#chart) widgets. This probably wouldn't be all that useful in practice, but might be fun anyway :)

Cool. Hadn't heard about ratatui. But yeah, I think a file-based backend and maybe an actual GUI backend might be more useful options.

sharkdp:
This changeset now uses plotly, which allows us to open interactive plots in the browser:

image

sharkdp:
@irevoire I had a new idea for the API. It's a bit hacky, but could be enough for a first draft. I'm now using your enhanced reverse-apply operator to add optional arguments like `xlabel` and `ylabel`. Please see the updated PR description for details, and let me know what you think, if you're interested.

irevoire:
Oh yes we can now do a builder pattern quite easily! 
IMO, that's nicer than having a « Default » implementation for the structure 🤔 

On another subject, I’m really starting to think that we may need scopes earlier than I expected:
```
Plot.line_plot(amplitude, 0 s, 8 s)
  |> Plot.xlabel("Time")
  |> Plot.ylabel("Amplitude")
  |> Plot.show
```

Something like that seems easier to understand, and it will free up a lot of symbols for the users.

sharkdp:
> On another subject, I’m really starting to think that we may need scopes earlier than I expected:

Yes, absolutely. It's a problem already since I can't reuse things like `ylabel` for other plot types. I added a slightly hacky solution for now in order to be able to have just one `show` function for both line plots and bar plots. The latter of which are now also supported (see PR description).

PR #469, opened 2024-06-13
Auto-generated reference documentations (functions)

sharkdp:
Looks like this:

image


sharkdp:
This is now a lot nicer with the new (revived) function-signature printing

Issue #448, opened 2024-06-04
Lists are embarrassingly slow


sharkdp:
Lists were implemented without having performance in mind. There's certainly lots of things we can do to optimize them a bit.

We do not need highly performant lists. But `sum(map(sqrt, range(0, 10000)))` shouldn't take *seconds*.

Possible things to look into:
- Lists are currently represented as  ~~`Vec`s~~ `Deque`s in the VM. This can be challenged.
- We might want to implement some functions like `element_at` via the FFI. This would be especially true for functions that we use in lots of places (e.g. `map`?).
- The Numbat implementations of most list algorithms are certainly not optimal. To improve this, we would need #201

sharkdp:
First step in the right direction: #478

sharkdp:
I looked into optimizing `map` for a bit.

Below is the code that we currently generate for a naive recursive implementation of map
```py
fn map<A, B>(f: Fn[(A) -> B], xs: List<A>) -> List<B> =
  if is_empty(xs)
    then []
    else cons(f(head(xs)), map(f, tail(xs)))
```

![image](https://github.com/sharkdp/numbat/assets/4209276/782feb6b-b967-4a16-b61a-78635d7d95ab)

This doesn't look too bad, but the `GetLocal` calls are extremely expensive, since we currently clone the full list. I tried experimenting with a `Cow<'a, Deque<Value>>` as a representation for lists which would only require copies if we actually modified the lists (not in Numbat, but in Rust). For example, for a function like `f(xs) = is_empty(xs)`, we would not have to copy the list, just to pass it to `is_empty`. However, for most of the list functions, we call both `head(xs)` as well as `tail(xs)`. Both of these operations mutate the passed-in list. We would probably need support for pattern matching (`case xs of x:rest -> …`) in order to make this efficient, because then we would see the full split-into-head-and-tail operation all at once.

I also tried tried to implement `map` natively in Rust instead of Numbat (as a FFI function), but this is also not straightforward. We are being passed a function reference, and we can't do anything with it inside a FFI function. There currently no way of executing Numbat code from a FFI function.

Finally, I thought about compiling `map` calls into a loop. But it's not clear to me how this would be best achieved in a stack-based VM. We have three things to deal with: the input list `xs`, the function reference `f`, and a  list of output elements that we collected so far (`ys`). Let's say we start by pushing the arguments to the stack:
```
[f] [xs]
```
We could then create an initially empty list for `ys`, and push it as well:
```
[...]

PR #478, opened 2024-06-23
Faster lists

sharkdp:
See #448 

Benchmark
```
sum(map(sqrt, range(0, 10000)))
```
| state | time [s] |
| --- | --- |
| master | 12.4 |
| [remove unnecessary cloning of values in VM](f1e906395f85131af14ff59d98c6dbc472ca3bc1) | 6.7 |

The next biggest problem is that `GetLocal` ops clone the entire list. Maybe we can use `Cow<Value>`?

### DIFF
--- a/numbat/src/ffi/currency.rs
+++ b/numbat/src/ffi/currency.rs
@@ -1,13 +1,14 @@
 use super::macros::*;
+use super::Args;
 use super::Result;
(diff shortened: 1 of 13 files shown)


sharkdp:
I guess there is no harm in merging this as-is.

Issue #485, opened 2024-07-01
Add a 'run' link to all examples in the documenation


sharkdp:
Original idea by @eminence:

> it would be neat if all these examples had a "run" button that could immeditally run them in-browser 

In order to do this, we probably want to convert the [`build.sh`](https://github.com/sharkdp/numbat/blob/master/book/build.sh) into a Python script. Then, it should be fairly simple to turn the code of a given example into a URL by using standard URL encoding techniques to generate a `https://numbat.dev/?q=<code>` link.

PR #492, opened 2024-07-02
 Add a 'run' link to all examples in the documenation

Bzero:
Should resolve issue #485.


book/build.py
```diff
@@ -0,0 +1,54 @@
...
+    path_in = SCRIPT_DIR.parent / "examples" / f"{filename}.nbt"
+    path_out = SCRIPT_DIR / "src" / f"example-{filename}.md"
+
+    code = []
+    with open(path_in, "r") as fin:
+        for line in fin:
+            if not strip_assets or "assert_eq" not in line:
+                code.append(line)
+
+    url = f"https://numbat.dev/?q={urllib.parse.quote_plus(''.join(code))}"
```
sharkdp, line 19: Was about to comment that this doesn't work on Python 3.10 locally for me, but you already fixed that :+1:

book/build.py
```diff
@@ -0,0 +1,54 @@
...
+
+    with open(path_out, "w") as fout:
+        fout.write("<!-- This file is autogenerated! Do not modify it -->\n")
+        fout.write("\n")
+        fout.write(f'# {title} <a href="{url}"><i class="fa fa-play"></i></a>\n')
```
sharkdp, line 24: Nice idea with the icon! Layout-wise, I think I would prefer something a bit more conservative like:
```suggestion
        fout.write(f"# {title}\n")
        fout.write(f'<a href="{url}"><i class="fa fa-play"></i> Run this example</a>\n')
```

book/build.py
```diff
@@ -0,0 +1,54 @@
...
+generate_example("pipe_flow_rate", "Flow rate in a pipe", True)
+generate_example("population_growth", "Population growth", True)
+generate_example("recipe", "Recipe", True)
+generate_example("voyager", "Voyager", True)
+generate_example("xkcd_687", "XKCD 687", True)
+generate_example("xkcd_2585", "XKCD 2585", True)
+generate_example("xkcd_2812", "XKCD 2812", True)
+generate_example("workhours", "Workhours", True)
+
+generate_example("numbat_syntax", "Syntax overview", False)
```
sharkdp, line 48: Running this syntax overview (all at once) does not make a lot of sense. Can we skip the "Run" link for this page? We could add a fourth (defaulted) argument to `generate_example`

book/build.py
```diff
@@ -0,0 +1,54 @@
+import subprocess
+from pathlib import Path
+import urllib.parse
+
+
+SCRIPT_DIR = Path(__file__).parent
+
+
+def generate_example(filename, title, strip_assets):
```
sharkdp, line 9:
```suggestion
def generate_example(filename, title, strip_asserts):
```

sharkdp (changes requested):
Thank you very much — this is excellent!

book/build.py
Bzero: Yes, I forgot that this only works with newer versions.

book/build.py
Bzero: Agreed, this is a bit much of a wall of code. I added a parameter to turn the link insertion off.

book/build.py
Bzero: Sounds good to me!

book/build.py
```diff
@@ -0,0 +1,58 @@
...
+generate_example("musical_note_frequency", "Musical note frequency")
+generate_example("paper_size", "Paper sizes")
+generate_example("pipe_flow_rate", "Flow rate in a pipe")
+generate_example("population_growth", "Population growth")
+generate_example("recipe", "Recipe")
+generate_example("voyager", "Voyager")
+generate_example("xkcd_687", "XKCD 687")
+generate_example("xkcd_2585", "XKCD 2585")
+generate_example("xkcd_2812", "XKCD 2812")
+generate_example("workhours", "Workhours")
```
Bzero, line 50: The "workhours" example was missing in 'build.sh', I would think this was not intentional, right?

sharkdp: It was. Notice that there are a lot more examples in `examples/` that are not part of the documentation. I only hand-selected a few of them. I removed this line again.

sharkdp (approved):
Thank you for the updates. Your changes are live: https://numbat.dev/doc/example-voyager.html

book/build.py
Bzero: Yes, I noticed that, but it is the only example that is not present in `build.sh`/`build.py` but has an [associated file](https://github.com/sharkdp/numbat/blob/4709bc3d9b6a546a0d90b77578c54bd41faf5167/book/src/example-workhours.md) in `/book/src/`. Maybe that file should be removed too in this case?

sharkdp: Oh, right.  I guess it was part of the documentation at some point but then I removed it. I'll remove that file. Thank you.

PR #495, opened 2024-07-16
Add new heat energy units

### LINKED ISSUE #471: Add heat energy units therm and thermie
https://en.wikipedia.org/wiki/Therm
https://en.wikipedia.org/wiki/Thermie

MugheesQasim:
Closes #471


sharkdp:
Thank you. You can fix the doc build by following the suggestion in the CI job:

> Make sure to run 'build.py' in the book/ folder

MugheesQasim:
Thanks for the comment. PR is good to merge now.

numbat/modules/units/misc.nbt
```diff
@@ -30,6 +30,16 @@ unit calorie: Energy = 4.184 joule
 @aliases(Btu)
 unit BTU: Energy = 1055.05585262 joule
 
+@name("Therm")
+@url("https://en.wikipedia.org/wiki/Therm")
+@aliases(therms, thm : short)
+unit therm: Energy = 1.05506e8 joule
```
sharkdp, line 36: In #471, you wrote that there are multiple definitions. It would be good to document which one we chose. You can use the `@description(…)` decorator for that (it's not used for other units yet since this is a relatively recent feature).

Also, it would be great if we could define `therm` in terms of BTU. For the thest above, we would leave the value in Joule, but here, I think we could define
```suggestion
unit therm: Energy = 10_000 BTU
```
if that matches the definition.

numbat/modules/units/misc.nbt
```diff
@@ -30,6 +30,16 @@ unit calorie: Energy = 4.184 joule
...
+
+@name("Thermie")
+@url("https://en.wikipedia.org/wiki/Thermie")
+@aliases(thermies, th : short)
+unit thermie: Energy = 4.1868e6 joule
```
sharkdp, line 41: Could we use
```suggestion
unit thermie: Energy = 1000 kcal
```
here?

numbat/modules/units/misc.nbt
sharkdp: > I would suggest to create a new issue with title something like this "Use @description(...) decorator to add descriptions of units with multiple definitions

Let's please do that while we're at it anyway. Now we still know what kind of definition we picked. Opening a new ticket to do that for other units as well sounds like a good idea though.

sharkdp: > Using `BTU` instead of `joule` will remove the need to add description as the definition of `therm` depends on the definition of `BTU`.

Hm, not really. Users will not look up the definition of the unit in the source code. They will look that up in the documentation or via `info BTU`. `info BTU` includes the description field already, and we should probably add that to the [docs](https://numbat.dev/doc/list-units.html) as well.

MugheesQasim: > Users will not look up the definition of the unit in the source code. They will look that up in the documentation or via `info BTU`

I guess I was not able to convey my point. `1 therm` is exactly equal to `100_000 BTU`. `info therm` will show user this value `1 therm = 100_000 BTU`. Now the definition of `BTU` used in prelude will determine the value of `therm` in `joules`. `info BTU` should show user which exact definition of `BTU` is being used. And description is currently missing in the documentation for `BTU`

Let me know if I am misunderstanding something. Thank you : )

sharkdp: > `1 therm` is exactly equal to `100_000 BTU`. `info therm` will show user this value `1 therm = 100_000 BTU`.

Sure. But even if it's defined in terms of BTU, it still makes sense to write down which version of `therm` this is in a simple textual description for the user. If we don't add one, the user might not even know that there are multiple definitions... and that they might have chosen the wrong one for their use case.

Anyway, I'll merge this as-is. Maybe someone else feels like doing that in the future.

sharkdp (approved):

examples/tests/unit_consistency_misc.nbt
```diff
@@ -62,6 +62,8 @@ assert_eq(8e-15, 8 ppq, 1e-16)
 
 assert_eq(1 calorie, 4.184 J)
 assert_eq(1 BTU, 1055.05585262 J)
+assert_eq(1 therm, 1.05505585262e8 J)
+assert_eq(1 thermie, 4.184e6 J)
```
sharkdp, line 66: That is actually different from the value quoted on Wikipedia:

> The thermie is equivalent to 1000 [kcal](https://en.wikipedia.org/wiki/Calorie), 4.1868 [MJ](https://en.wikipedia.org/wiki/Megajoule) or 3968.3 [BTU](https://en.wikipedia.org/wiki/British_thermal_unit).

 I think Wikipedia might be wrong here since their values are contradictory. According to their own linked article, 1 kcal = 4.184 J, so if 1 thermie = 1000 kcal, then 1 thermie = 4.184 MJ.

The 4.1868 value refers to some older definition of the calorie (International Steam Table calorie, 1929 and 1956).

And the only source that they quote on the Thermie article is https://www.cnrtl.fr/definition/thermie, which also has a value of 4.184 MJ.

I'm going to change the Wikipedia article, I guess.

sharkdp: I opened a discussion on the Talk:Thermie page: https://en.wikipedia.org/wiki/Talk:Thermie#Should_the_value_of_the_thermie_be_4.184_MJ?

numbat/modules/units/misc.nbt
MugheesQasim: > If we don't add one, the user might not even know that there are multiple definitions... and that they might have chosen the wrong one for their use case.

this is a valid point, adding a description for the unit `therm`

examples/tests/unit_consistency_misc.nbt
MugheesQasim: This publication might also give some insight: [Publication:Calorie](https://www.degruyter.com/document/doi/10.1351/pac196408020095/html) (Specifically, pages 106 and 107)

sharkdp (approved):
Thanks!

PR #504, opened 2024-07-24
Color format conversions

sharkdp:
This was initially more for fun, but this might be actually useful. It's implemented in Numbat, without any FFI/language support. This would allow us to do things like

image

And if we want, we could add other color representations such as HSL, as well as other color spaces such as LCh, Lab*. We could also add color manipulation etc. I have written [a whole tool](https://github.com/sharkdp/pastel) for color operations like those, but it might be useful to have this inside an easily accessible calculator.

### DIFF
--- a/book/build.py
+++ b/book/build.py
@@ -255,6 +255,10 @@ list_of_functions(
                 "title": "Temperature conversion",
                 "modules": ["physics::temperature_conversion"],
             },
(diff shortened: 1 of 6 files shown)


sharkdp:
Let's merge this as a module in `extra`, which is not enabled by default.

PR #513, opened 2024-07-28
Allow closing parens and newlines

### LINKED ISSUE #507: Allow closing parens (and trailing comma?) in function calls to be on the next line
This should be allowed (and probably easy to fix):
```
atan2(
  2 m,
  5 m
)
```

While we're at it, we might also want to allow trailing commas, like many other programming languages do?
```
atan2(
  2 m,
  5 m,
)
```

irevoire:
Fixes #507

### DIFF
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -79,7 +79,7 @@ use thiserror::Error;
 
 #[derive(Error, Debug, Clone, PartialEq, Eq)]
 pub enum ParseErrorKind {
-    #[error("{0}")]
+    #[error(transparent)]
     TokenizerError(TokenizerErrorKind),
 
     #[error(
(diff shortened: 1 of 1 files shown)


numbat/src/parser.rs
```diff
@@ -1686,6 +1705,10 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn ignore_newlines(&mut self) {
+        while self.match_exact(TokenKind::Newline).is_some() {}
+    }
```
sharkdp, line 1710: Check out `skip_empty_lines` :smile:

sharkdp (changes requested):
Thank you very much! Just one minor comment

numbat/src/parser.rs
irevoire: Nice thank you

sharkdp (approved):

Issue #73, opened 2023-02-15
Plot functionality


sharkdp:
Combined with #51, this could be really powerful.

- Evaluate functions at multiple points
- Automatically label axes (x: input variable, input unit; y: function name, output unit)

irevoire:
> I recently played a lot with https://github.com/sharkdp/numbat/issues/73 and that is going to become so much faster.

About that I also played with it a little bit and implemented a version using [`egui`](https://www.egui.rs/)
<img width="806" alt="image" src="https://github.com/user-attachments/assets/b9249e7a-d785-424e-bd11-e72dce21e5b9">

It comes with a native mode and a wasm mode which can run in a browser.

Pros:
- It gives us great support for moving and zooming on the plot by supporting drag'n'drop + scrolling to zoom
- Great support for tactile screen as well for phone and tablet

<details open>
<summary>We can create our own mark on charts and display stuff with the unit we want</summary>

Basically, egui let us write the mark we wants and spaces them however we want.
For example I played a bit with this functionality here: http://meteo.tamo.cool/
When I'm far away I only display the years:
<img width="444" alt="image" src="https://github.com/user-attachments/assets/74f56954-18d6-434f-9c65-47517e17096a">

Once I can decide to additionally display the month+year
<img width="793" alt="image" src="https://github.com/user-attachments/assets/56d67476-2e84-4a78-a1a3-b6a463d9bd75">

And by zooming even more I start to display the hours:
<img width="1327" alt="image" src="https://github.com/user-attachments/assets/51bd6d91-7121-43b9-9cfb-89effe137f75">

With numbat in mind and https://github.com/sharkdp/numbat/issues/510 I can imagine a plot when the unit change while you zoom and that would be awesome.

</details>

Cons:
- It doesn't generate HTML like plotly, which means it may not play well with the jupyter notebook you're trying to do? (I don't know anything about jupyter notebook)
- The only way to save a graph on disk is to use the internal screenshot functionality, which _I think_ requires us to open a window

Here's my branch where I only did the bare minimum: https://github.com/sharkdp/numbat/pull/516
If you're interested by this approach let me know and maybe I could help

sharkdp:
> About that I also played with it a little bit and implemented a version using [`egui`](https://www.egui.rs/)

Nice! I just played with it as well and looks great.



> It gives us great support for moving and zooming on the plot by supporting drag'n'drop + scrolling to zoom

plotly can do this as well

[...]

PR #515, opened 2024-07-31
Make list slightly less embarrassingly slow

irevoire:
Make list faster in some cases.

Part of #448

On my computer, running:
```
time cargo run --release -- -e 'sum(map(sqrt, range(0, 10_000)))'
```

Take 42s on main vs 0.4s on my branch.

I don't think this approach is general enough, though. Having pattern matching would make performant numbat code easier to write.

-----

## How does it works?

I replaced the `NumbatList` type with a custom `ArcListView` type that shares its allocation with all other list referring to the same allocation.
It hold a reference counted ptr to the allocation + the bounds accessible by the allocation:

```rust
/// Reference counted list / list view
#[derive(Clone, Eq)]
pub struct ArcListView<T> {
    /// The original alloc shared between all values
    alloc: Arc<VecDeque<T>>,
    /// The indexes accessible to us. If `None` we own the whole allocation
    view: Option<(usize, usize)>,
}
```

Then, doing a `tail` is a matter of advancing the starting view by one.
`head` simply returns the first element.
And `push_front` check if he's the only owner of the allocation, if it's the case it never allocate (this is always the case while building the list for example so its important).
if it's not the only owner of the allocation, it clones the allocation entirely. We could probably do something smarter here with a chained list or something, but since it yields pretty good results, I didn't try harder.

We now need to take extra care while implementing PartialEq

-----

EDIT: While playing with list a little bit I noticed that the following code was still taking around 15s:
```rust
let a = now()
let list = range(0, 10_000)
print("Made the list in {human(now() - a)}")

let b = now()
let reversed = reverse(list)
print("Reversed the list in {human(now() - b)}")
```

Which is expected considering the number of operations we're doing to push something at the end of a list even though we internally uses a vecdeque.
By providing an ffi `cons_end` function I went from 15s to 0.002s

### DIFF
--- a/numbat/src/ffi/lists.rs
+++ b/numbat/src/ffi/lists.rs
@@ -36,3 +36,11 @@ pub fn cons(mut args: Args) -> Result<Value> {
 
     return_list!(list)
 }
+
+pub fn cons_end(mut args: Args) -> Result<Value> {
+    let mut list = list_arg!(args);
+    let element = arg!(args);
+    list.push_back(element);
+
+    return_list!(list)
+}
--- a/numbat/src/ffi/lists.rs
+++ b/numbat/src/ffi/lists.rs
@@ -11,10 +11,12 @@ pub fn len(mut args: Args) -> Result<Value> {
 }
 
 pub fn head(mut args: Args) -> Result<Value> {
-    let mut list = list_arg!(args);
+    let list = list_arg!(args);
 
-    if let Some(first) = list.pop_front() {
-        Ok(first)
+    // We don't need to pop or drop anything, the whole allocation will
+    // be dropped if we're its last owner.
+    if let Some(first) = list.front() {
+        Ok(first.clone())
     } else {
         Err(RuntimeError::EmptyList)
     }
@@ -23,11 +25,8 @@ pub fn head(mut args: Args) -> Result<Value> {
 pub fn tail(mut args: Args) -> Result<Value> {
     let mut list = list_arg!(args);
 
-    if list.pop_front().is_some() {
-        return_list!(list)
-    } else {
-        Err(RuntimeError::EmptyList)
-    }
+    list.advance_view()?;
+    Ok(list.into())
 }
 
 pub fn cons(mut args: Args) -> Result<Value> {
--- a/numbat/src/ffi/lists.rs
+++ b/numbat/src/ffi/lists.rs
@@ -13,10 +13,8 @@ pub fn len(mut args: Args) -> Result<Value> {
 pub fn head(mut args: Args) -> Result<Value> {
     let list = list_arg!(args);
 
-    // We don't need to pop or drop anything, the whole allocation will
-    // be dropped if we're its last owner.
-    if let Some(first) = list.front() {
-        Ok(first.clone())
+    if let Some(first) = list.head() {
+        Ok(first)
     } else {
         Err(RuntimeError::EmptyList)
     }
--- a/numbat/src/ffi/lists.rs
+++ b/numbat/src/ffi/lists.rs
@@ -23,7 +23,7 @@ pub fn head(mut args: Args) -> Result<Value> {
 pub fn tail(mut args: Args) -> Result<Value> {
     let mut list = list_arg!(args);
 
-    list.advance_view()?;
+    list.tail()?;
     Ok(list.into())
 }
 
--- a/numbat/src/list.rs
+++ b/numbat/src/list.rs
@@ -122,7 +122,7 @@ impl<T: Clone> NumbatList<T> {
         }
     }
 
-    /// Allocate iif the list being used by another value at the same time
+    /// Allocate if the list is being used by another value at the same time
     pub fn push_front(&mut self, element: T) {
         let (view, inner) = self.make_mut();
         if let Some((start, end)) = view {
@@ -141,7 +141,7 @@ impl<T: Clone> NumbatList<T> {
         }
     }
 
-    /// Allocate iif the list being used by another value at the same time
+    /// Allocate if the list is being used by another value at the same time
     pub fn push_back(&mut self, element: T) {
         let (view, inner) = self.make_mut();
         if let Some((_start, end)) = view {
--- a/numbat/src/list.rs
+++ b/numbat/src/list.rs
@@ -91,13 +91,8 @@ impl<T> NumbatList<T> {
     }
 
     pub fn iter(&self) -> impl Iterator<Item = &T> {
-        self.alloc
-            .iter()
-            .skip(self.view.map_or(0, |(start, _end)| start))
-            .take(
-                self.view
-                    .map_or(self.alloc.len(), |(start, end)| end - start),
-            )
+        let (start, end) = self.view.map_or((0, self.alloc.len()), |view| view);
+        self.alloc.iter().skip(start).take(end - start)
     }
 }
 
--- a/numbat/src/list.rs
+++ b/numbat/src/list.rs
@@ -1,0 +1,155 @@
+use std::{collections::VecDeque, fmt, sync::Arc};
+
+use crate::{value::Value, RuntimeError};
+
+/// Reference counted list / list view
+#[derive(Clone, Eq)]
+pub struct NumbatList<T> {
+    /// The original alloc shared between all values
+    alloc: Arc<VecDeque<T>>,
+    /// The indexes accessible to us. If `None` we own the whole allocation
+    view: Option<(usize, usize)>,
+}
+
+impl<T: fmt::Debug + Clone> fmt::Debug for NumbatList<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_list().entries(self.iter()).finish()
+    }
+}
+
+impl<T: PartialEq> PartialEq for NumbatList<T> {
+    fn eq(&self, other: &Self) -> bool {
+        // Best case scenario, the slice don't have the same len and we can early exit
+        if self.len() != other.len() {
+            return false;
+        }
+        // Second best case scenario, the slice comes from the same allocation
+        // and have the same view => they are equal
+        if Arc::as_ptr(&self.alloc) == Arc::as_ptr(&other.alloc) && self.view == other.view {
+            true
+        } else {
+            // Worst case scenario, we need to compare all the elements one by one
+            self.iter().zip(other.iter()).all(|(l, r)| l == r)
+        }
+    }
+}
+
+impl<T> NumbatList<T> {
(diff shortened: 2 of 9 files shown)


sharkdp:
Wow — thank you! Excellent idea and fantastic results! I'm really looking forward to this. I recently played a lot with #73 and that is going to become so much faster.

Can we move `ArcListView` to a new module (maybe call it just `ListView` or `NumbatList` or similar?) and add some unit tests?

I plan to do a more detailed review soon.

irevoire:
Done ✅ 
The PR is ready to review again

numbat/src/ffi/lists.rs
```diff
@@ -11,10 +11,12 @@ pub fn len(mut args: Args) -> Result<Value> {
...
 pub fn head(mut args: Args) -> Result<Value> {
-    let mut list = list_arg!(args);
+    let list = list_arg!(args);
 
-    if let Some(first) = list.pop_front() {
-        Ok(first)
+    // We don't need to pop or drop anything, the whole allocation will
+    // be dropped if we're its last owner.
+    if let Some(first) = list.front() {
+        Ok(first.clone())
```
sharkdp, line 19: If the `first` element is a large object itself (a struct or a list), we now need an additional clone here?

numbat/src/ffi/lists.rs
```diff
@@ -23,11 +25,8 @@ pub fn head(mut args: Args) -> Result<Value> {
 pub fn tail(mut args: Args) -> Result<Value> {
     let mut list = list_arg!(args);
 
-    if list.pop_front().is_some() {
-        return_list!(list)
-    } else {
-        Err(RuntimeError::EmptyList)
-    }
+    list.advance_view()?;
```
sharkdp, line 28: Minor "advance view" is sort of a implementation detail. Maybe we can call that function `tail` or `into_tail` or similar?

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
+//! This module defines a custom kind of list used in the [`numbat::Value::List`].
+//! The specificities of this list are that:
+//! - It's based on a `VecDeque`, which means insertion at the beginning or the end is `O(1)`.
+//! - It's stored behind an `Arc`, which makes cloning the value very cheap.
```
sharkdp, line 4: Maybe 
```suggestion
//! - It's stored behind an `Arc`, which makes cloning the [`numbat::Value`] very cheap.
```

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+//! - It tries to reduce the number of allocations as much as possible, even when shared between multiple values.
+//!   If you are the only owner of the list or while you're not pushing elements in it in, never re-allocate.
```
sharkdp, line 6: Maybe
```suggestion
//!   It never re-allocates unless there are multiple owners and elements are being added to the list.
```

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+use std::{collections::VecDeque, fmt, sync::Arc};
+
+use crate::{value::Value, RuntimeError};
+
+/// Reference counted list / list view
+#[derive(Clone, Eq)]
+pub struct NumbatList<T> {
+    /// The original alloc shared between all values
+    alloc: Arc<VecDeque<T>>,
+    /// The indexes accessible to us. If `None` we own the whole allocation
```
sharkdp, line 17:
```suggestion
    /// The indexes accessible to us. If this is `None`, we own the whole allocation
```

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+impl<T> Default for NumbatList<T> {
+    fn default() -> Self {
+        Self {
+            alloc: Default::default(),
+            view: Default::default(),
+        }
+    }
+}
```
sharkdp, line 28: Can this be `derive`d?

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+
+impl<T: fmt::Debug + Clone> fmt::Debug for NumbatList<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_list().entries(self.iter()).finish()
```
sharkdp, line 32: Cool — I learned something new :smile:

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+impl<T: PartialEq> PartialEq for NumbatList<T> {
+    fn eq(&self, other: &Self) -> bool {
+        // Best case scenario, the slice don't have the same len and we can early exit
```
sharkdp, line 38:
```suggestion
        // Best case scenario, the slice doesn't have the same length; exit early
```

numbat/src/list.rs
irevoire: No because I don't want to impose the `Default` trait to `T` and that makes rust dumb.
It won't implements `Default` on a structure if all of its generic parameters don't implements `Debug` as well 😔

numbat/src/ffi/lists.rs
irevoire: Huum, most of the time it's the same as before since we were already cloning the whole list.
But now it's true that we could check if we're the only owner of the list and only clone it when someone else is using the list.

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+        if self.len() != other.len() {
+            return false;
+        }
+        // Second best case scenario, the slice comes from the same allocation
+        // and have the same view => they are equal
+        if Arc::as_ptr(&self.alloc) == Arc::as_ptr(&other.alloc) && self.view == other.view {
```
sharkdp, line 44: Maybe use [`Arc::ptr_eq`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.ptr_eq) for these sort of checks?

```suggestion
        if Arc::ptr_eq(&self.alloc, &other.alloc) && self.view == other.view {
```

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+impl<T: PartialEq> PartialEq for NumbatList<T> {
+    fn eq(&self, other: &Self) -> bool {
+        // Best case scenario, the slice don't have the same len and we can early exit
+        if self.len() != other.len() {
+            return false;
+        }
+        // Second best case scenario, the slice comes from the same allocation
+        // and have the same view => they are equal
```
sharkdp, line 43:
```suggestion
        // Second best case, the other slice comes from the same allocation and
        // has the same view => they are equal
```

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = &T> {
+        self.alloc
+            .iter()
+            .skip(self.view.map_or(0, |(start, _end)| start))
+            .take(
+                self.view
+                    .map_or(self.alloc.len(), |(start, end)| end - start),
+            )
```
sharkdp, line 107: I wonder if this would be a bit easier to read like this
```suggestion
        if let Some((start, end)) = self.view {
          self.alloc.iter().skip(start).take(end - start)
        } else {
          self.alloc.iter()
        }
```

or maybe even
```suggestion
        if let Some((start, end)) = self.view {
          self.alloc[start..end].iter()
        } else {
          self.alloc.iter()
        }
```
if that works?

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+        assert_eq!(Arc::as_ptr(&list1.alloc), Arc::as_ptr(&list2.alloc));
+
+        // Pushing something new in the first list should re-allocate
+        list1.push_front(2);
+        assert_ne!(Arc::as_ptr(&list1.alloc), Arc::as_ptr(&list2.alloc));
+
+        // Now that list2 is alone on its allocation it should be able
+        // to push an element to the front without re-allocating anything
+        let alloc = Arc::as_ptr(&list2.alloc);
+        // Pushing something new in the first list should re-allocate
```
sharkdp, line 265:
```suggestion
        // Pushing something new in the first list should not re-allocate
```

numbat/src/list.rs
```diff
@@ -0,0 +1,297 @@
...
+        list1.push_back(0);
+
+        let mut list2 = list1.clone();
+
+        assert_eq!(list1, list2);
+
+        list2.advance_view().unwrap();
+        assert_ne!(list1, list2);
+
+        // Even if the list do not share the same allocation they should match
```
sharkdp, line 284:
```suggestion
        // Even if the lists don't share the same allocation, they should be equal
```

numbat/src/list.rs
irevoire: The first proposition doesn't work because the two iterators don't share the same type.
To make it work, I would need to wrap them in a `Box<dyn Iterator<...>>`, which is not ideal.
I’ll try the second one, but I think that's what I initially tried, and it wasn't working as well.
Will keep you updated

numbat/src/ffi/lists.rs
irevoire: Fixed in: https://github.com/sharkdp/numbat/pull/515/commits/f15ea87ef8c890ccbbd20ed22416638f16c2982e

numbat/src/list.rs
irevoire: Oh yes I remember now, the second solution doesn't work either because `VecDeque` doesn't implements `Index<SliceIndex>` like the `Vec` do, but only [`Index<usize>`](https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html#impl-Index%3Cusize%3E-for-VecDeque%3CT,+A%3E).

Which means you can only retrieve one element at a time in a `VecDeque` 😔 

I still tried to simplify it a little bit: https://github.com/sharkdp/numbat/pull/515/commits/56758d819870bb08c99929a3bbf3f178a7a57721

numbat/src/list.rs
```diff
@@ -0,0 +1,299 @@
...
+        if Arc::strong_count(&self.alloc) == 1 {
+            // safety: unwrap cannot fail because we ensured there was only one strong reference above
+            let mut alloc = Arc::try_unwrap(self.alloc).map_err(|_| ()).unwrap();
+            alloc.swap_remove_front(front)
+        } else {
+            self.alloc.get(front).cloned()
+        }
+    }
+
+    /// Allocate iif the list being used by another value at the same time
```
sharkdp, line 125: Minor: Either just "if" or "iff" (if and only if). Similar below.
```suggestion
    /// Allocate if the list being used by another value at the same time
```

sharkdp (approved):
Thank you for the updates and explanations.

Fantastic work!

PR #518, opened 2024-08-03
Make the reverse function call more useful

irevoire:
Hey, while working on https://github.com/sharkdp/numbat/pull/515, I didn't like the API around lists, especially the way we don't chain calls, which makes the code very hard to read.

With this PR, the following chunks of code are equivalent:
```ocaml
sum(map(sqrt, range(0, 10_000)))
range(0, 10_000) // map(sqrt) // sum
```

It's clearly inspired by the `|>` operator of OCaml; I believe Haskell has something similar with the `$` or `&` operator as well.
Obviously, it's not as clean as in these two languages since numbat uses parens to call functions.
But it still works, and it should become even more useful once we have closures.

> [!note]
> For the future, we should take care while designing the APIs (and especially the list/set/map APIs) that the data structure is always the last parameter

----

Also, I think I already asked you a long time ago and forgot the answer, but what do you think of renaming (or duplicating) this operator with another symbol?
I’ve never seen `//` anywhere else, and it doesn't ring a bell at all when I use it. I would be way more comfortable with the arroy of OCaml `|>`, which shows its intent more clearly.

### DIFF
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -108,6 +108,11 @@ pub enum ParseErrorKind {
     #[error("Expected identifier")]
     ExpectedIdentifier,
 
+    #[error(
+        "Expected identifier or function call after postfix apply (`//`) but instead got: `{0}`"
+    )]
+    ExpectedIdentifierOrCallAfterPostfixApply(String),
+
     #[error("Expected dimension identifier, '1', or opening parenthesis")]
     ExpectedDimensionPrimary,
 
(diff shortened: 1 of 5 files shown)


sharkdp:
> With this PR, the following chunks of code are equivalent:
> 
> sum(map(sqrt, range(0, 10_000)))
> range(0, 10_000) // map(sqrt) // sum

Very nice. I hadn't thought of that!

> I’ve never seen `//` anywhere else, and it doesn't ring a bell at all when I use it. I would be way more comfortable with the arroy of OCaml `|>`, which shows its intent more clearly.

`//` is inspired by Mathematica which uses it for the same (https://reference.wolfram.com/language/guide/FunctionCompositionAndOperatorForms.html)

> what do you think of renaming (or duplicating) this operator with another symbol?

I'd be okay with that. I don't think `//` is a particularly fine choice. My only minor concern is that it's easier to type than `|>` (at least on a German keyboard). I like how `|>` looks. And if it has precedent in OCaml, even better.

Let's go with a breaking change (instead of providing `|>` in addition to `//`).



> For the future, we should take care while designing the APIs (and especially the list/set/map APIs) that the data structure is always the last parameter

Is there anything that you would suggest changing now? The list feature is not even released officially, so I'm happy to change function signatures/names/etc.

numbat/src/parser.rs
```diff
@@ -1857,7 +1874,7 @@ mod tests {
         for input in inputs {
             match parse(input, 0) {
                 Err((_, errors)) => {
-                    assert_eq!(errors[0].kind, error_kind);
+                    assert_eq!(errors[0].kind, error_kind, "Failed on {}", input);
```
sharkdp, line 1877: Nice!

sharkdp (approved):
Very nice. Feel free to rewrite some (list-related) Numbat code using this new feature. But that is not a must, of course. I'm fine merging this as is

irevoire:
> Feel free to rewrite some (list-related) Numbat code using this new feature

I looked at all the functions in `list.nbt` and `statistics.nbt`, only `cons_end` had the list and element inverted.

I think we can merge now, I’ll do another PR to rename the operator so it's easier to revert if we need to.

> My only minor concern is that it's easier to type than |> (at least on a German keyboard).

Huum yes it's a bit harder on french keyboard as well I guess 🤔 
Haskell uses `&` from what I understand which I find strange since `&` is used so much for binary computation in other languages.
So that could be an option as well, I think it still convey the idea that you're going to chain stuff up.

But IMO `|>` it better at avoiding confusions.

> I like how |> looks. And if it has precedent in OCaml, even better.

And it looks even better with ligatures on 😁 
image

sharkdp (approved):
Thank you

irevoire:
Hey, coming back to this one thousand years later.
I discovered a language that uses **exactly** our syntax: The gleam https://gleam.run/
Just in case they change their front page, here's another random example from the internet: https://github.com/trilowy/exercism/blob/b12cb0becab36e88d0fb5c6b14e520b3505dd220/gleam/sum-of-multiples/src/sum_of_multiples.gleam#L5-L7

So I guess we weren't wrong about that in the end 😁

sharkdp:
Cool. Thank you for the feedback.

Issue #509, opened 2024-07-28
Support for multi-line functions


irevoire:
Hey, I’ve been using numbat more and more and being able to use multi-line functions would _really_ help.
I looked for issue on the subject but didn’t find any, would this be allowed if I made a PR?

And can we use `{}` like we do with structures?

sharkdp:
> being able to use multi-line functions would _really_ help

Yes. This is why I pinned ticket #201. I personally would really like to have what I proposed in https://github.com/sharkdp/numbat/issues/201#issuecomment-2174088579. Is there anything else you would want from "multi line" functions? For now, Numbat functions are pure, so the body is always an expression.. not a sequence of statements.

> And can we use `{}` like we do with structures?

This would make sense if we allow `let`-style bindings. Something like
```
fn f(x, y) {
  let z = x^2 + y^2
  sqrt(z)
}
```

For now, I personally prefer what I proposed in the reference ticket above:
```
fn f(x, y) = sqrt(z)
  where z = x^2 + y^2
```

But I also have to admit that I typed `{` a few times when I wanted to create a Numbat function. I'm open to discussing this.

irevoire:
> For now, Numbat functions are pure, so the body is always an expression.. not a sequence of statements.

Yep, that’s perfect. I don’t think we need anything else.

About this:
> Concerning the syntax, I would personally favor the let … in … syntax for [let-expressions](https://en.wikipedia.org/wiki/Let_expression) that seems common in other programming languages (Haskell) and the literature.
> But what I would like even more is the … where … syntax that is also used in functional programming languages, because it puts the important things first and reads a bit more naturally. It also extends more nicely to a multi-definition version:

Personally, I think I would prefer the `let ... in` statements, especially when there are conditionals going on:
```haskell
fn truc(n) = if m < 0
        then n
        else m
    where m = n / 2
```

vs

```ocaml
fn truc(n) =
    let m = n / 2 in
    if m < 0
        then n
        else m
```

imo it's easier to follow what's going on.

BUT, whatever you prefer I may have a bit of time to work on it next week

sharkdp:
> Is there a reason, though, for not having a sequence of statements?

Mostly simplicity of the grammar. And also the pureness property.  `<expr> where [list of bindings]` or `let [list of bindings] in <expr>` would still be an expression. Something like `{ print("hello"); <return-expr> }` is not an expression. And it's not a pure function due to the side effects.

[...]

Issue #201, opened 2023-10-10
Add support for local variables


eminence:
There are some "helper variables" in `physics/temperature_conversion.nbt` that currently leak through the prelude, like `offset_celsius` and `scale_fahrenheit`.   These variables feel like "implementation details" that should be contained only to the nbt file they are defined in.

Perhaps there could be a concept of local variables?

sharkdp:
Thank you for reporting this.

Yes, this is unfortunate. I see several ways to address this:
1. Add a way to have module-local constants/functions, like you proposed. We would probably need a new keyword like `private` for this (`let private offset_celsius = …`) or use a decorator like we have for unit declarations (`@private`)?
1. Do the opposite: have everything private by default and introduce a way to export entities from a module (`pub`/`public`/`@public`/`export`)
1. Add something similar to module scopes or namespaces, where we could have `mod details { let offset_celsius = …; }` and then use it internally with `details::offset_celsius`. That namespace would not be hidden, but it would not be imported by default when doing `use physics::temperature_conversion`. Only if someone would do `use physics::temperature_conversion::details`.
1. Add function-level local variables, maybe with a Haskell-style syntax:
   ```rs
   fn from_celsius(t_celsius: Scalar) -> Temperature =
     (t_celsius + offset_celsius) kelvin
     where offset_celsius = 273.15
   ```
   This would not really help in this case (as we want that constant to be present in multiple functions), but would certainly be helpful in a lot of scenarios

It's currently not clear to me yet which way we should take forward.

eminence:
I think if I had a magic wand that would give me one of these solutions for free, I probably would choose option 1 or 3.   I would probably not vote for option 2 because I think the majority of nbt code will want to export most everything (like how the current prelude is today).   

(Another short-term option might be to hide all variables that start with an underscore `_` from tab-completion and the output of "ls".  I think this would be pretty cheap to implement, but not very satisfying and not my preferred approach)

mhmdanas:
> (Another short-term option might be to hide all variables that start with an underscore _ from tab-completion and the output of "ls". I think this would be pretty cheap to implement, but not very satisfying and not my preferred approach)

I work a lot with Dart, and this is the approach it takes; I have to say that while it's pretty nice to be able to see whether an identifier is private at its use site, I agree that I wouldn't prefer it, and I think it just doesn't fit with the overall feel of the Numbat language.

yerke:
[...]

Issue #433, opened 2024-05-18
Type system extensions


sharkdp:
This issues is a collection of notes regarding future extensions of Numbats type system.

## Overview of what we have at the moment

[This page in the documentation](https://numbat.dev/doc/type-system.html) gives a rough overview.

### Types

- **Dimension types** for dealing with physical dimensions: [^kennedy]
  - Fundamental/base dimension types (`1`, `Length`, `Time`, `Mass`, …)
  - Combinations of dimension types (`Length / Time`, `Current × Time`, …)
  - Exponentiation (`Length³`)
  - Type aliases for dimension types (`dimension Frequency = Time^(-1)`)
- **Base types** (non-dimension types):
  - `Bool`
  - [`String`](https://numbat.dev/doc/list-functions.html#strings)
  - [`DateTime`](https://numbat.dev/doc/date-and-time.html)
  - `!` ("Never"), a bottom type that we currently use to type functions like `error` (`String -> !`).
- [Structs](https://numbat.dev/doc/structs.html) — nominally typed records like `struct Pair { s: Length, t: Time }`. In the PR that introduced structs, there is a discussion on whether we should have structurally typed or nominally typed records in Numbat.
- Function types
  - Example: `Fn[(A, B, C) -> X]` is the type of a function with parameter types `A, B, C` and return type `X`.

### Type system features

- Function types (but no closures so far, see #347)
  ```rs
  fn twice(f: Fn[(Scalar) -> Scalar], x: Scalar) = f(f(x))
  ```
- Type inference (limited, see #29)
  ```rs
  fn f<A, B>(a: A, c: A × B²) -> B = sqrt(c/a)
  f(2 m, 3 m s²)
  ```
- Parametric polymorphism (currently limited, see #166):
  ```rs
  fn max<T>(a: T, b: T) -> T = if a > b then a else b
  ```
- Recursion (somewhat limited, see [this page](https://numbat.dev/doc/function-definitions.html#recursive-functions))
  ```rs
  fn fib(n: Scalar) -> Scalar =
    if n ≤ 2
      then 1
      else fib(n - 2) + fib(n - 1)
  ```
- Some form of compile-time evaluation ("`consteval`") to support exponentiation of dimension types, as explained in [this article](https://numbat.dev/articles/intro.html)

### Types

[...]

PR #519, opened 2024-08-04
Multiline functions: local variables

irevoire:
Fixes https://github.com/sharkdp/numbat/issues/509
Fixes https://github.com/sharkdp/numbat/issues/201

The error reporting of the parser got a little bit worse. We can't detect trailing character at the end of a statement now and the error reported will all be `ExpectedPrimary`.
In order to fix that we would either need some special error handling when the last statement parsed was a function.
Or a different syntax.

Right now that's what I parse:
```ocaml
fn power_4(x: Scalar) = z
  where y = x * x
        z = y * y
```

Maybe we prefer something looking like that;

```ocaml
fn power_4(x: Scalar) = z
  where y = x * x
  where z = y * y
```

The issue with the current implementation is that after parsing a where I must continue to parse as many variable declaration as possible + all whitespace and newlines.

### DIFF
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -44,6 +44,12 @@ fn element_at<A>(i: Scalar, xs: List<A>) -> A =
     then head(xs)
     else element_at(i - 1, tail(xs))
 
+@description("Insert the element at index `i` in a list")
+fn insert_at<A>(el: A, i: Scalar, xs: List<A>) -> List<A> =
+  if i == 0
+    then cons(el, xs)
+    else cons(head(xs), insert_at(el, i - 1, tail(xs)))
+
 @description("Generate a range of integer numbers from `start` to `end` (inclusive)")
 fn range(start: Scalar, end: Scalar) -> List<Scalar> =
   if start > end
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -138,5 +138,7 @@ fn split(input: String, separator: String) -> List<String> =
     then []
     else if !str_contains(input, separator)
       then [input]
-      else cons(str_slice(input, 0, str_find(input, separator)),  # TODO: avoid duplication
-                split(str_slice(input, str_find(input, separator) + str_length(separator), str_length(input)), separator))
+      else cons(str_slice(input, 0, idx_separator),
+                split(str_slice(input, idx_separator + str_length(separator), str_length(input)), separator))
+  where
+    idx_separator = str_find(input, separator)
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -44,12 +44,6 @@ fn element_at<A>(i: Scalar, xs: List<A>) -> A =
     then head(xs)
     else element_at(i - 1, tail(xs))
 
-@description("Insert the element at index `i` in a list")
-fn insert_at<A>(el: A, i: Scalar, xs: List<A>) -> List<A> =
-  if i == 0
-    then cons(el, xs)
-    else cons(head(xs), insert_at(el, i - 1, tail(xs)))
-
 @description("Generate a range of integer numbers from `start` to `end` (inclusive)")
 fn range(start: Scalar, end: Scalar) -> List<Scalar> =
   if start > end
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -297,27 +297,12 @@ impl<'a> Parser<'a> {
             match self.peek().kind {
                 TokenKind::Newline => {
                     // Skip over empty lines
-                    while self.match_exact(TokenKind::Newline).is_some() {}
+                    self.skip_empty_lines();
                 }
                 TokenKind::Eof => {
                     break;
                 }
-                TokenKind::Equal => {
-                    errors.push(ParseError {
-                        kind: ParseErrorKind::TrailingEqualSign(
-                            self.last().unwrap().lexeme.clone(),
-                        ),
-                        span: self.peek().span,
-                    });
-                    self.recover_from_error();
-                }
-                _ => {
-                    errors.push(ParseError {
-                        kind: ParseErrorKind::TrailingCharacters(self.peek().lexeme.clone()),
-                        span: self.peek().span,
-                    });
-                    self.recover_from_error();
-                }
+                _ => {}
             }
         }
 
@@ -392,7 +377,7 @@ impl<'a> Parser<'a> {
         }
 
         if self.match_exact(TokenKind::Let).is_some() {
-            self.parse_let()
+            self.parse_variable().map(Statement::DefineVariable)
         } else if self.match_exact(TokenKind::Fn).is_some() {
             self.parse_function_declaration()
         } else if self.match_exact(TokenKind::Dimension).is_some() {
@@ -412,7 +397,7 @@ impl<'a> Parser<'a> {
         }
     }
 
-    fn parse_let(&mut self) -> Result<Statement> {
+    fn parse_variable(&mut self) -> Result<DefineVariable> {
         if let Some(identifier) = self.match_exact(TokenKind::Identifier) {
             let identifier_span = self.last().unwrap().span;
 
@@ -440,13 +425,13 @@ impl<'a> Parser<'a> {
                 let mut decorators = vec![];
                 std::mem::swap(&mut decorators, &mut self.decorator_stack);
 
-                Ok(Statement::DefineVariable(DefineVariable {
+                Ok(DefineVariable {
                     identifier_span,
                     identifier: identifier.lexeme.clone(),
                     expr,
                     type_annotation,
                     decorators,
-                }))
+                })
             }
         } else {
             Err(ParseError {
@@ -565,6 +550,16 @@ impl<'a> Parser<'a> {
                 Some(self.expression()?)
             };
 
+            let mut local_variables = Vec::new();
+            self.skip_empty_lines();
+            if self.match_exact(TokenKind::Where).is_some() {
+                self.skip_empty_lines();
+                while let Ok(local_variable) = self.parse_variable() {
+                    local_variables.push(local_variable);
+                    self.skip_empty_lines();
+                }
+            }
+
             if decorator::contains_aliases(&self.decorator_stack) {
                 return Err(ParseError {
                     kind: ParseErrorKind::AliasUsedOnFunction,
@@ -581,7 +576,7 @@ impl<'a> Parser<'a> {
                 type_parameters,
                 parameters,
                 body,
-                local_variables: todo!(),
+                local_variables,
                 return_type_annotation,
                 decorators,
             })
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -64,8 +64,8 @@
 
 use crate::arithmetic::{Exponent, Rational};
 use crate::ast::{
-    BinaryOperator, Expression, ProcedureKind, Statement, StringPart, TypeAnnotation,
-    TypeExpression, TypeParameterBound, UnaryOperator,
+    BinaryOperator, DefineVariable, Expression, ProcedureKind, Statement, StringPart,
+    TypeAnnotation, TypeExpression, TypeParameterBound, UnaryOperator,
 };
 use crate::decorator::{self, Decorator};
 use crate::number::Number;
@@ -440,13 +440,13 @@ impl<'a> Parser<'a> {
                 let mut decorators = vec![];
                 std::mem::swap(&mut decorators, &mut self.decorator_stack);
 
-                Ok(Statement::DefineVariable {
+                Ok(Statement::DefineVariable(DefineVariable {
                     identifier_span,
                     identifier: identifier.lexeme.clone(),
                     expr,
                     type_annotation,
                     decorators,
-                })
+                }))
             }
         } else {
             Err(ParseError {
@@ -581,6 +581,7 @@ impl<'a> Parser<'a> {
                 type_parameters,
                 parameters,
                 body,
+                local_variables: todo!(),
                 return_type_annotation,
                 decorators,
             })
@@ -2264,18 +2265,18 @@ mod tests {
     fn variable_definition() {
         parse_as(
             &["let foo = 1", "let foo=1"],
-            Statement::DefineVariable {
+            Statement::DefineVariable(DefineVariable {
                 identifier_span: Span::dummy(),
                 identifier: "foo".into(),
                 expr: scalar!(1.0),
                 type_annotation: None,
                 decorators: Vec::new(),
-            },
+            }),
         );
 
         parse_as(
             &["let x: Length = 1 * meter"],
-            Statement::DefineVariable {
+            Statement::DefineVariable(DefineVariable {
                 identifier_span: Span::dummy(),
                 identifier: "x".into(),
                 expr: binop!(scalar!(1.0), Mul, identifier!("meter")),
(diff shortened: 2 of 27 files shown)


numbat/src/typechecker/mod.rs
```diff
@@ -294,17 +294,19 @@ impl TypeChecker {
...
 
-                for Bound::IsDim(t) in qt.bounds.iter() {
-                    self.constraints.add(Constraint::IsDType(t.clone())).ok();
-                }
+                        for Bound::IsDim(t) in qt.bounds.iter() {
+                            self.constraints.add(Constraint::IsDType(t.clone())).ok();
+                        }
+                        qt.inner
+                    }
+                };
```
irevoire, line 307: I’m really not sure I was allowed to do that, and I don't really understand what it does anyway, so please triple-check this part

sharkdp:
Thank you so much for working on this. I am very excited about this feature.

> The issue with the current implementation is that after parsing a where I must continue to parse as many variable declaration as possible + all whitespace and newlines.

I think Haskell doesn't have this problem since they have a indentation-aware syntax, right? I'm hesitant, but would that be worth thinking about? For example, it might be reasonable to require the variable definitions to be indented more than the `where` keyword itself? Of course that would also mean that the rest of the language becomes somewhat indentation-aware. Because the statement *following* a function definition with a `where`-clause would need to have a lower indentation level than the `where` keyword.

Maybe that is the advantage of `let … in …`?

numbat/modules/core/lists.nbt
```diff
@@ -44,6 +44,12 @@ fn element_at<A>(i: Scalar, xs: List<A>) -> A =
     then head(xs)
     else element_at(i - 1, tail(xs))
 
+@description("Insert the element at index `i` in a list")
+fn insert_at<A>(el: A, i: Scalar, xs: List<A>) -> List<A> =
+  if i == 0
+    then cons(el, xs)
+    else cons(head(xs), insert_at(el, i - 1, tail(xs)))
```
sharkdp, line 51: Is this related to the PR?

irevoire:
> For example, it might be reasonable to require the variable definitions to be indented more than the where keyword itself?

Oh gosh, please no no no this is literally the feature I despise the most over all existing languages 😭 
- It will impact the whole language, as you said
- Make the code harder to write and maintain with dumb bugs over characters we don't see
- Make a code formatter pretty much useless and harder to write

I’m 100% against it.

----

Anyway, I see two ways of fixing the `where` personally:
- After parsing the `where` part of the function, we could push a `\n` as the latest one and let the rest of the code work (this is not supported currently, I believe)
- After parsing the `where`, we could set a boolean somewhere saying that the next time the block of code looking for a trailing character is executed, it should allow all trailing character
- While looking for trailing characters, we check if the latest stmt parsed was a function containing a `where`

None of them are really _that_ ugly IMO, even if it’s not ideal.

I believe the `let ... in` would work without any hacks, yes, but it would also make it slightly harder to share code between the current `let` and the `where` (not a big deal either tbh)

numbat/modules/core/lists.nbt
irevoire: Oops no sorry I’ll remove it

sharkdp:
> Oh gosh, please no no no this is literally the feature I despise the most over all existing languages 😭

Haha, fine. I mostly agree with you, which is why I was hesitant to even mention it. If we can keep the proposed `where` syntax without that, I'm all for it.

> Anyway, I see two ways of fixing the `where` personally:

I can't really judge which of your proposed solutions would be the best, without further looking into it myself. 

Is the problem "only" with the error reporting, or are there actual parsing ambiguities?

irevoire:
> Is the problem "only" with the error reporting, or are there actual parsing ambiguities?

As long as `=` is not allowed in an expression, I think there are no ambiguities. It's only an issue with the error reporting because we're currently expecting all statements to have newlines at the end before the next statement or expression.
But with the `where`, we need to eat all the newlines, which breaks the parsing of statements that want to remove newlines.

sharkdp:
Ok. I'm not 100% sure I understood the problem correctly, but please let me know if I can help move this forward :+1:

irevoire:
> Ok. I'm not 100% sure I understood the problem correctly, but please let me know if I can help move this forward 👍

Ok so I found a fourth solution. I didn't realize that numbat wasn't using a streaming parser and could come back in time easily.
So, I now track the current token before running into an error so I can reset myself to the latest useful character.

If you're ok with the rest of the code I believe we can merge 😄

sharkdp:
I plan to do a review later today hopefully. I added two commits (updating editor syntaxes, rewriting some Numbat code).

Doc generation seems to be broken at the moment. If I run `book/build.py`, the functions with a `where` clause seem to lose their `@decorator` information(?).

numbat/src/parser.rs
```diff
@@ -392,447 +392,493 @@ impl<'a> Parser<'a> {
...
+        } else if self.match_exact(TokenKind::At).is_some() {
+            self.parse_decorators()
+        } else if self.match_exact(TokenKind::Unit).is_some() {
+            self.parse_unit_declaration()
+        } else if self.match_exact(TokenKind::Use).is_some() {
+            self.parse_use()
+        } else if self.match_exact(TokenKind::Struct).is_some() {
+            self.parse_struct()
+        } else if self.match_any(PROCEDURES).is_some() {
+            self.parse_procedure()
```
sharkdp, line 409: Thank you :heart_eyes:. This was overdue.

numbat/src/typechecker/mod.rs
sharkdp: I'll try to explain this. Both for your understanding, but also for my own :smile:

First, we need to understand the difference between concrete and quantified types. All "normal" types like `Scalar`, `Meter`, `String`, `Bool`, `List<List<Bool>>` etc. are concrete. The distinction is only needed when type variables come into play. A generic function like
```
fn id<A>(x: A) -> A = x
```
has a quantified type (scheme) that can be written as `forall A. A -> A`. It is the same type as `forall Z. Z -> Z`. The `A` or the `Z` inside are not "concrete" types, because they are bound by the `forall` *quantifier*.

However, this does not mean that all type variables imply quantified types. During type inference, we need to *instantiate* type schemes like `forall A. A -> A` if a function (or identifier) is being used. For example, imagine someone uses the `id` function twice in an expression:
```
if id(true) then id("yes") else "no"
```
When we encounter `id(true)`, we instantiate the type scheme `forall A. A -> A` with a fresh type variable, meaning we pick a new name for a type variable that doesn't clash with anything that exists, say `T1`. The type of *this particular instance of* `id` then becomes `T1 -> T1`. A concrete type! Similarly, we instantiate the other `id` with a fresh type variable to get `T2 -> T2`; another concrete type. Constraint resolution will then find solutions for those type variables (`T1 ~ Bool`, `T2 ~ String`).

Another thing to understand is the opposite of instantiation: generalization. If you write a function like, say
```
fn f(x) = [x, x]
```
without any type annotations, the type inference algorithm will also pick a fresh type variable for `x` (say `T3`) and arrive at the concrete type `T3 -> List<T3>` for this function. As a final step *after* elaboration and constraint solving, the type checker will then perform generalization. This means it will quantify over all type variables that are not bound by the environment. So the type of `f` finally becomes `forall T3. T3 -> List<T3>`, which is equivalent to `forall A. A -> List<A>` or:
```
fn f<A>(x: A) -> List<A> = [x, x]
```
(try it [here](https://numbat.dev/?q=fn+f%28x%29+%3D+%5Bx%2C+x%5D%E2%8F%8E))

One final thing to realize is that identifiers can *also* have generic types, not just functions. For example:
```
let zero = 0
```
will lead to a type of `forall A. A` for `zero` (Side remark: there is currently no way to write down a generic type annotation for variables in Numbat).

When we use this identifier in an expression like
```
(meter + zero) / (second + zero)
```
we need to *instantiate* the type scheme `forall A. A` for each of these occurrences. In one of them, `zero` will have the type `Length`. In the other, `zero` will be of type `Time`. If we would only instantiate once and only introduce a single type variable, this code would not type check.

---

Now, back to this particular part of the code. The old code assumed that identifiers would always have quantified types. This was true before, since a variable definition would be elaborated + constraint-solved + generalized completely before anyone could use the variable in an expression.

With the change here, we can now have (multiple) variable definitions inside a single statement (the function definition). For example, if we write
```
fn f() = (meter + zero) / (second + zero)
  where zero = 0
```
the type checker will perform elaboration first. It will process the `zero = 0` variable definition and assign a fresh type variable `T4` to zero, because it will *instantiate* the type scheme for the expression `0` (which is `forall A. A`).

Next, it will try to elaborate the body of the function definition *and that will fail*. Because we did not run (constraint solving and) generalization. When `zero` is discovered in the body of the function, it will not have a quantified type of `forall A. A`, but a concrete type of `T4`.

And this is why you got *"Tried to instantiate concrete type"* errors here. The generalization procedure did not run in between the variable definition for `zero` and the elaboration of the function body.

A lot of ML-style languages solve this problem by a procedure called `let`-generalization. This complicates the type checking phase significantly. Especially with more advanced type systems like we have in Numbat (with the physical-dimension types). This is why there is actually a paper called ["Let Should not be Generalised"](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tldi10-vytiniotis.pdf) by no other than Simon Peyton Jones (amongst others), arguing that the above is not actually a big deal. Yes, this looks like it's too restrictive, but it turns out that no one actually writes code like that except for some very special cases.

In the original [type system extension ticket](https://github.com/sharkdp/numbat/issues/433), I already wrote down at the time that we should probably follow the suggestion of that paper, in particular since it mentions "dimension types" as one of the extensions that would be problematic for type systems with `let`-generalization.

In summary: you did everything correctly :smile:. And you were absolutely right in assuming that this was not a trivial thing and should be mentioned :+1:.

numbat/src/typed_ast.rs
```diff
@@ -927,6 +946,31 @@ impl PrettyPrint for Statement {
                     type_parameters.iter().map(|(n, _)| n.clone()).collect(),
                 ));
 
+                let mut pretty_local_variables = None;
+                if !local_variables.is_empty() {
+                    let mut plv = m::keyword("where");
```
sharkdp, line 951: This needs a space in front of `where`:

```
>>> fn f(x) = y where y = x

  fn f<A>(x: A) -> A = ywhere y: forall A. A = x
```

numbat/src/parser.rs
```diff
@@ -2620,6 +2676,58 @@ mod tests {
...
+                    type_annotation: None,
+                    decorators: vec![],
+                }],
+                return_type_annotation: None,
+                decorators: vec![],
+            },
+        );
+
+        parse_as(
+            &["fn kefirausaure(x) = z + y
```
sharkdp, line 2700: ? :smile:

numbat/src/parser.rs
```diff
@@ -392,447 +392,493 @@ impl<'a> Parser<'a> {
...
+                    step_back_to = self.current;
+                    self.skip_empty_lines();
+                    while let Ok(local_variable) = self.parse_variable() {
+                        local_variables.push(local_variable);
+                        step_back_to = self.current;
                         self.skip_empty_lines();
-                        dexprs.push(self.dimension_expression()?);
                     }
+                }
+                self.current = step_back_to;
```
sharkdp, line 583: This might be the problem with the missing decorators. `self.current` is not the only stateful thing in the parser. There's also the `decorator_stack`, which would need to be reset.

numbat/src/parser.rs
```diff
@@ -392,447 +392,493 @@ impl<'a> Parser<'a> {
...
-                }
+                self.skip_empty_lines();
+                let body = self.expression()?;
 
-                if self.match_exact(TokenKind::Equal).is_some() {
-                    self.skip_empty_lines();
-                    let mut dexprs = vec![self.dimension_expression()?];
+                // After parsing the `where` statement we have eaten all
+                // the newlines. We're not supposed to do that.
+                // We'll resume the parser to where we were at before doing that.
```
sharkdp, line 569: > we have eaten all the newlines. We're not supposed to do that.

That is not really the problem though, right? The problem is that we possibly "eat" more than just whitespace. We might have started to parse the beginning of a new statement. For example, in 
```
fn f() = x where x = 1

pi - 3
```
we would continue to parse `pi` as a possible identifier of an additional local variable. But then when we encounter the `-`, we would realize we made a mistake (we would expect a `=` in a variable definition).

I don't know about you, but I'm not super happy with this. The fact that we need to backtrack shows that the grammar is not context free. This will make it harder to write parsers (or "parsers") for Numbat code. For example: editor syntax highlighting integrations.

It's also not ideal that something like
```
fn f(x) = z where z = x * x

not_so_global_variable = 3
```
is parsed without any error, even though it really looks like one. The user probably just forgot to add `let` in front of the second variable definition.

And as you pointed out, this would mean we could not write a streaming parser for Numbat. Something that I initially set out to do, but was always too lazy to finish. It's not a big deal right now, since parsing is quite fast. But it would be another possible downside.

---

Would it make sense to discuss alternative syntax designs again?

Maybe
```
fn f(…) = …
  where
    x = 1,
    y = 2,
    z = 3
```
where the comma would be required between multiple local variable definitions?

Or maybe
```
fn f(…) = …
  where x = 1
    and y = 2
    and z = 3
```

irevoire: Yeah, that definitely seems like a good idea.
I also wanted to make a streaming parser on numbat.

I like that `and` but no strong on my side.
Any preference on your side?

numbat/src/parser.rs
irevoire: Kefir is just the tiny puppy dog I adopted earlier this year. He now lives in all my tests 
image
image
image

irevoire:
Hey, FYI, I have a lot of friends at home currently.
I probably won't work on this before next week

numbat/src/typed_ast.rs
sharkdp: Fixed

numbat/src/typed_ast.rs
```diff
@@ -927,6 +946,31 @@ impl PrettyPrint for Statement {
...
+                    for DefineVariable(
+                        identifier,
+                        _decs,
+                        expr,
+                        _annotation,
+                        _type,
+                        readable_type,
+                    ) in local_variables
+                    {
+                        plv += m::space()
```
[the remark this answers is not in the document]
sharkdp: Fixed

numbat/src/parser.rs
sharkdp: Fixed

numbat/src/parser.rs
sharkdp: This was not the problem, but I fixed that.

numbat/src/parser.rs
```diff
@@ -392,447 +392,493 @@ impl<'a> Parser<'a> {
...
+                let mut step_back_to = self.current;
 
-                    while self.match_exact(TokenKind::Equal).is_some() {
+                let mut local_variables = Vec::new();
+                self.skip_empty_lines();
+                if self.match_exact(TokenKind::Where).is_some() {
+                    step_back_to = self.current;
+                    self.skip_empty_lines();
+                    while let Ok(local_variable) = self.parse_variable() {
```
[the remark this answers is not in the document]
sharkdp: Fixed

sharkdp (approved):
Thank you @irevoire. I solved the remaining TODOs and changed to a `where … and …` syntax. We still need to do some look-ahead in the parser, but we don't have to backtrack.

The `where … and …` syntax does not look as great as what we had before, but it has the advantage that it can be used in a single line, which is an actual win in the terminal, where I always need to think whether it's shift+enter or alt+enter to get a newline without sending the query prematurely.

The lookahead would also be useful for common expressions like
```
let my_result = some * complicated * terms
                + more * complicated * stuff
```
where the `+' is on the next line. But not sure if that's really possible to do with the current mechanism.

irevoire:
Hey awesome, thanks a lot!

numbat/src/typechecker/mod.rs
irevoire: That’s a big explanation, not gonna lie I didn't understand everything but it makes a great resource thanks for writing it down! 
I’ll probably re-read it a few times in the future 😅

sharkdp: Of course. Let me know if you have any questions.

PR #522, opened 2024-08-09
Use egui_plot instead of plotly.rs

sharkdp:
A continuation of the work by @irevoire, now with Bar charts as well

Compared to plotly.js, this adds an even bigger dependency tree to Numbat. It also increases the numbat binary size considerably. And this probably harder to use on the web (?).

I also didn't manage to label the entries in the bar chart according to the x_labels. Setting the name of the `Bar` is not enough, apparently.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,112 @@
 version = 3
 
 [[package]]
+name = "ab_glyph"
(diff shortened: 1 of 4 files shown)


sharkdp:
Closing this in favor of #464 for now.

PR #527, opened 2024-08-09
Improve error message for assert_eq/3

### LINKED ISSUE #505: assert_eq: Better error messages
Consider this example:
```
>>> assert_eq(2m, 201cm, 0.1cm)
error: Assertion failed
  ┌─ <input:2>:1:15
  │
1 │ assert_eq(2m, 201cm, 0.1cm)
  │           --  ^^^^^ 201 cm
  │           │    
  │           2 m
  │
  = Assertion failed because the following two quantities differ by more than 0.1 cm:
      2 m
      201 cm
```
This is not very helpful, since we show the LHS/RHS values in different units.

And if the contrived example is not enough, here's a real world error message that I got today:
```
>>> assert_eq(-77° + 0′ + 32″, -77.0089°, 1e-4°)
error: Assertion failed
  ┌─ <input:1>:1:28
  │
1 │ assert_eq(-77° + 0′ + 32″, -77.0089°, 1e-4°)
  │           ---------------  ^^^^^^^^^ -77.0089°
  │           │                 
  │           -1.34375
  │
  = Assertion failed because the following two quantities differ by more than 0.00000174533:
      -1.34375
      -77.0089°
```

We certainly want to do a unit conversion here (to the same unit that is also used for the actual comparison, I would guess the LHS unit). Maybe we even want to show a comparison in the two possible units (the LHS unit and the RHS unit)?

xmbhasin:
Closes #505

Improves the error message for assert_eq/3 using consistent units for the input comparands (quantities). Before the change, the error message may have caused confusion as different units are used for the comparands.

Converting everything to the units of epsilon is intuitive and allows a user to see the difference in the error message in different units easily by changing the unit of epsilon without having to mess with the lhs or rhs expressions.

With this change, assert_eq/3 will convert comparands to the units of epsilon before comparison and show original units where possible in parentheses.

Example error message after this change:

```
>>> assert_eq(2m, 201cm, 0.1cm to mm)
error: Assertion failed
  ┌─ <input:1>:1:15
  │
1 │ assert_eq(2m, 201cm, 0.1cm to mm)
  │           --  ^^^^^ 2010.0 mm (201 cm)
  │           │
  │           2000 mm (2 m)
  │
  = Assertion failed because the following two quantities differ by 10.0 mm, which is more than or equal to 1 mm:
      2000 mm (2 m)
      2010.0 mm (201 cm)
```

Example error message before this change:

```
>>> assert_eq(2m, 201cm, 0.1cm to mm)
error: Assertion failed
  ┌─ <input:1>:1:15
  │
1 │ assert_eq(2m, 201cm, 0.1cm to mm)
  │           --  ^^^^^ 201 cm
  │           │
  │           2 m
  │
  = Assertion failed because the following two quantities differ by more than 1 mm:
      2 m
      201 cm
 ```

### CODE REVIEW

**Comment by xmbhasin** [comment]
One concern I had was that this could be a breaking change in some edge cases where converting everything to the epsilon units instead of the lhs units could result in a different result from assert_eq/3. I'm not sure if this is possible.

File: numbat/src/ffi/procedures.rs [hunk 1]
```diff
@@ -107,29 +107,42 @@ fn assert_eq(_: &mut ExecutionContext, mut args: Args, arg_spans: Vec<Span>) ->
...
+                } else {
+                    ControlFlow::Break(RuntimeError::AssertEq3Failed(
+                        span_lhs,
+                        lhs_original.clone(),
+                        lhs_converted.clone(),
+                        span_rhs,
+                        rhs_original.clone(),
+                        rhs_converted.clone(),
+                        eps.clone(),
+                        Quantity::new_f64(diff_abs, diff.unit().clone()),
```
xmbhasin, line 142: Is there any simpler way to convert a Quantity to it's absolute value?

sharkdp: Probably not, but feel free to add a `Quantity::abs` method!

examples/runtime_error/assert_eq_3.nbt
```diff
@@ -0,0 +1,2 @@
+# fail approximate equality test with epsilon in units that are different from either comparand
+assert_eq(2m, 201cm, 0.1cm to mm)
```
sharkdp, line 2: Why not
```suggestion
assert_eq(2m, 201cm, 1mm)
```

numbat/src/interpreter.rs
```diff
@@ -30,8 +30,17 @@ pub enum RuntimeError {
...
+    #[error("Assertion failed because the following two quantities differ by {7}, which is more than or equal to {6}:\n  {2} ({1})\n  {5} ({4})")]
+    AssertEq3Failed(
+        Span,
+        Quantity,
+        Quantity,
+        Span,
+        Quantity,
+        Quantity,
+        Quantity,
+        Quantity,
```
sharkdp, line 42: Maybe turn this into a struct (`AssertEq3Error`)? We could then implement `Display` for that struct and use `#[error("{0}")]` here instead.

This would allow us to have a bit more control over how the message looks like. For example, this would allow us to prevent the duplication in a (pretty standard) case like this:
```
>>> assert_eq(200 cm, 202 cm, 1 cm)
error: Assertion failed
  ┌─ <input:10>:1:19
  │
1 │ assert_eq(200 cm, 202 cm, 1 cm)
  │           ------  ^^^^^^ 202 cm (202 cm)
  │           │        
  │           200 cm (200 cm)
  │
  = Assertion failed because the following two quantities differ by 2 cm, which is more than or equal to 1 cm:
      200 cm (200 cm)
      202 cm (202 cm)
```

xmbhasin: A struct sounds good. To check my understanding, do you mean to reduce the noise from printing the lhs and rhs quantities in their original units in parentheses when the units are the same?

Something like this?:
```
>>> assert_eq(2000mm, 201cm, 1mm)
error: Assertion failed
  ┌─ <input:3>:1:19
  │
1 │ assert_eq(2000mm, 201cm, 1mm)
  │           ------  ^^^^^ 2010.0 mm (201 cm)
  │           │
  │           2000 mm
  │
  = Assertion failed because the following two quantities differ by 10.0 mm, which is more than or equal to 1 mm:
      2000 mm

      201 cm (2010.0 mm)
```

examples/runtime_error/assert_eq_3.nbt
xmbhasin: No strong reason, but this makes it easier to vary the units of the comparison and error message without having to change the digits of epsilon. For example, starting with `0.1cm` changing to `1mm` is a little bit harder than just adding `to mm`.

numbat/src/interpreter.rs
sharkdp: Exactly.

I think I would also "more than" instead of "more than or equal". Any change the actual logic accordingly.

sharkdp:
> One concern I had was that this could be a breaking change in some edge cases where converting everything to the epsilon units instead of the lhs units could result in a different result from assert_eq/3. I'm not sure if this is possible.

That could very well be. But I think that should be errors on the order of the floating point precision. I wouldn't worry about it too much. It should only happen if someone has set their threshold to more or less the exact boundary value. Which is never a good idea with floating point math.

But thanks for raising the concern. In general, I'm still quite liberal with breaking changes in Numbat. The language is still very young and I'd rather get the defaults right. Also, I don't think there are any larger existing "code bases" in Numbat — yet! :smile:. Except for the ones in this repo, of course. But those are covered by tests.

sharkdp:
I think it would be great if we could also add a few (integration) tests to `numbat/tests/interpreter.rs`. You can use insta for this. Basically, you add a new test like
```rs
#[test]
fn assert_eq_3() {
    insta::assert_snapshot!(fail("assert_eq(2m, 201cm, 0.1cm)"), @"");
}
```
and then run `cargo insta test --review` which should open an interactive prompt to `a`ccept the current error message and fill it in automatically.

And then we could go through a few different cases by adding more `insta::assert_snapshot!(…)` calls.

xmbhasin:
Would you prefer a single squashed commit or granular commits?

numbat/tests/interpreter.rs
```diff
@@ -817,3 +834,42 @@ fn test_statement_pretty_printing() {
...
+    Assertion failed because the following two quantities differ by 1 cm, which is more than 0.1 cm:
+      200 cm
+      201 cm
+    "###);
+
+    // approximate equality failure with different units for rhs shows original values in parentheses for rhs only
+    insta::assert_snapshot!(fail("assert_eq(2000mm, 201cm, 0.1cm to mm)"), @r###"
+    Assertion failed because the following two quantities differ by 10.0 mm, which is more than 1 mm:
+      2000 mm
+      2010.0 mm (201 cm)
```
sharkdp, line 859: It feels like this could be improved even further by choosing the same number of decimal digits for both values. But maybe that's kind of hard to do properly. So I'm also fine with merging this as is.

sharkdp:
> Would you prefer a single squashed commit or granular commits?

Both are fine. While it's still in review, I definitely prefer granular commits. And I usually just merge/rebase as is, without squashing. Thank you for asking!

numbat/tests/interpreter.rs
xmbhasin: I think it would be better to address this in a new issue/PR as it could be quite complex and may be relevant to and affect other places where pretty printing of numbers is done.

There is some floating point math imprecision which causes the conversion of 201 cm to mm to output `2010.0000000000002`.

This can be seen from a small test like so:
```rust
#[test]
fn conversion_floating_point() {
    let length = Quantity::new_f64(201.0, Unit::centimeter());
    let length = length.convert_to(&Unit::millimeter()).unwrap();

    assert_eq!(length.unsafe_value().to_f64(), 2010.0);
}
```

The `pretty_dtoa` crate formats the floating point value with 6 significant digits in `numbat/src/number.rs::pretty_print`. For `2010.0000000000002`, this should give `2010.00` or so, but additional code trims extra 0s to a single 0.

For a number like `212121001.123456`, it would return `212121000` or `2.12121e+8` in scientific notation, and we may want to change this for clarity depending on epsilon. We may need to account for how many decimal places lhs, rhs, epsilon and maybe delta would have if pretty printed and use the maximum.

I'm not sure how realistic this example is, but you can see here that the pretty printing hides the difference between the lhs and rhs.
```rust
insta::assert_snapshot!(fail("assert_eq(212121000.123456cm, 212121001.123456cm, 0.9cm)"), @r###"
Assertion failed because the following two quantities differ by 1 cm, which is more than 0.9 cm:
  2.12121e+8 cm
  2.12121e+8 cm
"###);
```

Separating these issues would avoid this complexity and prevent it from slowing down the improvement offered here.

xmbhasin: Perhaps I was overthinking it a bit though. We could probably just format all the values to use the number of decimal places in epsilon to keep it simple. There may be some edge cases where epsilon may have more decimal places than expected due to unit conversion or the use of an expression that we can ignore until a more precise underlying type is used for Quantities.

Example:
```rust
insta::assert_snapshot!(fail("assert_eq(2.0000006cm, 243.311336cm, 0.0001mm)"), @r###"
Assertion failed because the following two quantities differ by 2413.1134 mm, which is more than 0.0001 mm:
  0020.0000 mm (2.0 cm)
  2433.1134 mm (243.311 cm)
"###);
```

sharkdp: > We could probably just format all the values to use the number of decimal places in epsilon to keep it simple.

That sounds very reasonable to me (I hope there is a clean way to extract the number of necessary digits from the value of epsilon). Maybe we can just add one digit as a safety margin (unless we want to think this through in more detail and figure out it's not necessary… or that even more digits might be necessary).

xmbhasin:
I wanted to move interpreter.rs into a module folder for better organization (see 
c7fde27). I can revert that commit if you think it would break or affect other open work.

PR #528, opened 2024-08-10
Bump Bytes from yanked version

RossSmyth:
The version of the "bytes" crate that is in the lock file currently is yanked. 

> warning: package `bytes v1.6.0` in Cargo.lock is yanked in registry `crates-io`


sharkdp:
Thank you!

PR #543, opened 2024-08-18
Avoid cloning the typechecker in `elaborate_statement`

GuerricChupin:
Running `numbat` with `cargo flamegraph`, I found that a significant portion of the startup time was spent in the `DefineFunction` branch of `elaborate_statement` in the typechecker just cloning `TypeChecker` (line 1313 of `numbat/src/typechecker/mod.rs`). I think this is only due to the need to save and restore the `environment`, `value_namespace` and `type_namespace` members of `TypeChecker` after we have finished checking the function definition to remove any local names before continuing.

To avoid this, I experimented with replacing the `HashMap`s in `Environment` and `NameSpace` with a stack of `HashMap`s (a `MapStack` for lack of a better name) that can be used to save the current state of the map (when we start checking a definition) and restore it (when we are done) but otherwise behaves like a hash map. This seems to give good results to reduce startup time with little impact on the typechecker's code itself. Running `numbat` on the files in the `example/` directory with `hyperfine`, I observe a ~20-25% reduction in time. I've attached the results as Markdown files for `master` and this branch as well as the output of `cargo flamegraph` for master.

[bench-master.md](https://github.com/user-attachments/files/16648567/bench-master.md)
[bench-avoid-cloning-the-typechecker.md](https://github.com/user-attachments/files/16648566/bench-avoid-cloning-the-typechecker.md)

[flamegraph.svg](https://github.com/user-attachments/assets/a3386bda-2804-46ad-96ce-1d6d9ae745f9)

### DIFF
--- a/numbat/src/typechecker/map_stack.rs
+++ b/numbat/src/typechecker/map_stack.rs
@@ -1,0 +1,82 @@
+use std::hash::Hash;
+use std::{borrow::Borrow, collections::HashMap};
+
+/// A stack of hash maps. All insertions affect the hash map at the top of the
+/// stack (which is the last element of the `stack` vector), preserving any
+/// entries in maps below. The `save` function can be used to push a new map on
+/// the top of the stack, in effect saving the current state of the map, which
+/// one can then restore with `restore`.
+///
+/// The stack vector should never be empty
+#[derive(Debug, Clone)]
+pub(crate) struct MapStack<K, V> {
+    stack: Vec<HashMap<K, V>>,
+}
+
+impl<K, V> Default for MapStack<K, V> {
+    fn default() -> Self {
+        MapStack {
+            stack: vec![Default::default()],
+        }
+    }
+}
+
+impl<K: Hash + Eq, V> MapStack<K, V> {
+    fn iter_dict(&self) -> impl Iterator<Item = &HashMap<K, V>> {
+        self.stack.iter().rev()
+    }
+
+    fn iter_dict_mut(&mut self) -> impl Iterator<Item = &mut HashMap<K, V>> {
+        self.stack.iter_mut().rev()
+    }
+
+    pub(crate) fn iter(&self) -> impl Iterator<Item = (&K, &V)> {
+        self.iter_dict().flatten()
+    }
+
+    pub(crate) fn iter_mut(&mut self) -> impl Iterator<Item = (&K, &mut V)> {
+        self.iter_dict_mut().flatten()
+    }
+
+    pub(crate) fn keys(&self) -> impl Iterator<Item = &K> {
+        self.iter_dict().map(|dict| dict.keys()).flatten()
+    }
+
+    pub(crate) fn get<Q>(&self, key: &Q) -> Option<&V>
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq + ?Sized,
+    {
+        self.iter_dict().filter_map(|dict| dict.get(key)).nth(0)
+    }
+
+    pub(crate) fn insert(&mut self, key: K, value: V) {
+        let _ = self.stack.last_mut().unwrap().insert(key, value);
+    }
+
+    pub(crate) fn contains_key<Q>(&self, key: &Q) -> bool
+    where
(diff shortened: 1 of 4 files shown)


numbat/src/typechecker/map_stack.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::{borrow::Borrow, collections::HashMap};
+
+/// A stack of hash maps. All insertions affect the top hash map, preserving
+/// any entries in maps below. The `save` function can be used to push a new
+/// map on the top of the stack, in effect saving the current state of the map,
+/// which one can then restore with `restore`.
+#[derive(Debug, Clone)]
+pub(crate) struct MapStack<K, V> {
+    top: HashMap<K, V>,
+    stack: Vec<HashMap<K, V>>,
```
sharkdp, line 11: We probably only use two layers at the moment, but it's great to have this more general stack, since we're going to need it once we add nested functions, I guess.

numbat/src/typechecker/map_stack.rs
```diff
@@ -0,0 +1,80 @@
+use std::hash::Hash;
+use std::{borrow::Borrow, collections::HashMap};
+
+/// A stack of hash maps. All insertions affect the top hash map, preserving
+/// any entries in maps below. The `save` function can be used to push a new
+/// map on the top of the stack, in effect saving the current state of the map,
+/// which one can then restore with `restore`.
+#[derive(Debug, Clone)]
+pub(crate) struct MapStack<K, V> {
+    top: HashMap<K, V>,
```
sharkdp, line 10: I wonder why you chose to keep this as a separate element in the struct instead of keeping it in the `Vec`? Doing so might simplify the implementation of some functions below, but probably has other disadvantages?

numbat/src/typechecker/map_stack.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub(crate) fn get<Q>(&self, key: &Q) -> Option<&V>
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq + ?Sized,
+    {
+        self.iter_dict().filter_map(|dict| dict.get(key)).nth(0)
+    }
+
+    pub(crate) fn insert(&mut self, key: K, value: V) -> Option<V> {
+        self.top.insert(key, value)
```
sharkdp, line 53: Does this yield proper semantics? `HashMap::insert` returns `Some(…)` if the map previously contained the value that is being inserted. But you are not checking the other maps on the stack here.

It might not be problematic for our current usage of the map (maybe we never check the return value of `insert`?). But then we should maybe change the return type to `()`?

sharkdp (commented):
Excellent analysis, great idea & execution and very nice results!

I can reproduce your results using `cargo bench` (almost 30% speedup). We have a "micro" benchmark to measure the time it takes to load the "prelude", which is basically equivalent to the startup time of Numbat:

`master`:
```
Import prelude          time:   [71.882 ms 72.311 ms 72.914 ms]
```

this branch:
```
Import prelude          time:   [51.336 ms 51.584 ms 51.917 ms]
                        change: [-29.391% -28.663% -28.060%] (p = 0.00 < 0.05)
                        Performance has improved.
```

I only have two minor questions.

numbat/src/typechecker/map_stack.rs
GuerricChupin: I've changed it to `()`, it makes more sense indeed

numbat/src/typechecker/map_stack.rs
GuerricChupin: Initially it was a bit more convenient to have it this way, because it saved some `unwrap/except` in a bunch of places. However now that the data structure is in its own file, it's not that different. I've changed the implementation to just use a stack with the function `restore` in charge of ensuring its not empty and `insert` panicking if it can't find anything at the top of the stack (which should be impossible because `restore` will panic before). 

Also that allowed me to see a bug in my implementation: the stack vector should be traversed in reverse order, I've fixed `iter_dict` and `iter_dict_mut`.

sharkdp (approved):
Thank you!

Issue #536, opened 2024-08-14
Notepad/soulver-like interface for Numbat?


endigma:
Are there any plans to develop a Notepad/soulver-like interface for Numbat? Something like:

- https://www.figr.app/
- https://numara.io/
- https://soulver.app/
- https://parsify.app/
- https://numi.app/

I think using numbat as the backend for something like this could be really good, most of these struggle with actually manipulating the units they "support", for example:

![image](https://github.com/user-attachments/assets/f842b702-0083-4ba1-9f33-865acc94605a)
or
![image](https://github.com/user-attachments/assets/94e1de4f-8647-40c5-9ec8-b6a9a150c432)

I think this could likely be achieved by essentially re-running numbat with 1..n lines of the file provided and adding the output per-line, however there's probably a more clever way to do this. If it's supported, adding something like `total` or `subtotal` could also be very useful.

sharkdp:
I would love that, too. I mentioned something similar [here](https://libera.irclog.whitequark.org/numbat/2024-06-05#36451572). Something that comes quite close are the interactive examples in my article about the Numbat type system [here](https://numbat.dev/articles/intro.html). The JavaScript code for that ([here](https://github.com/sharkdp/numbat/blob/master/assets/articles/index.js)) can probably serve as a good starting point. It feels like it should be possible to build a PoC rather quickly.

sharkdp:
In order to run this locally, one needs to build the WASM version of Numbat using
````
cd numbat-wasm
bash build.sh
```
And then copy the generated `www/pkg/` folder over to the directory mentioned above.

endigma:
This looks interesting, what do you think of how to get outputs working per-line of the doc? I'm pretty sure implementing it with a for loop basically feeding lines 1..n of the file into numbat would "work", but I'd really rather see some more sane implementation. What if numbat could have some sort of machine readable output mode where it spits out incremental results with source line numbers?

Also, what do you think about impl for stuff like `total` and `subtotal`? I'm not super familiar with how numbat works but I assume you're parsing everything into an AST, does whitespace awareness get lost in that process?

sharkdp:
> This looks interesting, what do you think of how to get outputs working per-line of the doc? I'm pretty sure implementing it with a for loop basically feeding lines 1..n of the file into numbat would "work", but I'd really rather see some more sane implementation. What if numbat could have some sort of machine readable output mode where it spits out incremental results with source line numbers?

[...]

Issue #218, opened 2023-10-29
Website doesn't works well on phone


irevoire:
Hey, it's really hard to use the [website](https://numbat.dev/) on my iPhone.
Basically, everytime I write a letter it scrolls and I can't see what I'm writting
https://github.com/sharkdp/numbat/assets/7032172/431b8e88-08a6-4f7e-9c23-7826f19b1704

Second thing is I can't fix my typos. I basically can't move the cursor at all.

I had the same issues with insect.sh btw

giomf:
Have the similar issues on my Android phone.
When I hit space after a unit, it evaluates the line I just wrote.

Works fine without units.

sharkdp:
I was afraid that this would be the case :frowning_face: 

It sort of works on Firefox for Android — which is the only browser I tested. But even there, it is very far from being fun to work with.

*Any* help in improving this situation would be very much appreciated. Maybe it can be fixed with a few CSS changes, but my CSS "skills" are not enough to work on this myself.

I could also imagine that it is easier to improve the situation by switching to another terminal-in-the-browser library, such as http://xtermjs.org/?

sharkdp:
Just as a side remark.. I am working on a mobile version. I think it makes sense to have a mobile-oriented user interface that does not focus on typing in the same way that the current desktop-oriented interface does with its readline-like interface (which relies on the existence of arrow keys, Ctrl-R, etc).

Sneak peek:

<img src="https://github.com/sharkdp/numbat/assets/4209276/a92adb05-214a-4ac3-b753-8172245c77ca" width="400">

bsidhom:
As a quick workaround, I tried the suggestions [here](https://developer.chrome.com/blog/viewport-resize-behavior/) and listening to `resize` and `orientationchange` events on the window. Unfortunately, the new meta attributes are not supported on iOS Safari (though I'm not sure that would fix the issue anyway). Resize events _do_ seem to fire when the virtual keyboard pops up, but `window.innerSize` _has not updated_ by this point. Instead, the inner size changes _after_ you start typing, but no event is fired at that point. I guess we could use a mutation listener or something to capture it, but this is pretty tricky to get right. The better solution does seem to be to use a mobile-oriented UI. I'm trying to figure out if jQuery Terminal has any features to overcome this but haven't found anything yet.

bsidhom:
[...]

Issue #51, opened 2023-02-13
Jupyter Notebook Integration?


sharkdp:
See https://github.com/sharkdp/numbat/pull/147 for some initial work on this.

PR #550, opened 2024-08-21
Numbat online editor

sharkdp:
This is a first prototype of an online editor for Numbat with "immediate" results on the right hand side panel:

[Try the latest version here](https://numbat.dev/editor.html)

image

Any contributions to the HTML/JS/CSS part a very welcome! I'm not a frontend engineer...


rben01:
I wonder how vertical alignment can work in the face of errors in the results pane. They usually take up many more lines than the input that generated them. Does the Ace editor support making certain lines multiple lines tall?

sharkdp:
> I wonder how vertical alignment can work in the face of errors in the results pane. They usually take up many more lines than the input that generated them. Does the Ace editor support making certain lines multiple lines tall?

Maybe that's okay because we would skip evaluating further lines if there was an error?

Goju-Ryu:
I’m currently experimenting with that actually. It is a problem to be sure and ace doesn't seem to support features that makes it easy. I am however looking into some more 'hacky' solutions like manually setting the height or top of certain lines as a possible fix or if nothing else works add extra lines in the editor to compensate similar to how results are currently handled.

sharkdp:
Maybe we need multiple editors? One for each input?

Goju-Ryu:
> Maybe we need multiple editors? One for each input?

If we do that, then I think a jupyter like interface would be a better fit than the side by side view. It would at least not be trivial to make it feel cohesive if each part was handled as a different editor while keeping the appearance of a single editor for the user.

rben01:
> I’m currently experimenting with that actually. It is a challenge to be sure and ace doesn't seem to support features that makes it easy. I am however looking into some more 'hacky' solutions like manually setting the height or top of certain lines as a possible fix or if nothing else works add extra lines in the editor to compensate similar to how results are currently handled.

I took a quick look at this and Ace resets lines’ `top` on every update, so you'd constantly be fighting it.

Goju-Ryu:
> I took a quick look at this and Ace resets lines’ `top` on every update, so you'd constantly be fighting it.

That's unfortunate. I’m pretty sure they are positioned with absolute as well so setting the height alone wouldn't work either. 
The hack with inserting new lines should work but feels bad and would re-trigger the interpreter in all likelihood. 
I’m considering if there could still be a css fix for it.

sharkdp:
Maybe it's useful to do a bit of research into similar tools (see #536, #19). It's not like we *have* to use ace.js either. It's just something that worked reasonably well for the blog post (https://numbat.dev/articles/intro.html), but I haven't tried anything else.

Goju-Ryu:
> Maybe it's useful to do a bit of research into similar tools (see #536, #19). It's not like we *have* to use ace.js either. It's just something that worked reasonably well for the blog post (https://numbat.dev/articles/intro.html), but I haven't tried anything else.

I think that would probably be a good idea. Getting a more clear direction and finding tools that best fit that vision would probably be necessary in the long term in any case. 

Should we maybe open a discussion about which kinds of UIs/frontends  are desired and how to prioritize their development?

sharkdp:
> Should we maybe open a discussion about which kinds of UIs/frontends are desired and how to prioritize their development?

Sounds great.

The following Numbat "frontends"/clients currently come to mind:

- REPL interface, native (https://github.com/sharkdp/numbat/tree/master/numbat-cli)
- REPL interface, web (https://github.com/sharkdp/numbat/tree/master/numbat-wasm)
- REPL interface, web as PWA (#371)
- REPL interface, mobile (https://github.com/sharkdp/numbat/issues/218#issuecomment-1957084709)
- Editor interface, web (#550)
- Launcher-like interface (https://github.com/MatthiasGrandl/Loungy, https://github.com/MatthiasGrandl/Loungy/blob/main/src/commands/root/numbat.rs)
- Numbat kernel for Jupyter notebooks (#51, #147)
- IRC bot with Numbat plugin (https://github.com/eminence/anna/)

Goju-Ryu:
I think a [Maple](https://www.maplesoft.com/products/maple/) style interface would fit numbat well. It is a variation of the editor interface, but a style that I think would be well suited for many uses of numbat in larger calculations. 

I'll also mention the possibility of a CodeMirror language pack, that while not strictly a frontend for the language, would make it significantly more accessible for third party editors to exist, making use of the wasm or cli frontend to run code and the language pack for editor functionality. This would probably only be worth it currently, if we end up using CodeMirror for the official editor though. 

Using the vs-code extension as a pseudo frontend might also be worth considering. I seem to remember there being a reason it isn't published, but I couldn't find it. If indeed there is something stopping it from being published this suggestion can be disregarded. 
My idea with this is that many scripting languages can be written in run through visual studio code, and would provide an easily accessible way of using numbat locally for longer lived usecases.

sharkdp:
Not really an answer to your post, but maybe something else that might be interesting: This is the editor that is being used in VS code. It's open source: https://microsoft.github.io/monaco-editor/. Maybe this is a better ace alternative?

Goju-Ryu:
> Maybe this is a better ace alternative?

Hmm that is definitely worth checking out. I'll take a look at it and see if it might be a better fit.

endigma:
Monaco is definitely the way to go here, you can probably unify development of any extra features here with the vscode extension if you do so.

PR #551, opened 2024-08-24
add docker support

blairun:
# Add Docker support for web deployment

This pull request adds Docker support to easily deploy the Numbat web application. It includes a `Dockerfile` for building the application and a `docker-compose.yaml` for orchestrating the deployment.

### New file: `Dockerfile`

Added a multi-stage `Dockerfile` that:
1. Builds the WebAssembly package using Rust and wasm-pack.
2. Sets up an Nginx server to host the built files.

### New file: `docker-compose.yaml`

Added a `docker-compose.yaml` file that:
1. Defines the service for the Numbat web application.
2. Provides two options for building:
   a. From the local codebase
   b. Directly from the GitHub repository

## How to use

To deploy the Numbat web application using Docker:

1. Ensure Docker and Docker Compose are installed on your system.
2. Navigate to the project root directory.
3. Run `docker-compose up --build` to build and start the container.
4. Access the application at `http://localhost:8192`.

## Notes

- The `docker-compose.yaml` file includes two build options. Uncomment the preferred option before running.
- The application is exposed on port 8192. Adjust this in the `docker-compose.yaml` file if needed.

## Testing

- [ ] Verify that the Docker build process completes successfully.
- [ ] Confirm that the web application is accessible and functioning correctly when deployed using Docker.
- [ ] Test both build options in the `docker-compose.yaml` file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,8 +19,8 @@ FROM nginx:alpine
 # Copy the built files from the builder stage
 COPY --from=builder /usr/src/numbat/numbat-wasm/www /usr/share/nginx/html
 
-# Expose port 8192
-EXPOSE 8192
+# Expose port 80
+EXPOSE 80
 
 # The default command starts Nginx
 CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+# Use a Rust base image
+FROM rust:1.76.0 as builder
+
+# Install wasm-pack
+RUN curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+
+# Set the working directory
+WORKDIR /usr/src/numbat
+
+# Copy the entire project
+COPY . .
+
+# Build the WebAssembly package
+RUN cd numbat-wasm && wasm-pack build --target web --out-dir www/pkg
+
+# Use a lightweight web server as the final image
+FROM nginx:alpine
+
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you for your contribution.

This works fine, but I wonder if it's really necessary and for what exactly? Once built, Numbat is a static web application. It can be "deployed" by simply copying it to a server. During development, I just run `python -m http.server` to work with it (see https://github.com/sharkdp/numbat/blob/master/numbat-wasm/README.md).

blairun:
Thanks for the feedback.  Docker support is just meant as an alternative for those who prefer containerized deployments.  I prefer deploying numbat this way for its isolation, portability, and ease of setup and maintenance.  Of course, if you don't think it fits in the current scope of the project, I completely understand.

irevoire:
I must say I’m a bit lost here as well, once you have your nginx running in your container you still need to write some conf from the host machine right?
Like, configure the host nginx to do a reverse proxy to your numbat container?
In this case, it always seems easier (and less error-prone) to simply make nginx point to the static website.

Adding support for Docker in this repository means we would have to maintain the Docker container. Your PR doesn’t have a large surface area, but basically, it means we’ll have to update the Docker image as soon as possible and republish it every time Nginx or Alpine has a CVE or something (which happened not that long ago on Alpine iirc).
It seems better to rely on the host nginx directly for security and maintenance on our side imo

sharkdp:
Tentatively closing this for now.

Issue #394, opened 2024-03-12
Functionality to easily copy output of last calculation to clipboard


hamirmahal:

hamirmahal:
I think it'd be useful if there was some sort of keybinding to automatically copy the result of the last calculation to the keyboard. This is a bit more convenient for users who don't want to switch over to precisely highlighting text with a mouse.

hamirmahal:
https://askubuntu.com/questions/759358/select-copy-paste-in-terminal-using-only-the-keyboard/759487#759487 has an interesting option, `screen`, but so far it seems like that just lets me copy and paste in the terminal, whereas I'd like to be able to copy text from the terminal, and then paste it into an editor, which `screen` doesn't seem to do.

If it's not too much work, maybe `numbat` could implement this functionality itself.

rusinikita:
I started using numbat for system design calculation. I think it's useful to have ability to copy in markdown to paste in notion, coda.io, etc.

rusinikita:
Do you know the codebase? I think we can do fork or PR

rben01:
Once PR #555 lands I can work on adding a `copy` command. I need to think about 1. how to recognize passed options, and what options should be passable (`--commas`, `--group-char=' '`, `--scientific`, `--no-units`, ...) I don't really want to implement general CLI parsing here, but output formatting is one place that options abound 2. how to configure default options (env variable? config file?)

sharkdp:
> 1. how to recognize passed options, and what options should be passable (`--commas`, `--group-char=' '`, `--scientific`, `--no-units`, ...) I don't really want to implement general CLI parsing here, but output formatting is one place that options abound

Hm, I really don't think commands should have options. I don't think anyone is going to type something like `copy --group-char=' ' --scientific`. It's faster to just type `copy` and then edit the pasted quantity as needed.

Adding configuration options in Numbat's config file is something we can certainly add if needed, but I would first try to start with a good default, and see how far it takes us.

How that default format for copying should look like is an open question. It's probably *not* the same format that we use for displaying. For example, the underscore-decimal separators are used in some other languages and tools (e.g. Python), but it's far from being a standard. I would probably not add decimal separators at all in the default format.

> 2. how to configure default options (env variable? config file?)

Just like we do with other config options. Via Numbats config file (https://numbat.dev/doc/cli-customization.html#configuration). And possibly via command-line options, if we think this is useful.

rben01:
[...]

PR #555, opened 2024-09-07
Restructured how commands like `list`, `help`, `quit` are parsed to implement `save`

### LINKED ISSUE #187: 'save <filename>' command for the REPL
When a user enters
```
save script.nbt
```
Write all of the accumulated input that we received so far to that file (`script.nbt`). If a user input has resulted in an error, do not add that line to the accumulated input.

rben01:
Closes #187 

To implement the `save` command, I implemented a simple command parser. This handles the following commands:

- `help`
- `info`
- `clear`
- `quit` / `exit`
- `list` / `ls`
- `save` (new)

These were formerly implemented as a bespoke command matcher, matching on e.g., `"list functions" | "ls functions"`. Now they are fully parsed, which both enables more flexibility such as additional whitespace between arguments, and error reporting in the event that a command is misused. For instance, before, `list foo` would result in 

``` text
>>> list foo
error: while type checking
  ┌─ <input:2>:1:1
  │
1 │ list foo
  │ ^^^^ unknown identifier
  │
  = Did you mean 'pint'?
```

Now it results in

```text
>>> list foo
error: while parsing
  ┌─ <input:1>:1:6
  │
1 │ list foo
  │      ^^^ Invalid command: if provided, the argument to `list` or `ls` must be one of: functions, dimensions, variables, units
```

And

```text
>>> help sin
error: while type checking
  ┌─ <input:4>:1:1
  │
1 │ help sin
  │ ^^^^ unknown identifier
  │
  = Did you mean 'hex'?
```

is now

```text
>>> help sin
error: while parsing
  ┌─ <input:6>:1:6
  │
1 │ help sin
  │      ^^^ Invalid command: `help` takes 0 arguments; use `info <item>` for information about an item
```

Similar error reporting is implemented for the other commands.

But all of this was merely done to provide an easy way to add the `save` command — I didn't just want to continue down the path of ad-hoc string parsing. That said, currently the argument to `save` must have no whitespace characters in it, or else it will parse as multiple arguments. Nor does it process backslash-escape characters. I was debating whether to use Tokenizer to parse the argument the same way a String is parsed, but I figured I'd keep this PR small. That can be added later.

The `save` command now looks like

```text
save hist.nbt
```

This will unconditionally overwrite the specified file if it exits. If writing fails, the error looks like

```text
>>> save .
error: runtime error
 = Could not write to file: "."
```

Some improvements that might be desirable in the future:

- Filenames with spaces, presumably using ordinary string syntax (which would also handle backslash escape sequences)
    - Interpolation? Not sure how much utility this would have but it would be natural if using strings as an argument
- Confirmation when overwriting an existing file
- Omitting lines that errored. IIUC this would require some changes to how history is stored, since it only stores strings and not results
- Save per-session history, not entire history over all sessions ever

### DIFF
--- a/numbat/src/command.rs
+++ b/numbat/src/command.rs
@@ -32,7 +32,7 @@ impl CommandKind {
     fn new(word: &str) -> Option<Self> {
         use CommandKind::*;
         Some(match word {
-            "help" => Help,
+            "help" | "?" => Help,
             "info" => Info,
             "list" => List,
             "clear" => Clear,
--- a/numbat/src/command.rs
+++ b/numbat/src/command.rs
@@ -1,0 +1,204 @@
+use crate::{
+    parser::ParseErrorKind,
+    resolver::{CodeSource, Resolver},
+    span::{SourceCodePositition, Span},
+    ParseError,
+};
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum ListItems {
+    Functions,
+    Dimensions,
+    Variables,
+    Units,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum Command<'a> {
+    Help,
+    List { items: Option<ListItems> },
+    Clear,
+    Save { dst: &'a str },
+    Quit,
+}
+
+/// For tracking spans. Contains `(start, start+len)` for each (whitespace-separated)
+/// word in the input
+fn get_word_boundaries(input: &str) -> Vec<(u32, u32)> {
+    let mut word_boundaries = Vec::new();
+    let mut prev_char_was_whitespace = true;
+    let mut start_idx = 0;
+    for (i, c) in input.char_indices() {
+        if prev_char_was_whitespace && !c.is_whitespace() {
+            start_idx = u32::try_from(i).unwrap();
+        } else if !prev_char_was_whitespace && c.is_whitespace() {
+            word_boundaries.push((start_idx, u32::try_from(i).unwrap()));
+        }
+        prev_char_was_whitespace = c.is_whitespace();
+    }
+
+    // if no whitespace after last word, need to add last word
+    if !prev_char_was_whitespace {
+        word_boundaries.push((start_idx, u32::try_from(input.len()).unwrap()));
+    }
+
+    word_boundaries
+}
+
+fn span_from_boundary((start, end): (u32, u32), code_source_id: usize) -> Span {
+    Span {
+        start: SourceCodePositition {
+            byte: start,
+            line: 1,
+            position: start,
+        },
+        end: SourceCodePositition {
+            byte: end,
+            line: 1,
+            position: end,
+        },
+        code_source_id,
+    }
+}
+
+macro_rules! handle_arg_count {
+    (count = 0, $words:expr, $word_boundaries:expr, $code_source_id:expr, "help", $if_ok:expr $(,)?) => {{
+        if $words.next().is_some() {
+            let start = $word_boundaries[1].0;
+            let end = $word_boundaries.last().unwrap().1;
+            return Some(Err(ParseError {
+                kind: ParseErrorKind::InvalidCommand(
+                    "`help` takes 0 arguments; use `info <item>` for information about an item",
+                ),
+                span: span_from_boundary((start, end), $code_source_id),
+            }));
+        }
+
+        $if_ok
+    }};
+    (count = 0, $words:expr, $word_boundaries:expr, $code_source_id:expr, $command:literal, $if_ok:expr $(,)?) => {{
+        if $words.next().is_some() {
+            let start = $word_boundaries[1].0;
+            let end = $word_boundaries.last().unwrap().1;
+            return Some(Err(ParseError {
+                kind: ParseErrorKind::InvalidCommand(concat!("`", $command, "` takes 0 arguments")),
+                span: span_from_boundary((start, end), $code_source_id),
+            }));
+        }
+
+        $if_ok
+    }};
+}
+
+/// Attempt to parse the input as a command, such as "help", "list <args>", "quit", etc
+///
+/// Returns:
+/// - `None`, if the input does not begin with a command keyword
+/// - `Some(Ok(Command))`, if the input is a valid command
+/// - `Some(Err(_))`, if the input starts with a valid command but has the wrong number
+///   or kind of arguments, e.g. `list foobar`
+pub fn parse_command<'a>(
+    input: &'a str,
+    resolver: &mut Resolver,
+) -> Option<Result<Command<'a>, ParseError>> {
+    let word_boundaries = get_word_boundaries(input);
+    let code_source_id = resolver.add_code_source(CodeSource::Text, input);
+
+    let mut words = input.split_whitespace();
+    let Some(command_str) = words.next() else {
+        // should never hit this branch in practice because all-whitespace inputs are
+        // skipped over
+
+        return Some(Err(ParseError {
+            kind: ParseErrorKind::InvalidCommand("invalid empty command"),
+            span: span_from_boundary((0, u32::try_from(input.len()).unwrap()), code_source_id),
+        }));
+    };
+
+    let command = match command_str {
+        "help" => handle_arg_count!(
+            count = 0,
+            words,
+            word_boundaries,
+            code_source_id,
+            "help",
+            Command::Help,
+        ),
+        "clear" => handle_arg_count!(
+            count = 0,
+            words,
+            word_boundaries,
+            code_source_id,
+            "clear",
+            Command::Clear,
+        ),
+        "quit" => handle_arg_count!(
+            count = 0,
+            words,
+            word_boundaries,
+            code_source_id,
+            "quit",
+            Command::Quit,
+        ),
+        "exit" => handle_arg_count!(
+            count = 0,
+            words,
+            word_boundaries,
+            code_source_id,
+            "exit",
+            Command::Quit,
+        ),
+        "list" | "ls" => {
+            let items = match words.next() {
+                None => None,
+                Some("functions") => Some(ListItems::Functions),
+                Some("dimensions") => Some(ListItems::Dimensions),
+                Some("variables") => Some(ListItems::Variables),
+                Some("units") => Some(ListItems::Units),
+                _ => {
+                    return Some(Err(ParseError {
+                        kind: ParseErrorKind::InvalidCommand(
+                            "if provided, the argument to `list` or `ls` must be \
+                             one of: functions, dimensions, variables, units",
+                        ),
+                        span: span_from_boundary(word_boundaries[1], code_source_id),
+                    }));
+                }
+            };
+            if words.next().is_some() {
+                let start = word_boundaries[2].0;
+                let end = word_boundaries.last().unwrap().1;
+                return Some(Err(ParseError {
+                    kind: ParseErrorKind::InvalidCommand("`list` takes at most one argument"),
+                    span: span_from_boundary((start, end), code_source_id),
+                }));
+            }
+
+            Command::List { items }
+        }
+        "save" => {
+            let err_msg = "`save` requires exactly one argument, the destination";
+            let Some(dst) = words.next() else {
+                return Some(Err(ParseError {
+                    kind: ParseErrorKind::InvalidCommand(err_msg),
+                    span: span_from_boundary(word_boundaries[0], code_source_id),
+                }));
+            };
+
+            if words.next().is_some() {
+                let start = word_boundaries[2].0;
+                let end = word_boundaries.last().unwrap().1;
+                return Some(Err(ParseError {
+                    kind: ParseErrorKind::InvalidCommand(err_msg),
+                    span: span_from_boundary((start, end), code_source_id),
+                }));
+            }
+
+            Command::Save { dst }
+        }
+
+        _ => return None,
+    };
+
+    Some(Ok(command))
+}
--- a/numbat/src/command.rs
+++ b/numbat/src/command.rs
@@ -1,6 +1,5 @@
 use crate::{
     parser::ParseErrorKind,
-    resolver::{CodeSource, Resolver},
     span::{SourceCodePositition, Span},
     ParseError,
 };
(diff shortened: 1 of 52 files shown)


sharkdp:
Thank you very much for your contribution — great changeset!



> I implemented a simple command parser

Thanks, this is something I had planned but never got around to it :heart_eyes:. One huge benefit of your approach is that command parsing moves from the application (numbat-cli) to the library (numbat). This opens the possibility to offer command parsing / handling to other "frontends". For example, I think we could simplify this JS code as well:

https://github.com/sharkdp/numbat/blob/8c4d563233d4e1d20edf2cc55ec1733268761feb/numbat-wasm/www/index.js#L41-L76

In order to reduce code duplication even more (among frontends), we could potentially even move part of the command *handling* to the backend (library) as well. The whole "list xyz" => call ctx.print_xyz() logic, for example.

One complication is the following: not all frontends support the same set of commands. `save` would not be supported by the Web frontend, for example. And some commands definitely need frontend-specific code. Like a future `copy` command (#394).

I'm not suggesting that this should happen in this PR, by the way. Only if you are interested.

> Now they are fully parsed, which both enables more flexibility such as additional whitespace between arguments, and error reporting in the event that a command is misused.

Very nice

> That said, currently the argument to `save` must have no whitespace characters in it, or else it will parse as multiple arguments. Nor does it process backslash-escape characters. I was debating whether to use Tokenizer to parse the argument the same way a String is parsed, but I figured I'd keep this PR small. That can be added later.

I agree.

> This will unconditionally overwrite the specified file if it exits.

I think that's completely fine. And maybe even a better UX than asking (interactively) whether the file should be overwritten or not.

numbat/src/command.rs
```diff
@@ -0,0 +1,436 @@
...
+/// Attempt to parse the input as a command, such as "help", "list <args>", "quit", etc
+///
+/// Returns:
+/// - `None`, if the input does not begin with a command keyword
+/// - `Some(Ok(Command))`, if the input is a valid command
+/// - `Some(Err(_))`, if the input starts with a valid command but has the wrong number
+///   or kind of arguments, e.g. `list foobar`
+pub fn parse_command<'a>(
+    input: &'a str,
+    resolver: &mut Resolver,
```
sharkdp, line 255: Instead of passing in a `Resolver` reference, could we also pass in a code_source_id?

numbat/src/command.rs
```diff
@@ -0,0 +1,436 @@
...
+    Info { item: &'a str },
+    List { items: Option<ListItems> },
+    Clear,
+    Save { dst: &'a str },
+    Quit,
+}
+
+/// For tracking spans. Contains `(start, start+len)` for each (whitespace-separated)
+/// word in the input
+fn get_word_boundaries(input: &str) -> Vec<(u32, u32)> {
```
sharkdp, line 28: Could this be implemented using [`str::split_whitespace`](https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace)?

numbat/src/command.rs
```diff
@@ -0,0 +1,436 @@
...
+            span: span_from_boundary((0, u32::try_from(input.len()).unwrap()), code_source_id),
+        }));
+    };
+
+    let command = match command_str {
+        "help" => handle_arg_count!(
+            count = 0,
+            words,
+            &word_boundaries,
+            code_source_id,
```
sharkdp, line 145: I wonder if you could get rid of passing all these arguments every time if you implement the command parser as a `struct` with member variables?

numbat/src/command.rs
rben01: I don't think `split_whitespace` keeps indices, which is what I need. I thought of using `char_indices` and splitting on a predicate, but that would require two `Vec`s which seemed like a needless allocation.

numbat/src/command.rs
rben01: Yes, definitely. That is much cleaner.

I realize however that there is a bug in this PR, which is that for non-command inputs we end up running `resolver.add_code_source` twice, which results in line numbers being off by a factor of ~2 for erroneous non-command inputs. (The precise factor depends on how many commands we've run, since those only trigger `resolve.add_code_source` once.) I'm running into a bit of a chicken and egg problem: to report errors in commands requires a code source up front, but we don't know if we're even parsing a command until we parse at all. An obvious solution is to double-parse — check if we have a command at all (probably quite cheaply, using `split_whitespace` to just grab the first word) and then if we do, fall back to a full parse with a `code_source_id`. This way we only call `resolve.add_code_source` at most once per input. Does this sound good?

sharkdp: > I realize however that there is a bug in this PR, which is that for non-command inputs we end up running `resolver.add_code_source` twice, which results in line numbers being off by a factor of ~2 for erroneous non-command inputs. (The precise factor depends on how many commands we've run, since those only trigger `resolve.add_code_source` once.)

Minor: it's not really the line number that is wrong, but the "code source ID", right? The number in `<input:…>` angle brackets.

> I'm running into a bit of a chicken and egg problem: to report errors in commands requires a code source up front, but we don't know if we're even parsing a command until we parse at all. An obvious solution is to double-parse — check if we have a command at all (probably quite cheaply, using `split_whitespace` to just grab the first word) and then if we do, fall back to a full parse with a `code_source_id`. This way we only call `resolve.add_code_source` at most once per input. Does this sound good?

Hm. That would be a good reason to keep your old solution maybe? Or could you pass in a `get_code_source_id` callable that would only be executed in the error case?

rben01: > Minor: it's not really the line number that is wrong, but the "code source ID", right? The number in <input:…> angle brackets.

Yes, that was the issue.

> Hm. That would be a good reason to keep your old solution maybe? Or could you pass in a get_code_source_id callable that would only be executed in the error case?

My old solution never actually worked. What I did was split the parser into a struct `SourcelessCommandParser` containing just the input, and then the `CommandParser` wraps that and a `code_source_id`. `SourcelessCommandParser` has a fallible initializer that only succeeds if we are in fact looking at a command line. Only if it succeeds do we go ahead and generate the `code_source_id`, and then we have everything we need to parse the command. The key is that you can't construct a `CommandParser` in the first place unless it's been proven that we have a valid command line (meaning it starts with a command word, not that all the arguments are ok), so we never unnecessarily generate a `code_source_id`.

numbat/src/command.rs
```diff
@@ -0,0 +1,472 @@
...
+            }};
+            ($parser:expr, $cmd: literal, $err_msg_suffix:literal) => {{
+                if $parser.inner.words.next().is_some() {
+                    return Err($parser.err_through_end_from(
+                        1,
+                        concat!("`", $cmd, "` takes 0 arguments", $err_msg_suffix),
+                    ));
+                }
+            }};
+        }
```
sharkdp: Do we really need a macro here? If we are okay with a few allocations (this is just the error case, so it won't affect performance), we could do something like this:

```patch
diff --git a/numbat/src/command.rs b/numbat/src/command.rs
index 85070d3..b9da58d 100644
--- a/numbat/src/command.rs
+++ b/numbat/src/command.rs
@@ -137,20 +137,32 @@ impl<'a> CommandParser<'a> {
         }
     }
 
-    fn err_at_idx(&self, index: usize, err_msg: &'static str) -> ParseError {
+    fn err_at_idx(&self, index: usize, err_msg: impl Into<String>) -> ParseError {
         ParseError {
-            kind: ParseErrorKind::InvalidCommand(err_msg),
+            kind: ParseErrorKind::InvalidCommand(err_msg.into()),
             span: self.span_from_boundary(self.inner.word_boundaries[index]),
         }
     }
 
-    fn err_through_end_from(&self, index: usize, err_msg: &'static str) -> ParseError {
+    fn err_through_end_from(&self, index: usize, err_msg: impl Into<String>) -> ParseError {
         ParseError {
-            kind: ParseErrorKind::InvalidCommand(err_msg),
+            kind: ParseErrorKind::InvalidCommand(err_msg.into()),
             span: self.span_through_end(index),
         }
     }
 
+    fn ensure_zero_args(
+        &mut self,
+        cmd: &'static str,
+        err_msg_suffix: &'static str,
+    ) -> Result<(), ParseError> {
+        let message = format!("`{}` takes 0 arguments{}", cmd, err_msg_suffix);
+        if self.inner.words.next().is_some() {
+            return Err(self.err_through_end_from(1, message));
+        }
+        Ok(())
+    }
+
     /// Attempt to parse the input provided to Self::new as a command, such as "help",
     /// "list <args>", "quit", etc
     ///
@@ -159,20 +171,6 @@ impl<'a> CommandParser<'a> {
     /// - `Err(ParseError)`, if the input starts with a valid command but has the wrong
     ///   number or kind of arguments, e.g. `list foobar`
     pub fn parse_command(&mut self) -> Result<Command, ParseError> {
-        macro_rules! ensure_zero_args {
-            ($parser:expr, $cmd:literal) => {{
-                ensure_zero_args!($parser, $cmd, "")
-            }};
-            ($parser:expr, $cmd: literal, $err_msg_suffix:literal) => {{
-                if $parser.inner.words.next().is_some() {
-                    return Err($parser.err_through_end_from(
-                        1,
-                        concat!("`", $cmd, "` takes 0 arguments", $err_msg_suffix),
-                    ));
-                }
-            }};
-        }
-
         // expect() is guaranteed to succeed; CommandParser::new ensures there is at
         // least one word
         let command = match self
@@ -182,23 +180,19 @@ impl<'a> CommandParser<'a> {
             .expect("CommandParser::new is supposed to check that we have a valid command")
         {
             "help" => {
-                ensure_zero_args!(
-                    self,
-                    "help",
-                    "; use `info <item>` for information about an item"
-                );
+                self.ensure_zero_args("help", "; use `info <item>` for information about an item")?;
                 Command::Help
             }
             "clear" => {
-                ensure_zero_args!(self, "clear");
+                self.ensure_zero_args("clear", "")?;
                 Command::Clear
             }
             "quit" => {
-                ensure_zero_args!(self, "quit");
+                self.ensure_zero_args("quit", "")?;
                 Command::Quit
             }
             "exit" => {
-                ensure_zero_args!(self, "exit");
+                self.ensure_zero_args("exit", "")?;
                 Command::Quit
             }
             "info" => {
diff --git a/numbat/src/parser.rs b/numbat/src/parser.rs
index 144473c..9067121 100644
--- a/numbat/src/parser.rs
+++ b/numbat/src/parser.rs
@@ -238,7 +238,7 @@ pub enum ParseErrorKind {
     ExpectedLocalVariableDefinition,
 
     #[error("Invalid command: {0}")]
-    InvalidCommand(&'static str),
+    InvalidCommand(String),
 }
 
 #[derive(Debug, Clone, Error)]
```

numbat/src/command.rs
```diff
@@ -0,0 +1,517 @@
...
+    Functions,
+    Dimensions,
+    Variables,
+    Units,
+}
+
+enum ListAlias {
+    List,
+    Ls,
+}
```
sharkdp, line 20: Not really the topic of your PR, but since we're cleaning up anyway: I think we should get rid of the `ls` alias. It's not in line with the rest of the commands. It's a two-character identifier that someone might want to use for a variable/unit name. And it complicates the code here.

numbat/src/command.rs
rben01: If you're ok with those allocations, then sure.

numbat/src/command.rs
rben01: Done

numbat/src/command.rs
```diff
@@ -0,0 +1,472 @@
...
+    pub fn new(inner: SourcelessCommandParser<'a>, code_source_id: usize) -> Self {
+        Self {
+            inner,
+            code_source_id,
+        }
+    }
+
+    pub fn set_code_source_id(&mut self, code_source_id: usize) {
+        self.code_source_id = code_source_id;
+    }
```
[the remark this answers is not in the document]
rben01: Whoops, ty. Fixed.

rben01:
> One complication is the following: not all frontends support the same set of commands. save would not be supported by the Web frontend, for example. And some commands definitely need frontend-specific code. Like a future copy command (https://github.com/sharkdp/numbat/issues/394).

I will think about how to implement frontend-specific code for a separate PR. Where is the frontend we're using stored at runtime?

sharkdp:
> Save per-session history, not entire history over all sessions ever

Oh, I just found out about this after trying it. I think this is something we should implement before we merge this. Saving the entire history ever (1000 lines for me) is not very helpful.

rben01:
> Oh, I just found out about this after trying it. I think this is something we should implement before we merge this. Saving the entire history ever (1000 lines for me) is not very helpful.

Shouldn't be hard to implement a per-session history. Two questions:

1. Should per-session history be a separate PR (to land before this one) or part of this one? This PR is already quite large, I'm thinking separate.
2. Should the implementation go in `numbat` or `numbat-cli`?

sharkdp:
Sorry if that wasn't clear. I think ideally, the history as we have it right now on master should stay. This is helpful for Ctrl-R searching for commands from previous sessions.

But the save command should only store the commands of the current session. In order to turn a interactive session into a Numbat script that can be fine tuned in a text editor.

numbat/src/command.rs
```diff
@@ -0,0 +1,488 @@
...
+            span: self.span_through_end(index),
+        }
+    }
+
+    fn ensure_zero_args(
+        &mut self,
+        cmd: &'static str,
+        err_msg_suffix: &'static str,
+    ) -> Result<(), ParseError> {
+        let message = format!("`{}` takes 0 arguments{}", cmd, err_msg_suffix);
```
rben01, line 186: Move inside the `if`

numbat/src/command.rs
```diff
@@ -0,0 +1,488 @@
...
+    Help,
+    Info,
+    List,
+    Clear,
+    Save,
+    Quit(QuitAlias),
+}
+
+impl CommandKind {
+    fn new(word: &str) -> Option<Self> {
```
rben01, line 32: Replace with `FromStr`?

numbat/src/command.rs
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    /// Attempt to parse the input provided to Self::new as a command, such as "help",
+    /// "list <args>", "quit", etc
+    ///
+    /// Returns:
+    /// - `Ok(Command)`, if the input is a valid command with correct arguments
+    /// - `Err(ParseError)`, if the input starts with a valid command but has the wrong
+    ///   number or kind of arguments, e.g. `list foobar`
+    pub fn parse_command(&mut self) -> Result<Command, ParseError> {
```
rben01, line 200: Replace with `FromStr`?

numbat/src/command.rs
```diff
@@ -0,0 +1,488 @@
...
+        // expect() is guaranteed to succeed; CommandParser::new ensures there is at
```
rben01, line 201: Delete this out of date comment

numbat/src/command.rs
```diff
@@ -0,0 +1,488 @@
...
+                Command::Info { item }
+            }
+            CommandKind::List => {
+                let items = self.inner.args.next();
+
+                if self.inner.args.next().is_some() {
+                    return Err(self.err_through_end_from(2, "`list` takes at most one argument"));
+                }
+
+                let items = match items {
```
rben01, line 239: Any value to making this case insensitive?

numbat/src/command.rs
rben01: `FromStr` doesn't support borrowing from the input. `TryFrom` not worth it.

numbat/src/command.rs
sharkdp: I don't know. I'd leave it case sensitive for now.

sharkdp:
Thank you very much! Great work.

PR #564, opened 2024-09-15
Initial, simple implementation of saving per-session history

rben01:
This is branched from `master`, not from `commands`, so as to let it evolve in parallel and get feedback on design independently of `commands`. Once merged into `master`, I'll pull `master` into `commands` and update per-session history saving to work with the `Command` framework.

### DIFF
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -404,7 +404,7 @@ impl Cli {
                             }
                             "save" => {
                                 session_history.save(
-                                    "numbat_hist.nbt",
+                                    "history.nbt",
                                     SessionHistoryOptions {
                                         include_err_lines: false,
                                         trim_lines: true,
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -326,9 +326,7 @@ impl Cli {
             }
         }
 
-        let mut session_history = SessionHistory::default();
-
-        let result = self.repl_loop(&mut rl, &mut session_history, interactive);
+        let result = self.repl_loop(&mut rl, interactive);
 
         if interactive {
             rl.save_history(&history_path).context(format!(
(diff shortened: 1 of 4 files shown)


rben01:
Two questions about history, the first of which may apply to current forever-history as well. I wanted some clarity on differences between the forever-history and the session history, in particular prioritizing runnability of the resulting session history file.

1. What to do when an input has a newline in it, such as a function definition? Currently this gets saved in the forever-history as e.g., `fn f(x) = x*y\nwhere y = 3m` (backslash-escaped) but this will not parse when reading the file again. This seems fine for the forever-history (odds are no one will run it, it's just there for searching) but session history may well be run as a file. I'm thinking insert an actual newline (which is what currently happens in this PR), not an escaped newline, in the session history file so that the file remains runnable. The above function definition would then take two lines.
2. Should commands be saved? Again I'm leaning towards no because that makes the session history file un-runnable.

sharkdp:
Thank you for this. It might be two weeks before I can do a full review.

Regarding your questions:

1. Good question. In my opinion, the global history file format is an implementation detail. No one is supposed to make use of the fact that they could be directly executed (except for the newline problem). And I also don't see a use case for that. So is history retrieval works for those multiline inputs, then everything is fine. If that has problems as well, we can also think about changing the format.

For the session history: yes. I would also simply save this as if it had been entered in multiple lines (which is not always possible, but..).

2. Commands should definitely not go into the session history. I'm not even sure what we do for the global history, but thinking about it, I guess it makes sense to include them there. Users might be surprised if they enter something like "info au", and then use up-arrow to get back the "info au" in order to change it to "info as", for example.

rben01:
Currently commands are saved to global history due to https://github.com/sharkdp/numbat/blob/b175c315aee8eede9c481ac8e1b444b79d24867b/numbat-cli/src/main.rs#L347-L348  I agree this is the right call. Given that we're in agreement about the utility of the session history file, I'll leave the implementation more or less as is and just think about cleaning up any tidbits. (For instance I don't like that I've got `parse_and_evaluate` taking an owned input just to be able to add it to the history; will think about how to restructure this.)

rben01:
I've merged this all into the `commands` pull request.

PR #566, opened 2024-09-15
Example decorators

### LINKED ISSUE #503: Add @example decorator
We could add a `@example(…)` decorator for function definitions. This way, we could show exemplary usage of a function. For the documentation, we would (1) run this through the Numbat interpreter and also show the result; and (2) show a "try this" link to https://numbat.dev

Bzero:
This PR resolves #503, implements the `@example` decorator and adds basic examples to most standard library functions.

The `@example` decorator takes the example code as the first positional argument and optionally an example description as second argument. Multiple examples can be defined by using `@example` multiple times:

```nbt
@example("example_function(0)")
@example("example_function(1)", "Evaluate the function at $1$.")
fn example_function(x) = x
```

At the moment the examples are only used to compile the documentation, however it might be nice to make them accessible in the CLI or web interface in the future. In the documentation the example code is run trough the Numbat interpreter and is shown together with the results as if it was typed in the CLI interface. Additionally the example description is displayed (if available) and a "Run this example" link is shown. Since the examples can be a bit lengthy, I chose to display them collapsed by default.

I have not added any examples to non pure functions like `random()`, or `now()` for the moment. While this would work in principle, it would modify the documentation every time someone re-runs `book/build.py` which doesn't feel very practical.


numbat/examples/inspect.rs
```diff
@@ -86,15 +79,126 @@ fn inspect_functions_in_module(ctx: &Context, module: String) {
...
+                }
+                for l in example_output.lines() {
+                    println!("    {}", l);
+                }
+                println!("  ```");
+            } else {
+                eprintln!(
+                        "Warning: Example \"{example_code}\" of function {fn_name} did not run successfully."
+                    );
+            }
```
Bzero, line 171: I am not sure if this is the right choice here. Most often it is useful to check that examples run successfully, however in some cases one may want to deliberately trigger an error.

numbat/examples/inspect.rs
```diff
@@ -86,15 +79,126 @@ fn inspect_functions_in_module(ctx: &Context, module: String) {
...
+        }
+        println!("</details>");
+        println!();
```
sharkdp, line 174: I think this needs to go inside the conditional.

sharkdp (commented):
Great work, as always! Thank you. It might take a while (> 2 weeks) before I can do a full review.

numbat/examples/inspect.rs
Bzero: Yes, thanks for catching this.

numbat/examples/inspect.rs
```diff
@@ -86,15 +79,126 @@ fn inspect_functions_in_module(ctx: &Context, module: String) {
...
+        if !examples.is_empty() {
+            println!("<details>");
+            println!("<summary>Examples</summary>");
+            println!();
+
+            for (example_code, example_description) in examples {
+                let mut example_ctx = prepare_context();
+                let _result = example_ctx
+                    .interpret("use prelude", CodeSource::Internal)
+                    .unwrap();
```
sharkdp, line 92: This could be done just once for the entire program. Then we could create a new example context by cloning an existing one (where the prelude has already been included). Parsing, type checking, interpreting the prelude for every single example seems a bit wasteful and currently takes ~50 ms, which would lead to an overhead of 5 seconds for 100 examples.

numbat/modules/core/functions.nbt
```diff
@@ -2,64 +2,85 @@ use core::scalar
 
 @name("Identity function")
 @description("Return the input value.")
+@example("id(8kg)")
```
sharkdp, line 5: Minor nitpick: can we please be consistent with how we format quantities? The official SI version is to have a space between the number and the unit:
```suggestion
@example("id(8 kg)")
```

numbat/modules/core/functions.nbt
```diff
@@ -2,64 +2,85 @@ use core::scalar
...
 fn id<A>(x: A) -> A = x
 
 @name("Absolute value")
 @description("Return the absolute value $|x|$ of the input. This works for quantities, too: `abs(-5 m) = 5 m`.")
 @url("https://doc.rust-lang.org/std/primitive.f64.html#method.abs")
+# @example("abs(-22.2m)")
```
sharkdp, line 11: This is now fixed, thanks to you.
```suggestion
@example("abs(-22.2 m)")
```

numbat/modules/core/functions.nbt
```diff
@@ -2,64 +2,85 @@ use core::scalar
...
 fn abs<T: Dim>(x: T) -> T
 
 @name("Square root")
 @description("Return the square root $\\sqrt\{x\}$ of the input: `sqrt(121 m^2) = 11 m`.")
 @url("https://en.wikipedia.org/wiki/Square_root")
+@example("sqrt(4are) -> m")
```
sharkdp, line 17:
```suggestion
@example("sqrt(4 are) -> m")
```

numbat/modules/core/functions.nbt
```diff
@@ -2,64 +2,85 @@ use core::scalar
...
 fn sqrt<D: Dim>(x: D^2) -> D = x^(1/2)
 
 @name("Cube root")
 @description("Return the cube root $\\sqrt[3]\{x\}$ of the input: `cbrt(8 m^3) = 2 m`.")
 @url("https://en.wikipedia.org/wiki/Cube_root")
+@example("cbrt(8l) -> cm")
```
sharkdp, line 23:
```suggestion
@example("cbrt(8 L) -> cm")
```

numbat/modules/core/lists.nbt
```diff
@@ -119,12 +139,14 @@ fn _linspace_helper(start, end, n_steps, i) =
     else cons(start + (end - start) * i / (n_steps - 1), _linspace_helper(start, end, n_steps, i + 1))
 
 @description("Generate a list of `n_steps` evenly spaced numbers from `start` to `end` (inclusive)")
+@example("linspace(0, 10, 11)")
```
sharkdp, line 142: Maybe
```suggestion
@example("linspace(-5 m, 5 m, 11)")
```

numbat/modules/core/strings.nbt
```diff
@@ -3,27 +3,35 @@ use core::functions
...
 @description("Convert a string to lowercase")
+@example("lowercase(\"Numbat\")")
 fn lowercase(s: String) -> String
 
 @description("Convert a string to uppercase")
+@example("uppercase(\"Numbat\")")
 fn uppercase(s: String) -> String
 
 @description("Concatenate two strings")
+@example("str_append(\"Numbat\",\"!\")")
```
sharkdp, line 30:
```suggestion
@example("str_append(\"Numbat\", \"!\")")
```

numbat/examples/inspect.rs
Bzero: Yes, that makes sense.

numbat/modules/core/functions.nbt
```diff
@@ -2,64 +2,85 @@ use core::scalar
...
 
 @name("Ceil function")
 @description("Returns the smallest integer greater than or equal to $x$. See also: `ceil_in`.")
 @url("https://doc.rust-lang.org/std/primitive.f64.html#method.ceil")
+@example("ceil(5.5)")
 fn ceil(x: Scalar) -> Scalar
 
 @name("Ceil function")
-@description("Returns the smallest integer multuple of `base` greater than or equal to `value`. For example: `ceil_in(m, 5.3 m) == 6 m`.")
+@description("Returns the smallest integer multuple of `base` greater than or equal to `value`.")
```
sharkdp, line 63: Unrelated typo (not yours): multuple

numbat/modules/core/lists.nbt
```diff
@@ -88,6 +104,7 @@ fn _merge(xs, ys, cmp) =
 
 
 @description("Sort a list of elements, using the given key function that maps the element to a quantity")
+@example("fn map_fn(x) = mod(x, 10)\nsort_by_key(map_fn, [701, 313, 9999, 4])","Sort by last digit.")
```
sharkdp, line 107: Maybe rename "map_fn" to "last_digit"? And possibly remove the description?

numbat/modules/core/lists.nbt
```diff
@@ -110,6 +129,7 @@ fn intersperse<A>(sep: A, xs: List<A>) -> List<A> =
 
 fn _add(x, y) = x + y # TODO: replace this with a local function once we support them
 @description("Sum all elements of a list")
+@example("sum([3, 2, 1])")
```
sharkdp, line 132: Maybe use 3 m, 200 cm and 1000 mm as values?

sharkdp:
Thank you for the update. Will continue my review over the next days.

This is great!

numbat/examples/inspect.rs
sharkdp: I think we should fail hard here (instead of just showing a warning). Those examples are a great source of additional tests. Either we fail the generation of the documentation here (easy option), or we write a new unit test that checks all `@example`s (harder option, probably nicer for future development, since the feedback loop is quicker).

If we ever need failing examples (e.g. if we want to add an `@example` for the `error` function), we can think of a solution then. I don't see a strong need for this right now.

numbat/examples/inspect.rs
```diff
@@ -86,25 +79,137 @@ fn inspect_functions_in_module(ctx: &Context, module: String) {
...
+                        "* {}\n\n  <a href=\"{}\"><i class=\"fa fa-play\"></i> Run this example</a>",
+                        replace_equation_delimiters(example_description),
+                        example_url
+                    );
+                    } else {
+                        println!(
+                            "* <a href=\"{}\"><i class=\"fa fa-play\"></i> Run this example</a>\n",
+                            example_url
+                        );
+                    }
```
sharkdp, line 152: Can we please move the "Run this example" links *below* the code snippets? I think that would look better because the "play" icons wouldn't line up with the bullets. And most of the time, users will only read the example output instead of trying it interactively.

image

numbat/examples/inspect.rs
```diff
@@ -86,25 +79,137 @@ fn inspect_functions_in_module(ctx: &Context, module: String) {
...
+
+                    println!("  ```nbt");
+                    for l in example_input.lines() {
+                        println!("    {}", l);
+                    }
+                    for l in example_output.lines() {
+                        println!("    {}", l);
+                    }
+                    println!("  ```");
```
sharkdp, line 161: Can we trim the final newline in the examples? That would make the output a bit more concise.

image

numbat/examples/inspect.rs
```diff
@@ -86,25 +79,137 @@ fn inspect_functions_in_module(ctx: &Context, module: String) {
...
+                    );
+
+                    //Assemble the example output
+                    let mut example_output = String::new();
+                    example_output += "\n";
+
+                    for statement in &statements {
+                        example_output += &plain_text_format(&statement.pretty_print(), true);
+                        example_output += "\n\n";
+                    }
```
sharkdp, line 126: Do you think we could skip pretty printing, or do you see any problems with this for some of the examples? For most of the examples, I think it only adds visual noise:

image

numbat/examples/inspect.rs
```diff
@@ -86,25 +79,137 @@ fn inspect_functions_in_module(ctx: &Context, module: String) {
...
+            println!("</details>");
+            println!();
+        }
     }
 }
 
-fn main() {
-    let module_path = Path::new(&std::env::var_os("CARGO_MANIFEST_DIR").unwrap()).join("modules");
+// Replace $..$ with \\( .. \\) for mdbook.
+fn replace_equation_delimiters(text_in: String) -> String {
```
sharkdp, line 175: Thanks!

numbat/modules/core/lists.nbt
```diff
@@ -3,67 +3,82 @@ use core::error
...
 
 @description("Generate a new list by applying a function to each element of the input list")
+@example("map(sqr, [3, 2, 1])", "Square all elements of a list.")
 fn map<A, B>(f: Fn[(A) -> B], xs: List<A>) -> List<B> =
   if is_empty(xs)
     then []
     else cons(f(head(xs)), map(f, tail(xs)))
 
 @description("Filter a list by a predicate")
+@example("fn filter_fn(x) = x > 1\nfilter(filter_fn, [3, 2, 1, 0])", "Filter all elements greater than $1$.")
```
sharkdp, line 81: We could maybe use an existing predicate function to make the example a bit more concise?
```suggestion
@example("filter(is_finite, [0, 1e10, sqrt(-1), asin(2)])")
```

numbat/modules/core/strings.nbt
```diff
@@ -3,27 +3,35 @@ use core::functions
 use core::error
 
 @description("The length of a string")
+@example("str_length(\"Numbat\")")
 fn str_length(s: String) -> Scalar
 
 @description("Subslice of a string")
+@example("str_slice(\"Numbat\", 0, 2)")
```
sharkdp, line 10: Maybe use different arguments here to show that the second argument is not the length, but the end-index:
```suggestion
@example("str_slice(\"Numbat\", 3, 6)")
```

numbat/modules/math/geometry.nbt
```diff
@@ -2,9 +2,11 @@ use core::functions
 use math::constants
 
 @description("The length of the hypotenuse of a right-angled triangle $\\sqrt\{x^2+y^2\}$.")
+@example("hypot2(3, 4)")
 fn hypot2<T: Dim>(x: T, y: T) -> T = sqrt(x^2 + y^2)
 
 @description("The Euclidean norm of a 3D vector $\\sqrt\{x^2+y^2+z^2\}$.")
+@example("hypot3(4, 1, 4)")
```
sharkdp, line 9: We could use a [Pythagorean quadruple](https://en.wikipedia.org/wiki/Pythagorean_quadruple) as an example :smiley: 

```suggestion
@example("hypot3(8, 9, 12)")
```

numbat/modules/math/geometry.nbt
```diff
@@ -2,9 +2,11 @@ use core::functions
 use math::constants
 
 @description("The length of the hypotenuse of a right-angled triangle $\\sqrt\{x^2+y^2\}$.")
+@example("hypot2(3, 4)")
```
sharkdp, line 5: One nice thing is that `hypot2` (and 3) take dimensionful arguments, so maybe we could use
```suggestion
@example("hypot2(3 m, 4 m)")
```
here?

numbat/modules/math/number_theory.nbt
```diff
@@ -4,6 +4,7 @@ use core::functions
 @name("Greatest common divisor")
 @description("The largest positive integer that divides each of the integers $a$ and $b$.")
 @url("https://en.wikipedia.org/wiki/Greatest_common_divisor")
+@example("gcd(60,42)")
```
sharkdp, line 7: Minor: 
```suggestion
@example("gcd(60, 42)")
```

numbat/modules/numerics/diff.nbt
```diff
@@ -3,6 +3,7 @@ use core::quantities
 @name("Numerical differentiation")
 @url("https://en.wikipedia.org/wiki/Numerical_differentiation")
 @description("Compute the numerical derivative of the function $f$ at point $x$ using the central difference method.")
+@example("fn polynomial(x) = x² -x -1\ndiff(polynomial, 1)", "Compute the drivative of $f(x) = x² -x -1$ at $x=1$.")
```
sharkdp, line 6: Maybe we could also add a dimensionful example?
```suggestion
@example("fn polynomial(x) = x² - x - 1\ndiff(polynomial, 1)", "Compute the derivative of $f(x) = x² -x -1$ at $x=1$.")
@example("fn distance(t) = 0.5 g0 t²\nfn velocity(t) = diff(distance, t)\nvelocity(2 s)")
```

numbat/modules/physics/temperature_conversion.nbt
```diff
@@ -5,20 +5,24 @@ use units::si
 let _offset_celsius = 273.15
 
 @description("Converts from degree Celsius (°C) to Kelvin.")
+@example("from_celsius(300)", "$300 °C$ in Kelvin.")
```
sharkdp, line 8: Minor nitpick: Formatting this using LaTeX like this leads to an italic `C` and too little space between the number and the unit (in the examples below). The proper style would use `\text{C}` and use `\,` for whitespace between the number and the unit (unless it's °C/°F). So maybe just:
```suggestion
@example("from_celsius(300)", "300 °C in Kelvin.")
```

numbat/modules/physics/temperature_conversion.nbt
```diff
@@ -5,20 +5,24 @@ use units::si
...
 @url("https://en.wikipedia.org/wiki/Conversion_of_scales_of_temperature")
 fn from_celsius(t_celsius: Scalar) -> Temperature = (t_celsius + _offset_celsius) kelvin
 
 @description("Converts from Kelvin to degree Celcius (°C). This can be used on the right hand side of a conversion operator: `200 K -> celsius`.")
+@example("300K -> celsius", "$300K$ in degree Celsius.")
```
sharkdp, line 13:
```suggestion
@example("300K -> celsius", "300 K in degree Celsius.")
```

numbat/modules/physics/temperature_conversion.nbt
```diff
@@ -5,20 +5,24 @@ use units::si
...
 @url("https://en.wikipedia.org/wiki/Conversion_of_scales_of_temperature")
 fn celsius(t_kelvin: Temperature) -> Scalar = t_kelvin / kelvin - _offset_celsius
 
 let _offset_fahrenheit = 459.67
 let _scale_fahrenheit = 5 / 9
 
 @description("Converts from degree Fahrenheit (°F) to Kelvin.")
+@example("from_fahrenheit(300)", "$300 °F$ in Kelvin.")
```
sharkdp, line 21:
```suggestion
@example("from_fahrenheit(300)", "300 °F in Kelvin.")
```

numbat/modules/physics/temperature_conversion.nbt
```diff
@@ -5,20 +5,24 @@ use units::si
...
 @url("https://en.wikipedia.org/wiki/Conversion_of_scales_of_temperature")
 fn from_fahrenheit(t_fahrenheit: Scalar) -> Temperature = ((t_fahrenheit + _offset_fahrenheit) × _scale_fahrenheit) kelvin
 
 @description("Converts from Kelvin to degree Fahrenheit (°F). This can be used on the right hand side of a conversion operator: `200 K -> fahrenheit`.")
+@example("300K -> fahrenheit", "$300K$ in degree Fahrenheit.")
```
sharkdp, line 26:
```suggestion
@example("300K -> fahrenheit", "300 K in degree Fahrenheit.")
```

numbat/modules/units/mixed.nbt
```diff
@@ -5,23 +5,27 @@ use units::imperial
 @name("Degrees, minutes, seconds")
 @description("Convert an angle to a mixed degrees, (arc)minutes, and (arc)seconds representation. Also called sexagesimal degree notation.")
 @url("https://en.wikipedia.org/wiki/Sexagesimal_degree")
+@example("DMS(46.5858°)")
```
sharkdp, line 8:
```suggestion
@example("46.5858° -> DMS")
```

numbat/modules/units/mixed.nbt
```diff
@@ -5,23 +5,27 @@ use units::imperial
...
 fn DMS(alpha: Angle) -> String =
   _mixed_units(alpha, [deg, arcmin, arcsec], ["° ", "′ ", "″"], true)
 
 @name("Degrees, decimal minutes")
 @description("Convert an angle to a mixed degrees and decimal minutes representation.")
+@example("DM(46.5858°)")
```
sharkdp, line 14:
```suggestion
@example("46.5858° -> DM")
```

numbat/modules/units/mixed.nbt
```diff
@@ -5,23 +5,27 @@ use units::imperial
...
 @url("https://en.wikipedia.org/wiki/Decimal_degrees")
 fn DM(alpha: Angle) -> String =
   _mixed_units(alpha, [deg, arcmin], ["° ", "′"], false)
 
 @name("Feet and inches")
 @description("Convert a length to a mixed feet and inches representation.")
 @url("https://en.wikipedia.org/wiki/Foot_(unit)")
+@example("feet_and_inches(180cm)")
```
sharkdp, line 22:
```suggestion
@example("180cm -> feet_and_inches")
```

numbat/modules/units/mixed.nbt
```diff
@@ -5,23 +5,27 @@ use units::imperial
...
 fn feet_and_inches(length: Length) -> String =
   _mixed_units(length, [foot, inch], [" ft ", " in"], false)
 
 @name("Pounds and ounces")
 @description("Convert a mass to a mixed pounds and ounces representation.")
 @url("https://en.wikipedia.org/wiki/Pound_(mass)")
+@example("pounds_and_ounces(1kg)")
```
sharkdp, line 29:
```suggestion
@example("1 kg -> pounds_and_ounces")
```

sharkdp (commented):
I finished my review. Again: thank you very much for implementing this — and adding examples to all existing functions!

numbat/examples/inspect.rs
Bzero: Failing hard makes sense here I think, examples in the standard library should really be enforced to run without errors. I will change that.

I am not so sure how suitable the examples are as tests however. They do provide some basic check if the code runs without error but we can't assert that functions returns the right result so there is still a lot of room to make undetected mistakes.

> e.g. if we want to add an `@example` for the `error` function

Yes, this was the only real example that came to mind maybe next to hypothetical examples of how not to use a function so I would agree that there is not much need for it right now.

Bzero:
> I finished my review. Again: thank you very much for implementing this — and adding examples to all existing functions!

Thanks a lot for the review.

numbat/modules/core/lists.nbt
Bzero: Being more concise would be nice but in my opinion this actually makes the example more complicated than it was before.

numbat/examples/inspect.rs
Bzero: Sure, that is fine for me.

numbat/examples/inspect.rs
Bzero: Yes or course, I initially tried to get the output exactly as it would appear if the input was typed in an interactive session where there is the extra empty line at the end but it does make sense to save some space.

numbat/examples/inspect.rs
Bzero: I think pretty printing can be helpful for some examples, e.g to see how functions are typed, but for most examples it doesn't add much value so I am fine to remove it to make the examples more concise.

numbat/modules/core/lists.nbt
sharkdp: Yes, that's a good idea.

numbat/examples/inspect.rs
sharkdp: > I am fine to remove it to make the examples more concise.

That would be great, thanks.

numbat/examples/inspect.rs
Bzero: Actually, after trying it out I think it doesn't look exactly nice with the "Run this example" links below the code snippets either, at least when there is no example description. Maybe we should not use a bullet list here at all.

image

Bzero: I changed the display and tried to mimic the mdBook Rust Playground feature "play button":

image

The implementation is a little hacky as it requires to insert html code blocks in the markdown documents, however since they are entirely autogenerated I think it should be okay.

sharkdp: Really nice! Thank you :+1:

PR #569, opened 2024-09-18
Boxed errors

rben01:
clippy warned that some `Result` types were quite large due to their massive error variant. For instance, `interpreter::RuntimeError` is 304 bytes, which means the following function returns a whopping 304 bytes (I'm assuming there's a niche somewhere; if not, then 305) despite its `Ok` variant having no payload!

https://github.com/sharkdp/numbat/blob/b175c315aee8eede9c481ac8e1b444b79d24867b/numbat/src/bytecode_interpreter.rs#L308-L312

The same goes for `TypeCheckError` (264 bytes) and `NumbatError` (also 304 as it wraps `RuntimeError`). This seemed like a lot of data to allocate space to the error case.

I took clippy’s advice and boxed the errors so that they'd take 8 bytes at most. This obviously adds an indirection to the error path but uses much less stack space for the `Ok` path.


sharkdp:
Thank you very much!

rben01:
> Thank you very much! It would be great if you could resolve the conflict (or let me know if you need help).

Done!

sharkdp:
The wasm build is currently failing. There was an issue on master which I fixed, but now the failure seems to be related to this branch.

rben01:
Should be good to go now. I was ignoring WASM build errors before but yeah, this one was legit. For whatever reason rust-analyzer can't analyze the numbat-wasm crate, maybe because it can't see through `#[wasm_bindgen]` (?), so I wasn't getting any error messages.

sharkdp:
> For whatever reason rust-analyzer can't analyze the numbat-wasm crate

The wasm package is not part of the workspace (there was a technical reason why this didn't work, but maybe it's worth revisiting). That's why rust analyzer doesn't see it. It works if you open the numbat-wasm project / folder.

PR #571, opened 2024-09-18
Add lowercase aliases for currency units

### LINKED ISSUE #441: Add lowercase aliases for currencies?
As brought up in https://github.com/sharkdp/numbat/issues/319#issuecomment-1931630969, do you think there are any good reasons not to add lowercase aliases for currencies (e.g. `usd`, `gbp`), for ease of use?

triallax:
Fixes #441 

i didn't add  `myr`/`rm` because they conflict with `milliyear`/ `rontometer` respectively, not sure whether to special-case these ones to resolve to the currencies instead


sharkdp:
Thank you!

sharkdp:
Would be great if we could add `eur` as well.

PR #572, opened 2024-09-20
Dimensionful `abs` function

### LINKED ISSUE #565: Panic when using `abs()` with a non-scalar argument
The `abs()` function panics when attempting to use it on a non-scalar argument:

```nbt
>>> abs(2)

  abs(2)

    = 2

>>> abs(2s)
thread 'main' panicked at numbat/src/ffi/math.rs:28:1:
called `Result::unwrap()` on an `Err` value: IncompatibleUnits(Product { factors: [UnitFactor { unit_id: UnitIdentifier { name: "second", canonical_name: CanonicalName { name: "s", accepts_prefix: AcceptsPrefix { short: true, long: false } }, kind: Base }, prefix: Metric(0), exponent: Ratio { numer: 1, denom: 1 } }] }, Product { factors: [] })
```

I think this is a regression as I am quite sure this used to work fine in the past.

Bzero:
Resolves #565.


sharkdp (approved):
Thank you!

PR #578, opened 2024-09-24
Shrink span

rben01:
Span is now 16 bytes instead of 32 (on a 64 bit machine) due to the removal of fields `line` and `position` from `SourceCodePositition` [sic]. Related: #559

This was a remarkably simple change. I was surprised to find that none of the error reporting actually relied on `line` and `position`; it was all just byte-based. Those fields were only used in tests. If later desired, we can hang onto cumulative line lengths and convert byte index into line + position as needed. (But apparently that is not currently needed.)

One thing to note is that tests that were formerly based on `position` are now based on `byte`, which means they're off by 1 (position `n` became byte `n-1`).

### DIFF
--- a/numbat/src/command.rs
+++ b/numbat/src/command.rs
@@ -2,7 +2,7 @@ use std::str::{FromStr, SplitWhitespace};
 
 use crate::{
     parser::ParseErrorKind,
-    span::{SourceCodePositition, Span},
+    span::{ByteIndex, Span},
     ParseError,
 };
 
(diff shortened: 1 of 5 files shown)


sharkdp:
> This was a remarkably simple change. I was surprised to find that none of the error reporting actually relied on `line` and `position`; it was all just byte-based.

Oh, haha :smile:. I guess it might have been used at an earlier stage. Or maybe I just implemented it because I thought it would be required, and then forgot about it. Thank you very much for all your changes. This is really valuable work!

sharkdp (approved):

Issue #525, opened 2024-08-09
Improve startup time (regression)


sharkdp:
The new type checker from #443 introduced a severe regression in startup times:

```
hyperfine --warmup 3 --reference "./numbat-master -e '1+2'" "./numbat-1.12 -e '1+2'" --export-markdown -
```

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `./numbat-master -e '1+2'` | 78.9 ± 1.2 | 77.1 | 83.8 | 2.28 ± 0.09 |
| `./numbat-1.12 -e '1+2'` | 34.5 ± 1.2 | 32.4 | 42.5 | 1.00 |

We should look into potential measures to improve this again.

If the constraint checker is the culprit, we might want to look into de-duplication of constraints before attempting to solve them. But it might be something else entirely (e.g. the mutation/traversal/cloning of ASTs during elaboration)

sharkdp:
Thanks to the changes in #543 by @GuerricChupin, we've moved a significant step in the right direciton:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `./numbat-master -e '1+2'` | 59.9 ± 1.2 | 57.9 | 64.3 | 1.78 ± 0.08 |
| `./numbat-1.12 -e '1+2'` | 33.7 ± 1.3 | 32.6 | 44.8 | 1.00 |

sharkdp:
Thanks to the changes in #537 by @irevoire, we've moved another significant step in the right direction:

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `./numbat-master -e '1+2'` | 47.0 ± 0.9 | 45.5 | 50.3 | 1.37 ± 0.04 |
| `./numbat-1.12 -e '1+2'` | 34.2 ± 0.9 | 32.2 | 36.7 | 1.00 |

sharkdp:
Thanks to various changes by @rben01, there is another 5% improvement in startup time. For more details, see https://github.com/sharkdp/numbat/pull/579#issuecomment-2389482040.

Also, note that some of the measurements above might be *underestimating* the true performance benefit, since Numbats prelude is growing. For a true comparison, we need to run both versions against the same version of the  prelude (typically the older one, as the new one might not be supported on the old interpreter).

sharkdp:
[...]

PR #579, opened 2024-09-26
More `String`s replaced with `&str`s

rben01:
A decent heuristic is that anything only used during parsing and not stored long term in the Environment can borrow from the input.


numbat/src/prefix_transformer.rs
```diff
@@ -280,13 +280,15 @@ impl Transformer {
...
-            .into_iter()
-            .map(|statement| self.transform_statement(statement))
-            .collect()
+        statements: impl IntoIterator<Item = Statement<'a>>,
+    ) -> Result<Vec<Statement<'a>>> {
+        let statements = statements.into_iter();
+        let mut ans = Vec::with_capacity(statements.size_hint().1.unwrap_or(0));
+        for s in statements {
+            ans.push(self.transform_statement(s)?);
+        }
```
sharkdp, line 291: Is this a performance improvement or was it somehow necessary to rewrite this iterator expression into a loop for another reason?

numbat/src/tokenizer.rs
```diff
@@ -144,7 +144,7 @@ pub enum TokenKind {
     Eof,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
sharkdp, line 147: Why this change? Even with your recent improvements, Token is still > 24 byte large(?). Maybe it's better to keep clones of this explicit, by not deriving Copy?

sharkdp (commented):
Thank you very much for this! I'm currently away from home but I plan to do some benchmarks soon to see if your changes result in performance improvements. They definitely result in memory improvements and save a ton of allocations which certainly also improves performance, especially on platforms with slow allocators (wasm?).

numbat/src/prefix_transformer.rs
rben01: Probably a vestige of me trying to fix a borrow checker error. Reverted.

numbat/src/tokenizer.rs
rben01: Probably a vestige of me trying to fix a borrow checker error. Reverted.

sharkdp (approved):
Thank you for the updates!

rben01:
> I plan to do some benchmarks soon to see if your changes result in performance improvements

According to hyperfine, `master d9e0e2c3e4ab6347ed34b3ab9bce4468b75abe75` is about 1.43 faster at evaluating the prelude than tag 1.13.0 (installed with homebrew). I do not know the various contributions of all the commits that have been made since then but I imagine reducing allocations is a big contributor.

```
Benchmark 1: numbat -e ""
  Time (mean ± σ):      58.4 ms ±   0.2 ms    [User: 47.5 ms, System: 10.7 ms]
  Range (min … max):    58.0 ms …  58.9 ms    49 runs
 
Benchmark 2: ./target/release/numbat -e ""
  Time (mean ± σ):      40.8 ms ±   0.2 ms    [User: 30.3 ms, System: 10.2 ms]
  Range (min … max):    40.4 ms …  41.3 ms    70 runs
 
Summary
  ./target/release/numbat -e "" ran
    1.43 ± 0.01 times faster than numbat -e ""
```

sharkdp:
> According to hyperfine, `master d9e0e2c3e4ab6347ed34b3ab9bce4468b75abe75` is about 1.43 faster at evaluating the prelude than tag 1.13.0 (installed with homebrew).

Yes. This is mostly due to the performance improvements documented in https://github.com/sharkdp/numbat/issues/525.



> I do not know the various contributions of all the commits that have been made since then but I imagine reducing allocations is a big contributor.

I don't think so. I did a few experiments, and it seems like your allocation-reducing PRs had no measurable runtime effect on x86. This does not mean that they were useless. As I said before, I *could* imagine that they help on other platforms (WASM). But I'm not sure how to measure performance in the browser. And your changes definitely help with memory consumption, which is an interesting metric in itself. Especially on non-desktop devices. Or when Numbat is used as a library in other applications.

sharkdp:
Ok, I'll take this back. There is a small performance improvement. I just checked out 29a99e9e235e3371777fc6eaa0a69c8f9786443b from Aug 29, before any of your changes:

```
commit 29a99e9e235e3371777fc6eaa0a69c8f9786443b
Author: David Peter <mail@david-peter.de>
Date:   Thu Aug 29 20:38:10 2024 +0200

    Add regression test for #534
```

and compared it with `master` (@ d9e0e2c3e4ab6347ed34b3ab9bce4468b75abe75).

The results are here:

```
Benchmark 1: ./numbat-master -e ""
  Time (mean ± σ):      65.0 ms ±   1.2 ms    [User: 55.5 ms, System: 8.8 ms]
  Range (min … max):    62.4 ms …  75.6 ms    300 runs
 
Benchmark 2: ./numbat-29a99e9 -e ""
  Time (mean ± σ):      68.1 ms ±   3.7 ms    [User: 58.5 ms, System: 9.0 ms]
  Range (min … max):    65.1 ms … 111.7 ms    300 runs
 
Summary
  ./numbat-master -e "" ran
    1.05 ± 0.06 times faster than ./numbat-29a99e9 -e ""

```

image

There seems to be a statistically-significant 5% performance improvement :+1: 

I ran both versions with the prelude from 29a99e9e235e3371777fc6eaa0a69c8f9786443b for a fair comparison.

sharkdp:
> I ran both versions with the prelude from [29a99e9](https://github.com/sharkdp/numbat/commit/29a99e9e235e3371777fc6eaa0a69c8f9786443b) for a fair comparison.

I think this is the key thing that I did wrong previously. If I measure `numbat-master` against the prelude on master (instead of the prelude @ 29a99e9e235e3371777fc6eaa0a69c8f9786443b), it takes 69 ms, i.e. it seemed to be slightly *slower* than the old Numbat version. But that comparison is not fair, as the prelude on master is different, and slightly larger:

```
▶ git checkout 29a99e9
▶ fd -e nbt . numbat/modules/ -X wc -l
…
2508
```

```
▶ git checkout master
▶ fd -e nbt . numbat/modules/ -X wc -l
…
2688
```

Some of this increase is from a new `extra::color` module with 44 lines which is not included by default. So the increase in lines is (2688 - 44) / 2508 = 1.054, i.e. about 5%. That seems to explain it perfectly.

PR #585, opened 2024-09-28
Remove allocations in `Markup` by switching to `Cow<'static, str>`

rben01:
Now strings are only allocated if non-static. (The vast majority of markup strings are `&'static`.)

There is of course a cost associated with this, the cost of checking the `Cow` variant, but given the number of static strings in use for markup, there's no way that this is a greater cost than allocating those as strings.


sharkdp (approved):
Thank you very much!

PR #590, opened 2024-09-29
Copy to clipboard (not complete as of submission)

rben01:
This is preliminary work on the copy-to-clipboard functionality that was mentioned in #394. Currently the `copy` command does not actually copy, but merely prints what would be copied (once I add that). In general I try to fall back to Numbat’s own defaults with user configuration where it makes sense. Here is the default config for this:

```toml
[copy-output-config.int-config]
grouping = "standard"
separator = ","
minus-sign = "-"

[copy-output-config.float-config]
max-sig-digits = 0
min-sig-digits = 0
max-width = 0
decimal = "."
capitalize-e = false

[copy-output-config.unit-config]
fancy-exponents = false
multiplication-operator = "·"
division-operator = "/"
space-btwn-operators = false
```

Please bike-shed away on these configuration options — names, default values, anything extraneous or missing...

With the above config (generated by `numbat --generate-config`), you'd get this kind of output:

```text
>>> copy  # first command in the session
error: no value to copy
>>> 1e12

  1.0e+12

    = 1.0e+12

>>> copy
1,000,000,000,000
>>> 1kg m^2/s^2

  1 kilogram × metre² / second²

    = 1 kg·m²/s²    [Energy or Torque]

>>> copy
1 kg·m^2/s^2
>>> 
```

Once I do a little more work investigating what's a good clipboard crate these days I can finish this up.

### DIFF
--- a/numbat-cli/src/config.rs
+++ b/numbat-cli/src/config.rs
@@ -1,3 +1,4 @@
+use crate::copy_to_clipboard::NumericDisplayConfig;
 use clap::ValueEnum;
 use serde::{Deserialize, Serialize};
 
@@ -65,6 +66,8 @@ pub struct Config {
     #[serde(skip_serializing)]
     pub load_user_init: bool,
     pub exchange_rates: ExchangeRateConfig,
+
+    pub copy_output_config: NumericDisplayConfig,
 }
 
 impl Default for Config {
@@ -78,6 +81,7 @@ impl Default for Config {
             load_user_init: true,
             exchange_rates: Default::default(),
             enter_repl: true,
+            copy_output_config: Default::default(),
         }
     }
 }
--- a/numbat-cli/src/copy_to_clipboard.rs
+++ b/numbat-cli/src/copy_to_clipboard.rs
@@ -1,0 +1,153 @@
+use numbat::{
+    markup::Markup, num_format::CustomFormat, num_format::Grouping, pretty_dtoa::FmtFloatConfig,
+    pretty_print::PrettyPrint, value::Value, FloatDisplayConfigSource, RuntimeError,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Default, Debug, Clone)]
+#[serde(remote = "Grouping", rename_all = "kebab-case")]
+pub enum IntGroupingConfig {
+    #[default]
+    Standard,
+    Indian,
+    #[serde(rename = "none")]
+    Posix,
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone)]
+#[serde(rename_all = "kebab-case", default, deny_unknown_fields)]
+pub struct IntDisplayConfig {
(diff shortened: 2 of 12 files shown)


rben01:
This works only for the CLI frontend. I need to think about how to enable similar functionality for other frontends, ideally without reimplementing it from scratch — hopefully the only thing to be reimplemented would be the clipboard copying itself. Related to #584

sharkdp:
Thank you very much. This does not seem to work for me on Linux (using a standard X server with i3). After typing `copy`, I get the expected message:
```
>>> 2+3

  2 + 3

    = 5

>>> copy
5 was copied to the clipboard
```

but the `5` is nowhere to be found, neither in the "ctrl+v" clipboard (X11: clipboard / secondary), nor in the "middleclick" clipboard (X11: primary).

rben01:
Hmm, how about now? (I don't have a linux box to test with.) Seems like linux does some weird stuff where dropping the clipboard clears its contents. Also it looks like when numbat-cli exits, the clipboard’s contents will be cleared for the same reason. Is it worth working around this? The arboard docs show how to do this, but it involves spawning a background process to host the clipboard, which seems like a lot of complexity for a small gain.

rben01:
Oh, so `wait` literally waits. I'm guessing then that using the non-Linux code for Linux should work, at least as long as numbat is running?

sharkdp:
> I'm guessing then that using the non-Linux code for Linux should work, at least as long as numbat is running?

You mean the state you had before? ([c7b4f89](https://github.com/sharkdp/numbat/pull/590/commits/c7b4f89c0f62d2da41f74db597b6dce2791ecbfc))

No. It doesn't work. Even if numbat is still open.

> Also it looks like when numbat-cli exits, the clipboard’s contents will be cleared for the same reason. Is it worth working around this?

I don't think so. A couple of applications have a similar behavior. It's slightly annoying, but I don't think it's a big problem for Numbat.

rben01:
How about now?

sharkdp:
Now it works — thank you!

I will do a full review shortly.

I will already note that I am a bit overwhelmed by the number of configuration options. I'm usually trying to keep the number of options very small, unless it's definitely necessary to add them. I'm not (yet) very strict about being backwards compatible, but once we are, we need to maintain all of those options.

rben01:
> I will already note that I am a bit overwhelmed by the number of configuration options.

That's fair. I naively added (nearly) as much as was supported, with almost no editing. `max-sig-digits`, `min-sig-digits`, and `max-width` can probably all go. The rest I feel are important to keep, though.

sharkdp:
> `max-sig-digits`, `min-sig-digits`, and `max-width` can probably all go. The rest I feel are important to keep, though.

If it's not too much trouble, it would be great if we could remove them for now. Thank you.

rben01:
Actually very little trouble! I also renamed `copy_output_config` (`[copy-output-config]` in TOML) to `copy_result`, as the docs refer to the outcome of a calculation as the result, and “config” seemed redundant in that case. (I left `int-config`, `float-config`, `unit-config` as is.)

PR #591, opened 2024-09-30
Cache `help` output and preinitialize it for speedier first access

rben01:
It bothered me that 1. `help` took a noticeable amount of time to print (due to spinning up a whole context and evaluating in it) and 2. it did this every time it was invoked. Since the `help` markup doesn't change, we might as well cache it. And since it takes time to generate, we might as well precompute it in a separate thread (in the CLI crate) before the user even has a chance to invoke it. This makes the `help` output appear immediately each time (assuming you invoke it for the first time at least a few tenths of a second after numbat starts up) without delaying program startup.

We could in theory even cache the call to `ansi_format(help, true)`, but this would require the parent `numbat` crate to hold the `ANSIFormatter` machinery (since `help_markup()` lives in crate `numbat`), which is probably not worth it.

### DIFF
--- a/numbat-cli/src/main.rs
+++ b/numbat-cli/src/main.rs
@@ -309,6 +309,8 @@ impl Cli {
         );
         rl.load_history(&history_path).ok();
 
+        std::thread::spawn(help_markup);
+
         if interactive {
             match self.config.intro_banner {
                 IntroBanner::Long => {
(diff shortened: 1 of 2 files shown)


sharkdp:
Thank you. I'm not sure this is the best way forward.

Spawning a separate thread seems like a big hammer to solve this problem. I was thinking that we could maybe just hard-code the results (and then add a test to make sure they are still consistent). But now that I think about it more, I'm not even sure that I like those examples in the `help` text. And even if we keep some examples in the help text, do we really need to include the results? I'd like to think about those questions first before proceeding with this.

rben01:
I think the examples are fine, could maybe use some fleshing out (e.g., unit and function definitions, lists and string). I think the outputs of the inputs are helpful because they reduce surprises and remove some of the onus of discovery from the user.

That said I don't think they necessarily have to be computed live. They could be precomputed and kept in sync with tests, or generated with `build.rs`. (`build.rs` could also have other uses, such as keeping the list of currencies in sync between Rust and the numbat currencies file.)

sharkdp:
Closing this for now

PR #594, opened 2024-10-01
Improved completion performance

rben01:
Removed non-matching candidates from consideration earlier, obviating need to filter them out
Removed clones of candidates that didn't match (before: clone then check; after: check then clone)

Also removed pointless unnecessary work in `print_environment` (code that was replaced but then not deleted; kinda surprised clippy doesn't warn about pointless `sort` on a vector that'll never actually be used).


sharkdp (approved):
Nice — thank you!

PR #598, opened 2024-10-01
Tokenizer improvements

rben01:
Shrunk `TokenKind` from 16 bytes to 2 by replacing `IntegerWithBase(usize)` with `IntegerWithBase(u8)` (max base is 16, no way do we exceed 255)

Removed Box from base-n digit predicate function by switching to function pointers


sharkdp:
Thank you very much!

PR #599, opened 2024-10-03
Improved error message when encountering a trailing equals

rben01:
Added heuristic to check whether user might have been trying to define a function

Before:

```
1 │ f(x,y) = x+y
  │        ^ Trailing '=' sign. Use `let ) = …` if you intended to define a new constant.
```

After:

```
1 │ f(x,y) = x+y
  │        ^ Trailing '=' sign. Use `fn f(x,y) = …` if you intended to define a function.
```

When defining a variable, the old message is left basically the same except that all identifiers left of the equals sign, not just the last, are part of the error message.


sharkdp (approved):
Very nice — thank you!

PR #600, opened 2024-10-05
Added numbat-wasm back to workspace

rben01:
Fixed resulting clippy errors

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -349,6 +349,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "console_error_panic_hook"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


sharkdp:
Thank you. We now get this cargo warning when running `cargo build` / `cargo run`:

```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /home/shark/numbat/numbat-wasm/Cargo.toml
workspace: /home/shark/numbat/Cargo.toml
```

rben01:
Hmm, not sure how to fix this. https://github.com/rust-lang/cargo/issues/8264

Mucking with the package layout just to fix this one warning doesn't seem worth it... I wonder if there's a workaround.

sharkdp:
I'd like to close this, as I find the warning confusing. If we don't find a way to silence it, it might be confusing to (first time) contributors as well.

PR #601, opened 2024-10-06
Cleaned up some todos and clippy warnings

rben01:


numbat/src/typed_ast.rs
```diff
@@ -41,10 +41,22 @@ type DtypeFactorPower = (DTypeFactor, Exponent);
...
+        Self { factors }
+    }
+
+    pub fn factors(&self) -> &[DtypeFactorPower] {
+        &self.factors
+    }
+
+    pub fn factors_mut(&mut self) -> &mut [DtypeFactorPower] {
+        &mut self.factors
+    }
```
sharkdp, line 58: I think the idea of the TODO comment was not to get rid of the `pub`, but rather to prevent mutable access to the `factors` member, in order to make sure that invariants of `DType` are always fulfilled (i.e. that `factors` are in canonical form).

rben01: Gotcha. I just changed a few things around so that a `DType` can be consumed to provide its `factors`, and then can only be created again via `from_factors`, ensuring it's always canonicalized.

sharkdp (approved):
Thank you!

PR #604, opened 2024-10-08
Add explicit use statement for non-prelude functions to examples

### LINKED ISSUE #602: Cannot find quadratic_equation on numbat.dev
image

Bzero:
Resolves #602.

### DIFF
--- a/book/src/list-functions-math.md
+++ b/book/src/list-functions-math.md
@@ -745,14 +745,16 @@ fn diff<X: Dim, Y: Dim>(f: Fn[(X) -> Y], x: X) -> Y / X
 <summary>Examples</summary>
 
 Compute the derivative of \\( f(x) = x² -x -1 \\) at \\( x=1 \\).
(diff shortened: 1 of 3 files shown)


sharkdp (approved):
Thank you!

sharkdp:
Should we maybe add an additional `>>>` here?

image

Bzero:
We can do that. Would you do it for all multi-line examples or only here with the `use` statements?

Goju-Ryu:
I’m thinking all of them to make it clear what is input and what is output.

sharkdp:
Just be aware that there might be multi-line *expressions* which should not be formatted this way. So something like
```
fn f(x) =
  x² + 1
```
should *not* be rendered as
```
>>> fn f(x) =
>>>  x² + 1
```
because it needs to be entered as a single line.

The easiest option might be to get rid of the `>>>` prompts all-together?

sharkdp:
> I think the `>>> ` prompts help to distinguish in and output in the examples, especially for the many one-liners we have so I would prefer to keep them.

But we also have an indented equals sign for the outputs. So we would go from
```
>>> floor_in(m, 5.7 m)

    = 5 m    [Length]
```
to
```
floor_in(m, 5.7 m)

    = 5 m    [Length]
```

I don't really see a problem with that. I think I even prefer the latter one. Some users might not immediately understand that `>>>` is supposed to be a prompt.

Bzero:
Yes that's true as well, I think it's fine to remove it then.

PR #607, opened 2024-10-09
Reduced allocations relating to `ast::Expression` + `typed_ast::Expression` and reduced allocations in `Transformer::transform_*` functions

rben01:


rben01:
A single box should result in both fewer requests of the allocator and better cache locality when accessing that data, although I can't imagine it'd be a particularly large difference with the data we're working with here. Especially since who knows what optimizations the compiler is making — maybe if it sees three boxes side by side, it just places them next to each other in the heap anyway. I can revert that portion if you'd like because it probably makes very little difference.

sharkdp:
> I can revert that portion if you'd like because it probably makes very little difference.

That would be great.

sharkdp:
Thank you! There's one more conflict that needs to be fixed.

sharkdp (approved):
Thank you!

PR #609, opened 2024-10-09
Add percentage-calculation functions

### LINKED ISSUE #494: 1000 - 20%
Can percentage expressions that work in calculators like Raycast and Numi, such as `1000 - 20%` work in Numbat?

...That would be great. 😃 This is a really common use case.

sharkdp:
```
>>> 72 € |> increase_by(15%)

    = 82.8 €    [Money]

>>> percentage_change(3.2 kg, 4.7 kg)

    = 46.875 %
```

Additional ideas / TODOs:

- [ ] Special syntax for `increase_by`/`decrease_by`?
- [ ] Special/custom error message when attempting to do something like `72 € - 15%`.

closes #494

### DIFF
--- a/book/build.py
+++ b/book/build.py
@@ -177,6 +177,10 @@ list_of_functions(
                 ],
             },
             {
+                "title": "Percentage calculations",
+                "modules": ["math::percentage_calculations"],
+            },
+            {
                 "title": "Geometry",
                 "modules": ["math::geometry"],
             },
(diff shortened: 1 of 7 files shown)


Goju-Ryu:
This looks great, but the name of `percentage_chance` is a bit confusing to me. I had to reread some examples to figure out what the intent was. 

Maybe something like `percentage_difference` would be more clear?

So the current:
```fs
assert_eq(percentage_chance(40 dollar, 50 dollar), +25%)
```
Would be this instead:

```fs
assert_eq(percentage_difference(50 dollar, 40 dollar), +25%)
# or with pipeline 
assert_eq((40 dollar |> percentage_difference(50 dollar)), +25%)
```

I’m not sure if it is exactly right, but I think naming it something without the word chance would make it more easily understood.

sharkdp:
> `percentage_chance`

> without the word chance

Is this a typo or did you misread? The function is called "percentage_chan*g*e".

Goju-Ryu:
> Is this a typo or did you misread? The function is called "percentage_chan*g*e".

It wasn't a typo, I see now that I misread it multiple times over. It makes a lot more sense with the naming now that I actually read it right.

sharkdp:
That'd be great — thanks.

PR #629, opened 2024-10-19
Vscode extension quality of life changes

Goju-Ryu:
This PR adds and modifies a few behaviors of the vs-code extension to improve the developer experience. Some changes may be controversial so I will list them here to open a discussion about if some of them should be reverted before merge.

- square brackets are now treated as brackets, allowing for colorization by vs code
- square brackets and angle brackets now auto close the same way normal parenthesis do
- The folding is changed to be based on code indentation[^1] (I expect this to be the most controversial)
- Strings with interpolated portions will now highlight the interpolation as with normal code, while the rest of the string gets the usual color of strings.
- Correctly display strings with escaped quotes (`\"`) inside strings

**Examples:**
Before fold:
image
After fold:
image
Examples of interpolations and escaped brackets:
image
Example with escaped quotation mark:
image

[^1]: This isn't strictly correct, but aligns with the code style used and in my opinion gives more intuitive folding in the prelude files


sharkdp:
Awesome — thank you. I'm surprised someone found the VScode extension :-)

Goju-Ryu:
> Awesome — thank you. I'm surprised someone found the VScode extension :-)

I use it heavily when writing code for the prelude. I just decided to read up on how it worked so I could fix the quirks that annoyed me in my own workflow. 

This also reminds me, is this extension available for install without running it in debug mode via vs code? I seem to remember reading about a reason it wasn't in the extensions catalog, but when i went looking for where I read it I couldn't find it.

sharkdp:
> This also reminds me, is this extension available for install without running it in debug mode via vs code?

No.



> seem to remember reading about a reason it wasn't in the extensions catalog, but when i went looking for where I read it I couldn't find it.

There's no reason except that I never published a VS code extension and I'm afraid it might involve some boring organizational things :smile:

sharkdp:
That would be great — thanks!

Goju-Ryu:
No problem. Back when I first discovered numbat I was slightly disappointed that I couldn't find where to get the vs code extension, so I’m excited to hopefully make it easier for newcomers in the future.

PR #634, opened 2024-10-24
Add android to cli-installation.md

archisman-panigrahi:
Installing [insect in Android](https://github.com/sharkdp/insect?tab=readme-ov-file#terminal-version) was a little easier because of npm, but it is still possible to install numbat in Android.


sharkdp:
Thank you.

I'd rather not add complicated install instructions like this into the documentation. Feel free to open a Discussion for this though: https://github.com/sharkdp/numbat/discussions

PR #638, opened 2024-10-26
Fail documentation generation for invalid examples & remove prompts in examples

Bzero:
This is a follow up on #566 and #604 regarding changes which were agreed upon but which I never got to implement (so far):

- Hard fail  the generation of the documentation if there are invalid examples (see https://github.com/sharkdp/numbat/pull/566#discussion_r1779808656)
- Remove the prompts in the example code (see https://github.com/sharkdp/numbat/pull/604#issuecomment-2405888212)

### DIFF
--- a/book/build.py
+++ b/book/build.py
@@ -2,6 +2,7 @@ import subprocess
 from pathlib import Path
 import urllib.parse
 import os
+import sys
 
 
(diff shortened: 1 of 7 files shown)


sharkdp (approved):
Thank you!

PR #639, opened 2024-10-27
Made `proper_function_call` static and moved `get_proper_function_reference` to `Environment` to prove to the compiler that mutable references don't overlap

rben01:
Alternative implementation of #606


sharkdp (approved):
Thank you!

Issue #512, opened 2024-07-28
Markdown in `@description()` decorators


Bzero:
#469 added the option to use (some) markdown in `@description()` decorators which I think is very neat for the online documentation and helps to make it more readable, however it also makes the description a bit inconvenient to read in the cli using the `info` command.

Possibly we could directly use Unicode characters instead of LaTeX commands in formulas (e.g. `$Γ(x)$` instead of `$\Gamma(x)$`) and use colors to highlight code and math blocks to make descriptions more easily readable in the cli.

sharkdp:
I only thought about improving the online documentation but forgot about `info` for a second there, sorry about that.

I know that there are some markdown viewers for the terminal. Maybe the easiest way forward would be to use one of these (as a library)? For example, they might be able to show markdown links using ANSI hyperlinks? The problem is that they would need to support Latex formulas as well, but maybe there are some?



> Possibly we could directly use Unicode characters instead of LaTeX commands in formulas (e.g. `$Γ(x)$` instead of `$\Gamma(x)$`)

If nothing else, we should definitely do that... if it will still be recognized by mdbook and its mathjax support.

> and use colors to highlight code and math blocks to make descriptions more easily readable in the cli

If we don't get this from an external library, we could certainly attempt to do that. But I'm yet not sure if I want a custom markdown parser just for that(?).

Bzero:
> Maybe the easiest way forward would be to use one of these (as a library)?

Yes, if there is a library that can display markdown including math blocks in the terminal this would certainly be the easiest and probably also best way. I had a brief look around but could unfortunately not find a terminal markdown viewer library that can display LaTeX formulas in a satisfactory way.

> But I'm yet not sure if I want a custom markdown parser just for that(?).

No, I wouldn't suggest that either, however I think math and code blocks are the most important markdown features for the use case in numbat and detecting those should not be too hard.

sharkdp:
Another option would be to have two fields (@doc and @info or similar). But the duplication would be a real shame. On the other hand, we would only have to do that for the functions with formulas in the descriptions. For the others, one of them could fall back to the other.

The LaTeX we're using so far is extremely basic. Maybe we can indeed write a small parser ourselves? And turns things like `\sqrt{…}` into `sqrt`; `\frac{…}{…}` => `(…)/(…)`. And for symbols like `\pi` and `\Gamma`, we could check if it's possible to use unicode characters.

PR #645, opened 2024-11-01
Math functions

rben01:
I made some choices of function names and functionality (e.g., `binom` taking non integer args). Let me know what you think.


numbat/modules/math/combinatorics.nbt
```diff
@@ -0,0 +1,34 @@
+use core::error
+use core::functions
+use math::transcendental
+
+@name("Factorial")
+@description("The product of the integers 1 through n, also written n!")
```
sharkdp, line 6: You can use `$…$` to embed LaTeX formulas in `@description`s. This way, they render more nicely in the online documentation.

numbat/modules/math/combinatorics.nbt
```diff
@@ -0,0 +1,34 @@
...
+@url("https://en.wikipedia.org/wiki/Factorial")
+fn factorial(n: Scalar) -> Scalar = n!
+
+@name("Falling factorial")
+@description("Equal to n⋅(n-1)⋅…⋅(n-k+2)⋅(n-k+1) (k terms total). If n is an integer, this is the number
+	of k-element permutations from a set of size n. k must always be an integer.")
```
sharkdp, line 12: Does this whitespace lead to problems in the rendered output? I think you can have multiple `@description` fields and they will be concatenated.

numbat/modules/math/combinatorics.nbt
```diff
@@ -0,0 +1,34 @@
...
+@url("https://en.wikipedia.org/wiki/Falling_and_rising_factorials")
+fn falling_factorial(n: Scalar, k: Scalar) -> Scalar =
+	if k < 0 || fract(k) != 0 then
```
sharkdp, line 15: We have `is_non_negative`. Maybe we should also add `is_integer` and use both functions here (and below)?

numbat/modules/math/combinatorics.nbt
sharkdp: Maybe this was your intent, but I think we could state more clearly that Numbat also supports `!` as an operator for factorials.

sharkdp (changes requested):
This looks great. Thank you very much. Just a few minor comments.

numbat/modules/math/combinatorics.nbt
rben01: Multiple `@description` fields are concatenated with newlines between them, which I'm not sure I want. But then again the literal newline is picked up in the string, so there may not be a way to get this stored as a single line while writing it across multiple lines.

rben01: Separately, this has me wondering whether there would be merit in having separate `@description`s for docs and for numbat CLI. I would like to write that product as `$n\cdot(n-1)\cdot\ldots\cdot(n-k+2)\cdot(n-k+1)$` for the docs, but I wouldn't want to have to read that in the CLI. (Is there a LaTeX-to-unicode compiler?)

Bzero: We discussed this partially in #512 but didn't come to a satisfactory conclusion so far. I think the best to do as of now would be to directly use the Unicode characters that would be produced during compilation in math blocks instead of their LaTeX commands (e.g `⋅` instead of `\cdot`, `…` for `\ldots` and so on). MathJax handles this well as far as I can tell and it is also decently readable as plain text.

numbat/modules/core/numbers.nbt
```diff
@@ -23,4 +23,9 @@ fn is_zero<D: Dim>(value: D) -> Bool = value == 0
...
 @example("is_nonzero(37)")
 @example("is_nonzero(0)")
-fn is_nonzero<D: Dim>(value: D) -> Bool = !is_zero(value)
\ No newline at end of file
+fn is_nonzero<D: Dim>(value: D) -> Bool = !is_zero(value)
+
+@description("Returns true if the input is an integer. If the input has units, then returns whether it is an integral number of its units.")
+@example("is_int(30 seconds)")
+@example("is_int(0.5 minutes)")
+fn is_int<D: Dim>(value: D) -> Bool
```
sharkdp, line 31: I think this should be a function on `Scalar`. If we make it work on dimensionful quantities, it introduces inconsistencies, as you demonstrated in the example:
```
true = is_int(30 seconds) ?=? is_int(0.5 minutes) = false
```
We should avoid this if possible (I know that there are already ways to introduce inconsistencies like this via `value_of`; but those functions are at least marked "unsafe").

Minor: I'd like to rename it to `is_integer`

Also, can we implement it in Numbat, using `fract`?
```
fn is_integer(value: Scalar) -> Bool = is_zero(fract(value))
```

rben01: Can do. What is the benefit of implementing functions in Numbat over Rust?

sharkdp: That is a good question.

I think of the "FFI bridge" between Rust and Numbat as something that helps Numbat to get up on its feet. We will always need support from the "runtime". We will probably never implement our own PRNG to implement `rand` in Numbat. We will almost certainly never have a full networking stack to be able to look up currency exchange rates from within Numbat. And we will never implement `print` in Numbat.

But as Numbat grows into a more mature language of its own, I think it makes sense to keep the foreign-function interface as small as possible. Especially in areas which belong to the core functionality of Numbat like mathematical computations.

Having part of the "standard library" written in Numbat itself helps us to design APIs in a way that makes them easy to use … in Numbat itself. #646 is a good example. If all string functions were written in Rust and re-exported to Numbat, may we wouldn't have found out that they are not as ergonomic as they could be.

The standard library is the biggest Numbat "code base" that we have, and it's a good "benchmark" to have when we make changes to the language.

There are also practical advantages to having functions implemented in Numbat itself. Even if we reimplement something that would be readily available in Rust, it's typically way less code to maintain. The function name appears just once in the `fn …` definition instead of multiple times for a FFI function. It also allows us to change/develop functions without having to recompile the Numbat interpreter.

There's also downsides of course. Performance, for example. If we see that a heavily-used function is too slow, we can think about re-implementing it in Rust.

Side remark: Note that there are also some functions like `map` that can't even be implemented in Rust at the moment, as we don't have a way to call other Numbat functions from FFI functions.

sharkdp (approved):
Thank you!

PR #646, opened 2024-11-02
String api update

### LINKED ISSUE #628: String functions work poorly with postfix apply
I'm currently developing functions that make heavy use of strings and the `core::strings` module. In this process I have been unable to use the postfix apply operator almost at all due to argument order. I would love if either the order got changed to make the string worked on the last argument or to make a set of postfix apply friendly versions.
I believe it would make string heavy code more easily readable and easier to refactor. As an example here is the current version of `str_replace` and a version using pipes on the assumption that the first argument of all string functions were moved to the end of the argument list:

```py
# Current implementation
fn str_replace(s: String, pattern: String, replacement: String) -> String =
  if pattern == ""
    then s
    else if str_contains(s, pattern)
           then if str_slice(s, 0, str_length(pattern)) == pattern
               then str_append(replacement, str_replace(str_slice(s, str_length(pattern), str_length(s)), pattern, replacement))
               else str_append(str_slice(s, 0, 1), str_replace(str_slice(s, 1, str_length(s)), pattern, replacement))
           else s

# same implementation but with arguments moved and using pipes (also used a hypothetical prepend function instead of append)
fn str_replace(pattern: String, replacement: String, s: String) -> String =
  if pattern == ""
    then s
    else if str_contains(s, pattern)
           then if str_slice(s, 0, str_length(pattern)) == pattern
               then (s |> str_slice(str_length(pattern), str_length(s)) |> str_replace(pattern, replacement) |> str_prepend(replacement))
               else (s |> str_slice(                  1, str_length(s)) |> str_replace(pattern, replacement) |> str_prepend(str_slice(s, 0, 1)))
           else s
```

Goju-Ryu:
This PR closes #628 by reordering arguments of all string functions. 
With this PR all string functions have the string being manipulated as the last argument.

### DIFF
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -182,9 +182,9 @@ fn join(xs: List<String>, sep: String) =
 fn split(input: String, separator: String) -> List<String> =
   if input == ""
     then []
-    else if !str_contains(input, separator)
+    else if !str_contains(separator, input)
       then [input]
-      else cons(str_slice(input, 0, idx_separator),
-                split(str_slice(input, idx_separator + str_length(separator), str_length(input)), separator))
+      else cons(str_slice(0, idx_separator, input),
+                split(str_slice(idx_separator + str_length(separator), str_length(input), input), separator))
   where
-    idx_separator = str_find(input, separator)
+    idx_separator = str_find(separator, input)
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -87,7 +87,7 @@ fn filter<A>(p: Fn[(A) -> Bool], xs: List<A>) -> List<A> =
       else filter(p, tail(xs))
 
 @description("Fold a function over a list")
-@example("foldl(str_append, \"\", [\"Num\", \"bat\", \"!\"])", "Join a list of strings by folding.")
+@example("foldl(str_prepend, \"\", [\"Num\", \"bat\", \"!\"])", "Join a list of strings by folding.")
 fn foldl<A, B>(f: Fn[(A, B) -> A], acc: A, xs: List<B>) -> A =
   if is_empty(xs)
     then acc
--- a/numbat/modules/core/lists.nbt
+++ b/numbat/modules/core/lists.nbt
@@ -87,7 +87,7 @@ fn filter<A>(p: Fn[(A) -> Bool], xs: List<A>) -> List<A> =
       else filter(p, tail(xs))
 
 @description("Fold a function over a list")
-@example("foldl(str_prepend, \"\", [\"Num\", \"bat\", \"!\"])", "Join a list of strings by folding.")
+@example("foldl(str_append, \"\", [\"Num\", \"bat\", \"!\"])", "Join a list of strings by folding.")
 fn foldl<A, B>(f: Fn[(A, B) -> A], acc: A, xs: List<B>) -> A =
   if is_empty(xs)
     then acc
--- a/numbat/modules/core/strings.nbt
+++ b/numbat/modules/core/strings.nbt
@@ -28,11 +28,11 @@ fn uppercase(s: String) -> String
 
 @description("Concatenate two strings")
 @example("\"Numbat\" |> str_append(\"!\")")
-fn str_append(a: String, str: String) -> String = "{str}{a}"
+fn str_append(a: String, b: String) -> String = "{a}{b}"
 
 @description("Concatenate two strings")
 @example("\"!\" |> str_prepend(\"Numbat\")")
-fn str_prepend(a: String, str: String) -> String = "{a}{str}"
+fn str_prepend(a: String, b: String) -> String = "{b}{a}"
 
 @description("Find the first occurrence of a substring in a string")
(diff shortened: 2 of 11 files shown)


numbat/modules/core/strings.nbt
```diff
@@ -27,43 +27,49 @@ fn lowercase(s: String) -> String
 fn uppercase(s: String) -> String
 
 @description("Concatenate two strings")
-@example("str_append(\"Numbat\", \"!\")")
-fn str_append(a: String, b: String) -> String = "{a}{b}"
+@example("str_append(\"!\", \"Numbat\")")
```
sharkdp, line 30: Maybe
```suggestion
@example("\"Numbat\" |> str_append(\"!\")")
```

numbat/modules/core/strings.nbt
```diff
@@ -27,43 +27,49 @@ fn lowercase(s: String) -> String
...
+fn str_append(a: String, str: String) -> String = "{str}{a}"
+
+@description("Concatenate two strings")
+@example("str_prepend(\"Numbat\", \"!\")")
```
sharkdp, line 34: Similar here?
```suggestion
@example("\"!\" |> str_prepend(\"Numbat\")")
```

numbat/modules/extra/color.nbt
```diff
@@ -37,7 +37,7 @@ fn color_rgb_float(color: Color) -> String =
 @description("Convert a color to its hexadecimal representation.")
 @example("rgb(225, 36, 143) -> color_hex")
 fn color_hex(color: Color) -> String =
-  str_append("#", str_replace(str_replace("{color -> _color_to_scalar -> hex:>8}", "0x", ""), " ", "0"))
+  str_prepend("#", str_replace(" ", "0", str_replace("0x", "", "{color -> _color_to_scalar -> hex:>8}")))
```
sharkdp, line 40: Do we maybe want to rewrite this using `|>` notation now?

numbat/modules/core/lists.nbt
```diff
@@ -87,7 +87,7 @@ fn filter<A>(p: Fn[(A) -> Bool], xs: List<A>) -> List<A> =
       else filter(p, tail(xs))
 
 @description("Fold a function over a list")
-@example("foldl(str_append, \"\", [\"Num\", \"bat\", \"!\"])", "Join a list of strings by folding.")
+@example("foldl(str_prepend, \"\", [\"Num\", \"bat\", \"!\"])", "Join a list of strings by folding.")
```
sharkdp, line 90: Well this change is… strange. The order reversal for `str_append` is consistent with the rest, sure. But it's also very much not the order I would expect. Can we maybe do some research how other functional programming languages handle this?

sharkdp:
Thank you!

Goju-Ryu:
Thank you for the review, I agree on all counts. I think the color function especially was a good call and got much easier to understand after the change.

numbat/modules/core/lists.nbt
Goju-Ryu: Sure, I will start to look into it.

Goju-Ryu: I had a difficult time finding languages with both an `|>` operator equivalent and the append/prepend functions. Below are my findings so far. It seems to me that there isn't a consensus yet, but it is slightly more normal to have the arguments the way they originally were as opposed to the ones proposed in this PR. 

The same as proposed:
- https://hackage.haskell.org/package/strings-1.1/docs/Data-Strings.html

The same as original:
- https://www.roc-lang.org/builtins/Str#concat (under the name of concat, but the idea is the same)
- https://package.elm-lang.org/packages/elm/core/latest/String#append (only has append)

No 1-to1 comparison:
- https://hexdocs.pm/elixir/String.html#replace_prefix/3 (close to the original but via a different function)
- https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-stringmodule.html (uses the String class method and not a functional approach)
- https://share.unison-lang.org/@unison/base/code/releases/3.23.0/latest/terms/Text/cons (uses `cons` and `snoc` instead)

I don't really mind either order for the arguments and only changed it in this PR to stay consistent. I think either piping or normal function call will always feel weird. Maybe by changing the naming we can make it less so, but I'm unsure how that would look like.

Goju-Ryu: Having thought about it a bit since, I think the best solution might be to reverse the arguments such that `str_append` functions as it did before. To use them nicely in pipelines would require remembering that quirk, but that would be a lot less complicated when lambdas are added someday.

I think the pipelines are important, and they are what I use most often, but thinking of how beginners would perceive the function has lead me to believe that the confusion my change to the function would cause isn't worth it. 

If you agree, I will change the signature of `str_append` and `str_prepend` and update the brach to make sure it won't break any code merged in the meantime.

sharkdp: > If you agree, I will change the signature of `str_append` and `str_prepend` and update the brach to make sure it won't break any code merged in the meantime.

That would be great, thanks.

sharkdp (approved):
Thank you!

PR #647, opened 2024-11-03
Fixed NaN comparison

### LINKED ISSUE #644: `NaN < 0` errors
I would expect that `NaN < 0` be `false`, but instead it gives

```
>>> NaN < 0.0
error: runtime error
 = Conversion error: unit '' can not be converted to ''
```

rben01:
All comparisons involving at least one NaN now return false, regardless of units

Closes #644

### DIFF
--- a/numbat/src/quantity.rs
+++ b/numbat/src/quantity.rs
@@ -332,8 +332,6 @@ impl PartialEq for Quantity {
     }
 }
 
-impl Eq for Quantity {}
-
 impl PartialOrd for Quantity {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
         let other_converted = other.convert_to(self.unit()).ok()?;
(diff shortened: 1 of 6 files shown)


rben01:
One question is whether NaN should be comparable with other quantities regardless of units. I think yes, it is like 0 in this regard: the answer is known to be false regardless of units.

numbat/src/quantity.rs
```diff
@@ -347,7 +345,36 @@ impl PrettyPrint for Quantity {
     }
 }
 
+pub(crate) enum QuantityOrdering {
+    IncompatibleUnits,
+    NanOperand,
+    Less,
+    Equal,
+    Greater,
+}
```
sharkdp, line 354: Would this lead to slightly less code?
```suggestion
pub(crate) enum QuantityOrderingResult {
    IncompatibleUnits,
    NanOperand,
    Ok(std::cmp::Ordering),
}
```

numbat/src/quantity.rs
```diff
@@ -347,7 +345,36 @@ impl PrettyPrint for Quantity {
...
+            return QuantityOrdering::IncompatibleUnits;
+        };
+
+        match self.value.partial_cmp(&other_converted.value) {
+            Some(cmp) => match cmp {
+                std::cmp::Ordering::Less => QuantityOrdering::Less,
+                std::cmp::Ordering::Equal => QuantityOrdering::Equal,
+                std::cmp::Ordering::Greater => QuantityOrdering::Greater,
+            },
+            None => unreachable!("unexpectedly got a None partial_cmp from non-NaN arguments"),
```
sharkdp, line 374: Maybe catch this with a `.expect(…)` above, using the same message?

sharkdp:
> One question is whether NaN should be comparable with other quantities regardless of units. I think yes, it is like 0 in this regard: the answer is known to be false regardless of units.

Yes, we want `NaN` to be polymorphic. If `0` is polymorphic, then so is `0 × inf == NaN`:
```
meter × NaN = meter × (0 × inf) = (meter × 0) × inf = 0 × inf = NaN
```
(ignoring associativity-violations in FP math for a second)

If `NaN` wouldn't be polymorphic, that equation above would not by typeable (`Length × Scalar = Scalar`?).

rben01:
Not sure how I accidentally closed this or why I ended up force pushing, or why the macOS job didn't complete, but this PR should be good now.

sharkdp (approved):
Thank you!

Issue #621, opened 2024-10-14
cryptic error when using `line_plot`


Jiehong:
Hello,

I use numbat on the cli, and I saw that some plot functions now exist. So I tried to run:

```
let A0 = 3 cm
let ω = 20 Hz
let λ = 1.5 s

fn amplitude(t) = A0 exp(-t/λ) cos(ω t)

line_plot(amplitude, 0 s, 8 s) |>
  xlabel("Time") |>
  ylabel("Amplitude") |>
  show
```

I expected the `show` to maybe fail, but maybe I could save it to svg or something.

However, it never reached that point at all, because the following line fails with an error:

```
>>> line_plot(amplitude, 0 s, 8 s)
error: runtime error
 = User error: …
```

The `...` are not something I added.

So I guess I am a user, and thus an error :D

So, perhaps functions not supported in the CLI show not be callable with an error saying `this function is only supported in the browser`, or (bigger fix I think), the plot function should work in the CLI, but `show` could be replaced by `save_svg` or something like that (`àla matplotlib: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html).

Goju-Ryu:
That is very likely caused unit_of being called on a value of zero. See #521 for more context.

Goju-Ryu:
I just checked the implementation and can confirm it is because your second argument is 0. Use a value slightly higher or lower and it should be resolved.

sharkdp:
Let's leave this open until this #521 mess is resolved :smile:. Users should not have to adapt their plot ranges because of this.

sharkdp:
Closed via e8bac1feeb02f983910af8a41993fedd3dfc4a23

Issue #521, opened 2024-08-08
unit_of is unsound


sharkdp:
The `unit_of` function
```rs
fn unit_of<T: Dim>(x: T) -> T
```
which returns e.g. `1 km/h` for an input of `30 km/h` can lead to soundness bugs when the argument is zero.

Consider this case, for example:
```rs
>>> fn normalize_length(x: Length) = unit_of(x)
>>> normalize_length(0) + 1 m
error: runtime error
 = Conversion error: unit 'm' can not be converted to ''
```
The type checker doesn't complain since the signature of `unit_of` promises to return a `Length` for this case. But `unit_of(0) == 1`, which leads to the subsequent conversion error.

It can even lead to crashes, like in this example:
```
fn diff(f, x) = f(x + unit_of(x)) / unit_of(x)
fn amp(t) = cos(1 Hz × t)

diff(amp, 0)
```

rben01:
Just to have an MRE, `cos(1Hz * unit_of(0))` is sufficient to crash numbat.

rben01:
Also it appears only to affect the literal 0, not something like 1-1. Neither `cos(unit_of((1-1)m)*Hz)` nor `cos(unit_of(1-1)*Hz)` type check, but neither crashes Numbat. Could be something to do with the constraint solver treating literal 0 differently?

If you could reach into a `unit_of` call and replace any literal 0s with 1-1, that would be an interesting hack.

rben01:
On second thought is the underlying issue that `unit_of` is unsound, or that handling of a literal zero obliterating its unit is special cased in the type checker? During elaboration (turning an `act::Statement` into a `typed_ast::Statement`), special care is taken to make sure scalars equal to 0 become polymorphic. But this is not extended to values equal to 0, such as 1-1. Therefore you end up being able to compute `cos(0m)` but not `cos((1-1)m`, as the second’s zero value is not polymorphic.

The solution probably has to be applied at a lower level than elaboration — something at runtime that makes a unitful zero coercible to any other unit. Then at this stage `unit_of` can just read off the unit currently attached to its zero. Of course, this would drastically change how `unit_of` is implemented, as it could no longer be an ordinary function; it would have to special-cased into the runtime.

sharkdp:
> Also it appears only to affect the literal 0, not something like 1-1.

> scalars equal to 0 become polymorphic. But this is not extended to values equal to 0, such as 1-1

[...]

PR #652, opened 2024-11-04
Add more descriptive `unit_of` error message

Goju-Ryu:
The previous error message was prone to confusion, and although temporary I think it is best to return a more helpful message.
I hope this change will help with https://github.com/sharkdp/numbat/issues/621 and similar issues until https://github.com/sharkdp/numbat/issues/521 is resolved.


sharkdp:
Thank you. I had a nice message for this once, but then I needed to change the implementation and put a temporary message in place, forgetting to copy it over.

PR #659, opened 2024-11-21
Add installation guide to VS Code Extension readme

Goju-Ryu:


vscode-extension/README.md
```diff
@@ -9,9 +9,16 @@ The Numbat extension provides language support for the [Numbat](https://github.c
 - comment toggling
 - autosurrounding brackets
 
+## Installation
+
+To install this extension you will need to download the extension installation file (VSIX file), which can be found in the assets of each [numbat release](https://github.com/sharkdp/numbat/releases) after v1.14.0.
```
sharkdp, line 14: > which can be found in the assets

can it? Am I missing something?

Goju-Ryu: I thought that was the approach we agreed on for the distribution of the add-in. I might have misunderstood but that at least was my understanding when writing this paragraph.

sharkdp: Ah, sorry — yes. But we haven't implemented that (in `.github/workflows/ci.yml`) yet :smile: To be honest, I don't even know what VSIX is and/or how to generate it.

Goju-Ryu: Aah fair. VSIX is just a visual studio code installation executable. So just a packaging format for code extensions to share them. It can be generated via npm with this script: https://github.com/sharkdp/numbat/blob/master/vscode-extension/package.json#L38

I didn't think about changing the GitHub Action but I could probably add it as a step without a problem.

Goju-Ryu:
There may be some pluming needed for the files to be added when creating a new release but the setup for generating the VSIX file is now in place

vscode-extension/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "numbat",
+    "name": "numbat-extension",
```
sharkdp, line 2: Maybe numbat-vscode-extension?

.github/workflows/ci.yml
```diff
@@ -355,6 +395,12 @@ jobs:
       with:
         name: ${{ steps.debian-package.outputs.DPKG_NAME }}
         path: ${{ steps.debian-package.outputs.DPKG_PATH }}
+    
+    - name: "Artifact upload: Debian package"
```
sharkdp, line 399:
```suggestion
    - name: "Artifact upload: VSCode extension"
```

sharkdp: I think you need to move this task to the vscode_extension_package job. It currently runs for the matrix build, which then fails to upload anything:

https://github.com/sharkdp/numbat/actions/runs/12576115802/job/35051885317?pr=659#step:16:17

vscode-extension/package.json
Goju-Ryu: Sure, no problem

.github/workflows/ci.yml
Goju-Ryu: I think you are right. I was playing around with it a bit to understand the steps and what what should go. I seem to have missed moving this with the rest.

Goju-Ryu:
> Looks like that doesn't work yet: https://github.com/sharkdp/numbat/actions/runs/12614910052/job/35154340319?pr=659#step:7:13. 

I will take another stab at it then. 

> You can also see all the uploaded artifacts here: https://github.com/sharkdp/numbat/actions/runs/12614910052?pr=659

Thank you! I'd been looking for where to check the uploaded artefacts, but couldn't find it. This will be very helpful!

Goju-Ryu:
It should now be fixed. Problem was that I set the base path for the bash scrips as being in the vscode-extension folder, but the upload-artefact action didn't use that base path implicitly like the bash scripts did.

sharkdp (approved):
Thank you

PR #661, opened 2024-11-25
Unit list based human function

### LINKED ISSUE #622: `human` gives last minute in seconds
I was working on implementing the `human` function in terms of `unit_list`, checking for inconsistencies in output when I came across this odd result from the current `human` function. When given specific values it will print a result with minutes followed by 60 seconds as shown in this image:
image
[try it here](https://numbat.dev/?q=123+min+-%3E+human%E2%8F%8E1234+min+-%3E+human%E2%8F%8E)

My implementation with `unit_list` produces the expected result, so a fix is coming up soon.

Goju-Ryu:
This PR replaces the current implementation of `human` with one that utilizes the `unit_list`. It still produces a string representation and produces the same outputs as the previous implementation, except it fixes #622.


sharkdp:
Awesome — thank you!

PR #663, opened 2024-11-30
Initial implementation of frontend-configurable command runner

rben01:


rben01:
Related to #584

numbat-cli/src/main.rs
```diff
@@ -353,121 +355,134 @@ impl Cli {
...
+                CommandControlFlow::Normal
+            })
+            .enable_list(|ctx, item| {
+                let m = match item {
+                    None => ctx.print_environment(),
+                    Some(command::ListItems::Functions) => ctx.print_functions(),
+                    Some(command::ListItems::Dimensions) => ctx.print_dimensions(),
+                    Some(command::ListItems::Variables) => ctx.print_variables(),
+                    Some(command::ListItems::Units) => ctx.print_units(),
+                };
```
sharkdp, line 380: This looks like something that could be part of the command runner? The callback would then directly take the `Markup` as an argument?

We could go one step further and let frontends basically specify a Markup=>Output function once. And that could be re-used generically for `help`, `info`, `list …`, etc.

numbat/src/command.rs
```diff
@@ -46,6 +47,104 @@ impl FromStr for CommandKind {
...
+        } = args;
+
+        match cmd {
+            Command::Help => self.help.unwrap()(),
+            Command::Info { item } => self.info.unwrap()(ctx, item),
+            Command::List { items } => self.list.unwrap()(ctx, items),
+            Command::Clear => self.clear.unwrap()(editor),
+            Command::Save { dst } => self.save.unwrap()(ctx, session_history, dst, interactive),
+            Command::Quit => self.quit.unwrap()(),
+        }
```
sharkdp, line 144: Is it possible to somehow have a tighter coupling between the `CommandRunner` and the `SourcelessCommandParser`? (potentially as a new struct "layered" on top)

It feels like there could be a way where the `.unwrap()`s here and the `let is_supported = …` part below could somehow be tied together.

Sorry for the high-level comment. Let me know if it's unclear.

sharkdp (commented):
Thank you very much for this. It looks like this is going in the right direction! Before I take a more detailed look, I have two high-level comments. In general, the more code we can move from `numbat-cli` towards `numbat`, the better. It doesn't look like we made a lot of progress on this front, given that `main.rs` is longer than it was before.

It might make sense to also work on the WASM frontend, to see how it all plays out.

numbat-cli/src/main.rs
rben01: Just to clarify: you're saying that the user of the command runner would not provide the whole closure above, just `println!("{}", ansi_format(&m, false));`, and the command runner itself would run that match statement and return `Continue` (since there is no other reasonable way to implement the guts of `list`)?

numbat/src/command.rs
rben01: You are correct. Made a bunch of changes, I think in line with what you were thinking.

Basically, instead of producing a `Command`, and then checking whether it's enabled, we only produce a `CommandKind`, check whether that's enabled, and then proceed to parse args. This combines the `is_supported` and `unwrap()`s into one logical component (a single `match`).

rben01: The changes I made amounted to a pile of spaghetti. I realized that the correct way to implement this:

> It feels like there could be a way where the .unwrap()s here and the let is_supported = … part below could somehow be tied together.

is to stick the functions themselves into the `Command` enum.  Then you can only produce a command if you've already got the corresponding function (i.e. it was `Some(fn(...))` and not `None`).'

So now we have two functions: `get_command` that produces such a command (and is the primary test point), and `try_run_command` that runs the command we just got, which has the necessary functions already stored in whichever variant we're running.

numbat-cli/src/main.rs
sharkdp: Exactly

rben01: I think basically fits the bill now.

numbat/src/command.rs
```diff
@@ -46,49 +53,365 @@ impl FromStr for CommandKind {
...
+pub enum CommandControlFlow {
+    Continue,
+    Return,
+    NotACommand,
+}
+
+pub struct CommandContext<'ctx, 'editor, Editor> {
+    pub ctx: &'ctx mut Context,
+    pub editor: &'editor mut Editor,
+}
```
sharkdp, line 67: I have a minor preference to inline these two arguments instead of bundling them in a public `CommandContext` struct. Right now, they are immediately unpacked in `try_run_command` anyways. Or is there another reason for this struct?

numbat/src/command.rs
```diff
@@ -46,49 +53,365 @@ impl FromStr for CommandKind {
...
+
+pub struct CommandRunner<Editor> {
+    print_markup: Option<fn(&Markup)>,
+    clear: Option<fn(&mut Editor) -> CommandControlFlow>,
+    session_history: Option<SessionHistory>,
+    quit: Option<()>,
+}
+
+impl<Editor> CommandRunner<Editor> {
+    pub fn new_all_disabled() -> Self {
```
sharkdp, line 156: Minor: I think I would prefer a simple 
```suggestion
    pub fn new() -> Self {
```

numbat/src/command.rs
```diff
@@ -46,49 +53,365 @@ impl FromStr for CommandKind {
...
+            clear: None,
+            session_history: None,
+            quit: None,
+        }
+    }
+
+    pub fn enable_print_markup(mut self, action: fn(&Markup)) -> Self {
+        self.print_markup = Some(action);
+        self
+    }
```
sharkdp, line 168: So this basically enables `list`/`info`/`help` and some other things. Maybe we could call this something like
```suggestion
    pub fn print_markup_using(mut self, action: fn(&Markup)) -> Self {
        self.print_markup = Some(action);
        self
    }
```
or just
```suggestion
    pub fn print_using(mut self, action: fn(&Markup)) -> Self {
        self.print_markup = Some(action);
        self
    }
```

numbat/src/command.rs
```diff
@@ -177,112 +479,79 @@ impl<'a> CommandParser<'a> {
...
+        assert!(runner.get_command(input, ctx).is_err());
     }
 
+    // fn parse(input: &'static str) -> Result<Command, ParseError> {
+    //     parser(input).unwrap().parse_command()
+    // }
+
+    // fn parse_uw(input: &'static str) -> Command {
+    //     parse(input).unwrap()
+    // }
```
sharkdp, line 553: Remove?

numbat-cli/src/main.rs
```diff
@@ -351,123 +350,71 @@ impl Cli {
...
     ) -> Result<()> {
-        let mut session_history = SessionHistory::default();
+        let mut cmd_runner =
+            CommandRunner::<Editor<NumbatHelper, DefaultHistory>>::new_all_disabled()
+                .enable_print_markup(|m| println!("{}", ansi_format(m, true)))
+                .enable_clear(|rl| match rl.clear_screen() {
+                    Ok(_) => CommandControlFlow::Continue,
+                    Err(_) => CommandControlFlow::Return,
+                })
+                .enable_save(SessionHistory::default())
```
sharkdp, line 360: What is the reasoning for a `session_history` argument here? Could this be `.enable_save()`? Or do we want frontends to be able to customize the session history object before passing it in for some reason?

numbat/src/command.rs
```diff
@@ -177,112 +479,79 @@ impl<'a> CommandParser<'a> {
...
+        CommandParser::new(input, 0)
     }
-}
 
-#[cfg(test)]
-mod test {
-    use super::*;
+    fn parser_uw(input: &'static str) -> CommandParser<'static> {
+        parser(input).unwrap()
+    }
```
sharkdp, line 531: This function is only used with `.word_boundaries`. Should it be modified to return them directly?

Minor: You can annotate functions like this with `#[track_caller]` so that a possible unwrap panic will be reported at the call site (where the actual mistake is), not here.

numbat/src/command.rs
```diff
@@ -46,49 +53,365 @@ impl FromStr for CommandKind {
...
+                + m::space()
+                + m::string(dst.to_compact_string());
+            print_markup(&markup)
+        }
+
+        Ok(())
+    }
+}
+
+pub struct CommandRunner<Editor> {
```
sharkdp, line 148: Minor: Maybe use
```suggestion
pub struct CommandRunner<Editor = ()> {
```
here, so clients which do not set up a `clear` command can simply use `CommandRunner` instead of `CommandRunner::<()>`

numbat/src/command.rs
```diff
@@ -398,91 +655,181 @@ mod test {
...
 
-        assert_eq!(parse!("quit").unwrap(), Command::Quit);
-        assert!(parse!("quit arg").is_err());
-        assert!(parse!("quit arg1 arg2").is_err());
+        let mut ctx = Context::new_without_importer();
 
-        assert_eq!(parse!("exit").unwrap(), Command::Quit);
-        assert!(parse!("exit arg").is_err());
-        assert!(parse!("exit arg1 arg2").is_err());
+        let runner = CommandRunner::<()>::new_all_disabled()
```
sharkdp, line 812: This would be
```suggestion
        let runner = CommandRunner::new_all_disabled()
```
with the suggestion above.

sharkdp (approved):
This is fantastic, thank you very much!

I have added a few minor review comments, and I have one remaining question:

The web frontend has a `reset` command:

https://github.com/sharkdp/numbat/blob/77decee691bf7fe77eeacdf313bc894edf084468/numbat-wasm/www/index.js#L44-L49

Do you think we can implement this using the current architecture?

We basically need mutable access to the `Context` to reset it to some (frontend-configurable) default state. And we currently also clear the screen from the `reset` command. I guess that could be done frontend-independently. If `reset` is exectuted, we could check if `clear_fn` is set, and just run it in addition.

> Making a note here to not forget to also merge the `copy-to-clipboard` branch into this one, as this currently does not include that command.

Can we maybe do this in a second step?

numbat/src/command.rs
rben01: No reason, likely a vestige of past code.

numbat/src/command.rs
rben01: How about `print_with`?

numbat-cli/src/main.rs
rben01: Yes, I wanted to leave the door open to letting frontends customize the SessionHistory, although currently there is no customization point.

rben01:
Pushed an initial implementation of `reset`. It simply asks the frontend to provide a function that can construct a fresh Context. (In theory it could reinitialize an existing Context, which would presumably be cheaper, but I didn't see how to achieve this easily.)

numbat/src/command.rs
sharkdp: Sounds good!

sharkdp:
Thank you for the updates!

sharkdp (approved):

PR #666, opened 2024-12-12
Switched to newer macOS GitHub Actions images

rben01:
Removed the obsolete macos-12, as GitHub no longer supports it 
Switched to macos-13 on x86 (Intel) and macos-15 on ARM64 (Apple Silicon)


sharkdp:
Thank you!

PR #677, opened 2025-01-05
Add missing metric cooking spoon units

### LINKED ISSUE #383: Metric teaspoon and tablespoon missing
These are defined as 5 mL and 15 mL, respectively, and are used in the UK and elsewhere where metric is used. See https://en.wikipedia.org/wiki/Tablespoon and https://en.wikipedia.org/wiki/Teaspoon#Metric_teaspoon.

The situation is even more confusing than at first glance because there's also an Australian teaspoon and tablespoon.

This is an extension of #375.

Goju-Ryu:
Closes #383 by adding metric table- and teaspoons as the default spoon units. The old default US customary have been given a `US_` prefix. Additionally I added the  imperial version with the `UK_` prefix as well for completeness [^1]

[^1]: There are a lot more regional variants but I think that is outside the scope of this PR without discussing it first. See a few of them here: https://en.wikipedia.org/wiki/Cooking_weights_and_measures#Metric_measures


sharkdp:
Thank you!

PR #695, opened 2025-02-21
Multi factorial

irevoire:
Hey, currently numbat doesn't handle multifactorial correctly, see the following screenshot:
image

This should in fact returns (5 * 3 * 1): https://en.wikipedia.org/wiki/Double_factorial#Generalizations

This PR adds support for it, it's a breaking change though since numbat was already accepting the expression.

### DIFF
--- a/numbat/tests/interpreter.rs
+++ b/numbat/tests/interpreter.rs
@@ -169,6 +169,12 @@ fn test_factorial() {
     expect_output("-5!", "-120");
     expect_output("-(5!)", "-120");
     expect_output("-(5)!", "-120");
+    expect_output("5!!", "15");
+    expect_output("5!!!", "10");
+    expect_output("5!!!!", "5");
+    expect_output("5!!!!!", "5");
+    expect_output("5!!!!!!", "5");
+    expect_output("(3!)!!!", "18");
 
     expect_failure(
         "(-1)!",
--- a/numbat/src/ast.rs
+++ b/numbat/src/ast.rs
@@ -1,3 +1,5 @@
+use std::num::NonZeroUsize;
+
 use crate::markup as m;
 use crate::resolver::ModulePathBorrowed;
 use crate::span::Span;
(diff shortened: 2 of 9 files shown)


numbat/tests/interpreter.rs
```diff
@@ -169,6 +169,12 @@ fn test_factorial() {
     expect_output("-5!", "-120");
     expect_output("-(5!)", "-120");
     expect_output("-(5)!", "-120");
+    expect_output("5!!", "15");
```
sharkdp, line 172: Can we expand at least the double-factorial tests a bit more and include some edge cases, as well as an example for even numbers. And maybe we could write the tests in a way that it's easy to see that we get the desired result. By moving them to a Numbat-source test suite? Something like
```
assert_eq(7!!, 7 * 5 * 3 * 1)
assert_eq(8!!, 8 * 4 * 2)
```

sharkdp (approved):
Thank you very much, this looks great. I'm not concerned about the breaking change.

numbat/tests/interpreter.rs
irevoire: Hey, that should be done in https://github.com/sharkdp/numbat/pull/695/commits/fb066f0e9c82c42c4b6be3d2db1a75c810157fb4

Could you confirm me that we cannot write an assert in a function in numbat currently?
I wanted to write a more « procedural » test with a recursive function that assert at each step that the function works on all numbers from 0 to 10 with multiple orders but then I didn’t see how to do it 🤔

Goju-Ryu: I can confirm that it isn't currently possible. I wanted to make a function to encapsulate a repeating pattern in the tests I wrote and learned it wasn't possible yet. 
There was some talk about the possibility of numbat having macros for such cases, but that's still just an idea.

sharkdp (approved):
Thank you

PR #698, opened 2025-03-04
Add immediate saving of every line of input to history file

rben01:
The current implementation shows an error after every line of input if appending to the history file fails. I could see this being annoying, but at the same time it is the safe option compared to only showing the warning once and having a user think they've fixed their issue with the history file, and then having it turn out they didn't, and having their history silently lost.


sharkdp:
Thank you very much!

> The current implementation shows an error after every line of input if appending to the history file fails. I could see this being annoying, but at the same time it is the safe option compared to only showing the warning once and having a user think they've fixed their issue with the history file, and then having it turn out they didn't, and having their history silently lost.

Sounds okay to me, I don't think it's a particularly common error, hopefully. Thank you for bringing it up.

PR #700, opened 2025-03-05
Don't generate multiple VM constants for variables with aliases

triallax:
While poking around with Numbat's VM, I noticed if there's a variable
with aliases, like so:

```
@Aliases(foo2)
let foo = 2 m ^ 3
```

then that is (hand-waving a bit here) compiled down as if it were this:

```
let foo = 2 m ^ 3
let foo2 = 2 m ^ 3
```

leading to redundant VM constants (e.g. two 2 constants),
redundant calculations, and even possibly different values if the
right-hand side of the assignment is not pure (if it contains `random()`
for instance).

This commit resolves that duplication by using the variable's name and
all its aliases when resolving identifiers to locals, instead of pushing
a separate local for each alias. While at it, I've also removed the
`depth` field from the `Local` struct, as it is equivalent to the index
of the vector of locals in which the local resides. The only place in
which the field is used is completely gratuituous anyway (the condition
always evaluates to true since all locals in `self.locals[d]` have the
depth `d`).

There are no real performance improvements from this change, but it does
give a nice reduction in the number of instructions and constants
generated by the bytecode writer. Specifically, for a very simple
Numbat program that just evaluates `1`, the number of constants goes
down from 780 to 736, and the number of instructions in the `<main>`
.CODE section goes down from 1235 to 1055 instructions.

### DIFF
--- a/numbat/src/bytecode_interpreter.rs
+++ b/numbat/src/bytecode_interpreter.rs
@@ -32,8 +32,7 @@ pub struct LocalMetadata {
 
 #[derive(Debug, Clone)]
 pub struct Local {
-    identifier: CompactString,
-    depth: usize,
+    identifiers: Vec<CompactString>,
     pub metadata: LocalMetadata,
 }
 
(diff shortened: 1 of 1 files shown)


numbat/src/bytecode_interpreter.rs
```diff
@@ -32,8 +32,7 @@ pub struct LocalMetadata {
 
 #[derive(Debug, Clone)]
 pub struct Local {
-    identifier: CompactString,
-    depth: usize,
+    identifiers: Vec<CompactString>,
```
triallax, line 35: would be nice to deduplicate this with `aliases` inside the `metadata` field, but I'm not sure what's the best way to go about having one field for both

sharkdp (approved):
This is great  — thank you very much!

PR #705, opened 2025-03-20
doc: add more references to _ and 'ans'

martinetd:
_/ans are only mentioned in the cli-usage page, this was hard to find so add it to the syntax overviews as well

------------
I actually found `_` after trying a lot of variations thinking there couldn't possibly not be such a keyword (I use `.` all the time in bc), and later found `ans` grepping for `_` in the doc... Having this here would have saved me 10 mins, so taking an extra few minutes to save anyone else the trouble.
Thanks for this and your other tool!

### DIFF
--- a/book/src/example-numbat_syntax.md
+++ b/book/src/example-numbat_syntax.md
@@ -58,6 +58,8 @@ pi/3 + pi |> cos  # Same, 'arg |> f' is equivalent to 'f(arg)'
                   # which makes it very useful for interactive
                   # terminals (press up-arrow, and add '|> f')
 
+_                 # Result of last calculation, also 'ans'
+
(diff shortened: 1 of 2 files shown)


Goju-Ryu:
I agree that it is too hidden away and I would like to update the documentation to highlight this feature a bit more. I am however not sure where it would make sense to add it. I have considered [tutorial](https://numbat.dev/doc//tutorial.html), [syntax overview](https://numbat.dev/doc//example-numbat_syntax.html) and a new page with handy tips and tricks to highlight some of the more useful features many may not discover at first.

sharkdp (approved):
Thank you. Let's add it here until we find a better place.

PR #708, opened 2025-03-30
Add Pixi installation instructions

apcamargo:
This PR adds instructions for installing numbat on Linux, macOS, and Windows using [Pixi](https://pixi.sh/) with the [`pixi global`](https://pixi.sh/latest/global_tools/introduction/) command.


sharkdp:
Thank you, I think I'd like to keep this focused on "mainstream" package managers (without having to further define what I mean by that, if that's okay).

PR #710, opened 2025-04-06
Massively improve `str_find` performance

Goju-Ryu:
I recently discovered that `str_find` had very poor performance. This PR fixes it by altering it's implementation such that it goes from having a time complexity of $O(2^n)$ where $n$ is the length of the haystack to $O(n)$.  

Here is a video demonstrating the speedup using the example from the documentation:
strings nbt

_Note:_
This also improves the `str_contains` function as it is implemented using `str_find`.


sharkdp (approved):
Fantastic — thank you!

PR #723, opened 2025-05-13
Store currencies from currencies.nbt to create list of currencies in `const CURRENCY_IDENTIFIERS: &[&str] = ...`

rben01:
Adds build.rs to parse and write the currencies in currencies.nbt to a file, then read from the file in `interpret_with_settings`.

Since the list of currencies is sorted, this would allow for binary searching the list of currencies for a given identifier instead of linear searching. Not sure how much we care — in the happy path, this search only happens once per interpreter session anyway.


rben01:
Since we have a build.rs now we can also consider using it to precompute and store the output of `help`, see #591.

PR #724, opened 2025-05-13
Use IndexMap for Vm.ffi_callables for fast lookup by name

rben01:
To facilitate this, removed `ForeignFunction`’s `name` field (since it's the key of the map) 

Correspondingly, made `ProcedureKind` itself aware of its own name, rather than relying on the `ForeignFunction` to provide it


sharkdp (approved):
Thank you!

PR #726, opened 2025-05-15
Dedupe vm constants

rben01:
Builds on https://github.com/sharkdp/numbat/pull/712 by making `Constant: Hash` so that we can stick it in an `IndexSet`. Therefore all the deduplication logic can rest solely on the nature of an `IndexSet`.

This required quite a few changes. Most notably, `Number` is now opaque instead of providing access to its contained f64 via `.0` because it no longer contains an f64, but rather a u64, which is converted to and from an f64 with `f6::to_bits`/`f64::from_bits`. But also, we have a new `Dummy` variant in `Constant` that stores an auto-incremented value each time it's created. (This is of course necessary for the deduplication logic to work correctly; identical dummies would occupy the same slot in the map.)

### DIFF
--- a/numbat/src/bytecode_interpreter.rs
+++ b/numbat/src/bytecode_interpreter.rs
@@ -51,7 +51,7 @@ impl BytecodeInterpreter {
     fn compile_expression(&mut self, expr: &Expression) -> Result<()> {
         match expr {
             Expression::Scalar(_span, n, _type) => {
-                let index = self.vm.add_constant(Constant::Scalar(n.to_f64()));
+                let index = self.vm.add_constant(Constant::Scalar(n.to_f64()), true);
                 self.vm.add_op1(Op::LoadConstant, index);
             }
             Expression::Identifier(_span, identifier, _type) => {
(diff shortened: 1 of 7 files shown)


sharkdp:
Thank you for working on this.

Before I review this, I'd like to understand the core motivation better. Is there any upsides to this beside what's mentioned in the PR description at https://github.com/sharkdp/numbat/pull/712#issue-2977793307?

rben01:
I believe this implementation allows for constant-time lookup of stored values, whereas the referenced implementation uses linear lookup (calls to .iter().position())

sharkdp:
> This required quite a few changes. Most notably, `Number` is now opaque instead of providing access to its contained f64 via `.0` because it no longer contains an f64, but rather a u64, which is converted to and from an f64 with `f6::to_bits`/`f64::from_bits`

I'm skeptical about this part. This won't work if we want to replace our number representation, right?

Goju-Ryu:
> I'm skeptical about this part. This won't work if we want to replace our number representation, right?

I was thinking this could be a step in that direction. Disentangling the underlying representation by making it opaque. I might be missing some consequences though.

sharkdp:
Making it opaque is certainly a good idea, but it sounded like we use the bits of the f64 for the integer? How would that work for some more complex heap-allocated precision datatype?

PR #733, opened 2025-05-24
feat: Add Nit convertion

### LINKED ISSUE #709: feat: Add Nit conversion
https://en.wikipedia.org/wiki/Candela_per_square_metre

reyalka:
ref: https://en.wikipedia.org/wiki/Candela_per_square_metre

resolve #709

### DIFF
--- a/numbat/modules/units/si.nbt
+++ b/numbat/modules/units/si.nbt
@@ -150,7 +150,7 @@ unit lumen: LuminousFlux = candela steradian
 unit lux: Illuminance = lumen / meter^2
 
 @name("Luminance")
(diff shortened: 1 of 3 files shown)


numbat/modules/units/si.nbt
```diff
@@ -149,6 +149,12 @@ unit lumen: LuminousFlux = candela steradian
 @aliases(lx: short)
 unit lux: Illuminance = lumen / meter^2
 
+@name("Luminance")
```
sharkdp, line 152: This should be `Nit`, I think?

sharkdp (approved):
Thank you very much, just one minor comment.

PR #742, opened 2025-06-12
feat: store history at NUMBAT_HISTORY if variable exists

t-webber:
Lots of REPLs and CLIs allow you to store the history of commands, for example:

`PYTHON_HISTORY`, for the python repl
`NODE_REPL_HISTORY`, for the node repl
`TS_NODE_REPL_HISTORY`, for the ts-node repl

After this PR, the history of commands run in `numbat` will be stored using the `NUMBAT_HISTORY` file, instead of using data_dir. data_dir is, though, used as a fallback.


sharkdp:
Thank you!

PR #743, opened 2025-06-14
feat: cubic_equation to solve cubic equations in extra::algebra

t-webber:
The `extra::algebra` contains a `quadratic_equation` function that computes the roots of a quadratic equation. This PR adds in this module a `cubic_equation` function that does the same for cubic equations.

Tests were added and documentation updated


sharkdp:
Thank you!

Not implying that we necessarily have to do this, but have you considered implementing this in Numbat itself? I understand that this might be more difficult, and might not be benefitial in terms of runtime, but it's always great if we can implement prelude functions in Numbat itself. It helps language development, and avoids duplication across the FFI. Having fewer functions imlemented in Rust also helps us when we make changes to the FFI or compiler internals.

t-webber:
Ok, fair point. I'll give it a try!

Btw, can I edit the `cbrt` function so that `cbrt(-1) = -1` instead of `NaN`? I need this for my use case

t-webber:
Thank you for your help! I hope that this agrees better with your expectations!

Goju-Ryu:
This looks really good, I’m happy it worked. 

I have some minor formatting nitpicks, but I don't know if they are worth addressing when we don't have a formal style guide yet. 
I’m however very pleased to see the argument order of map2 being exactly as I hoped it would be. 

Thank you for this nontrivial rewrite into numbat.

PR #750, opened 2025-07-06
Make logical operators short-circuit where possible

triallax:
More efficient but also in line with most other languages out there.

### DIFF
--- a/numbat/src/bytecode_interpreter.rs
+++ b/numbat/src/bytecode_interpreter.rs
@@ -109,6 +109,32 @@ impl BytecodeInterpreter {
                 self.compile_expression(lhs)?;
                 self.vm.add_op(Op::LogicalNeg);
             }
+            Expression::BinaryOperator(
+                _span,
+                op @ (BinaryOperator::LogicalOr | BinaryOperator::LogicalAnd),
+                lhs,
+                rhs,
+                _type,
+            ) => {
+                self.compile_expression(lhs)?;
+
+                let jump_offset = self.vm.current_offset() + 1; // +1 for the opcode
+                self.vm.add_op1(
+                    match op {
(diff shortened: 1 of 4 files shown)


numbat/src/bytecode_interpreter.rs
```diff
@@ -109,6 +109,32 @@ impl BytecodeInterpreter {
...
+            Expression::BinaryOperator(
+                _span,
+                op @ (BinaryOperator::LogicalOr | BinaryOperator::LogicalAnd),
+                lhs,
+                rhs,
+                _type,
+            ) => {
+                self.compile_expression(lhs)?;
+
+                let jump_offset = self.vm.current_offset() + 1; // +1 for the opcode
```
Goju-Ryu, line 121: How come you add 1 here? Can't you create it after adding the opcode to avoid the hardcoded +1?

numbat/src/bytecode_interpreter.rs
```diff
@@ -109,6 +109,32 @@ impl BytecodeInterpreter {
...
+                    },
+                    0xffff,
+                );
+                self.vm.add_op(Op::Pop);
+
+                self.compile_expression(rhs)?;
+
+                let else_block_offset = self.vm.current_offset();
+                self.vm
+                    .patch_u16_value_at(jump_offset, else_block_offset - (jump_offset + 2));
```
Goju-Ryu, line 136: This is a bit unclear to me. I think I follow, but a comment to explain what is happening would be helpful.

Goju-Ryu (commented):
Thank you for this contribution, I’m excited to see short-circuiting being added to numbat. 

I don't have a good enough understanding of this part of the project to give much feedback, but it looks really promising. I had two comments, mostly just to clarify some points.

PR #751, opened 2025-07-09
Add metric prefixes to arcsec

### LINKED ISSUE #706: Support for milliarcseconds and microarcseconds
In Astronomy it is common to express very small angles in parts of an arcseond. Usually milliarcseconds and in high resolution imaging even microarcseconds.

Currently numbat does not add metric prefixes to those units.

If there is interest, I have couple branches with:

1. adding metric prefixes to deg, arcmin and arcsec. [[1]](https://github.com/sharkdp/numbat/compare/master...phsilva:numbat:milliarcsec-v1?expand=1)
2. adding a specific unit just for milliarcsec to allow the usage of the common abbreviation `mas`. [[2]](https://github.com/sharkdp/numbat/compare/master...phsilva:numbat:milliarcsec-v2?expand=1)

Even tough approach one makes it more uniform, I like the commodity of using `mas` and potentially `μas` as well.

Let me know if there is interest and I can add a PR with either solution.

phsilva:
Fix #706

### DIFF
--- a/numbat/modules/units/si.nbt
+++ b/numbat/modules/units/si.nbt
@@ -213,6 +213,7 @@ unit arcminute: Angle = 1 / 60 × degree
 
 @name("Second of arc")
 @url("https://en.wikipedia.org/wiki/Minute_and_second_of_arc")
+@metric_prefixes
 @aliases(arcseconds, arcsec, ″: short)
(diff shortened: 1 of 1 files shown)


Goju-Ryu:
Thank you!

PR #752, opened 2025-07-12
Make compilation infallible in bytecode_interpreter where possible

rben01:
`compile_expression` and `compile_define_variable` were actually infallible — the only errors they could have returned were from calling themselves.


sharkdp (approved):
Thank you!

PR #754, opened 2025-07-13
Implement bitwise operators.

rehelmin:
The code in this pull request adds implementations for the following bitwise operators:

- And
- Or
- Xor
- Not
- Bit shift left
- Bit shift right


rehelmin:
This is my first pull request. I'm realizing I should probably be doing it from a feature branch rather than my master branch.

PR #755, opened 2025-07-13
Implement bitwise operators

rehelmin:
The code in this pull request adds implementations for the following bitwise operators:

- And
- Or
- Xor
- Not
- Bit shift left
- Bit shift right

Note that typically used '^' character for xor currently clashes with the power operator in numbat, so I have used the more cumbersome to type, but unique ⨁ character for performing the xor operation.

### DIFF
--- a/book/src/list-functions-math.md
+++ b/book/src/list-functions-math.md
@@ -321,6 +321,38 @@ fn mod<T: Dim>(a: T, b: T) -> T
 
 </details>
 
+### `wrapped_shl` (Bitwise left shift)
+
+```nbt
+fn wrapped_shl(lhs: Scalar, rhs: Scalar) -> Scalar
+```
+
+<details>
+<summary>Examples</summary>
+
+<pre><div class="buttons"><button class="fa fa-play play-button" title="Run this code" aria-label="Run this code"  onclick=" window.open('https://numbat.dev/?q=wrapped%5Fshl%280b0000%5F0101%2C%204%29%20%2D%3E%20bin')""></button></div><code class="language-nbt hljs numbat">wrapped_shl(0b0000_0101, 4) -> bin
+
+    = "0b1010000"    [String]
+</code></pre>
+
+</details>
+
+### `wrapped_shr` (Bitwise right shift)
+
+```nbt
+fn wrapped_shr(lhs: Scalar, rhs: Scalar) -> Scalar
+```
+
+<details>
+<summary>Examples</summary>
+
+<pre><div class="buttons"><button class="fa fa-play play-button" title="Run this code" aria-label="Run this code"  onclick=" window.open('https://numbat.dev/?q=wrapped%5Fshr%280b0101%5F0000%2C%204%29%20%2D%3E%20bin')""></button></div><code class="language-nbt hljs numbat">wrapped_shr(0b0101_0000, 4) -> bin
+
+    = "0b101"    [String]
+</code></pre>
+
+</details>
+
 ## Transcendental functions
 
 Defined in: `math::transcendental`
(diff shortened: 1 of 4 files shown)


rehelmin:
Updated. I'm not sure why all the builds are all failing now though? I'm able to build it just fine on my workstation.

numbat/src/number.rs
```diff
@@ -136,6 +136,46 @@ impl std::ops::Mul for Number {
...
+    fn bitxor(self, rhs: Self) -> Self::Output {
+        Number((self.0 as i64 ^ rhs.0 as i64) as f64)
+    }
+}
+
+impl std::ops::Shl for Number {
+    type Output = Number;
+
+    fn shl(self, rhs: Self) -> Self::Output {
+        Number(((self.0 as i64) << rhs.0 as i64) as f64)
```
sharkdp, line 167: Thank you for working on this!

I believe we need to write a few more test cases for extreme inputs, and we need to improve error handling here. These bitwise operations can overflow, and we don't want Numbat to crash in those cases (but rather emit some error, similar to the division-by-zero error). I'm also not sure about these `as` casts...

For example:
```
>>> 1 << 63

  1 << 63

    = -9.22337e+18

>>> 1 << 64

thread 'main' panicked at numbat/src/number.rs:167:16:
attempt to shift left with overflow
```

rehelmin: > I'm also not sure about these as casts...

What aspect of them are you unsure about? It was the only way I could figure out how to work around the fact that the underlying primitive type for Number is f64. None of the bitwise operations make sense in the context of an f64 type.

Good point on the overflowing. Looks like there are a bunch of different options that have different behavior defined for overflows that could occur during the shifting operations:

- unbounded: Any overflowing bits are dropped. If the rhs is larger than the type being shifted then all the bits are dropped resulting in a value of 0.
- wrapping: Masks off the higher order rhs bits that would cause an overflow to occur.
- checked: Returns None if the size of the rhs is larger than the size of the type being shifted.

My preference is the unbounded implementation, but I could be convinced that it would be better to throw an error rather than to silently fail in a defined way.

rehelmin: One side effect of the type casts is that it silently truncates the decimal part of any floats that someone tries to perform a bitwise oepration on. For example `5.5 | 2.1` silently truncates the .5 and .1 and gives the value of 7. I can see this being somewhat undesired. We could probably add some error checking that only allows bitwise operations to be performed when both arguments are scalars. I've already added this error checking for performing the bitwise not. I don't think it would be a big step to add it to the rest of the operations.

rehelmin: > I think it might be worth having more than one implementation, but I agree with preferring the unbounded version for the operator. The other versions could potentially be added as ffi functions to give the option if needed.

I would definitely need some help with the ffi function implementation. These updates are my first forays into the world of rust, so I've mostly been pattern matching off of existing code to implement them.

Goju-Ryu: > One side effect of the type casts is that it silently truncates the decimal part of any floats that someone tries to perform a bitwise oepration on. For example 5.5 | 2.1 silently truncates the .5 and .1 and gives the value of 7.

Yeah that would not be ideal. 

> We could probably add some error checking that only allows bitwise operations to be performed when both arguments are scalars. 

This sounds like a good idea to me, but I think this decision should be up to @sharkdp.

> I would definitely need some help with the ffi function implementation.

No problem, I think I can help with that. I’m not sure when I can, but I'll drop some updates with the ffi connections and a todo for where the function implementation should go if you want it. 

> These updates are my first forays into the world of rust, so I've mostly been pattern matching off of existing code to implement them.

I’m learning rust in much the same way. Thank you for taking on the challenge to add this feature 😄

Goju-Ryu: I've added a very simple ffi outline that should work for the left and right shifts. I've left comments with what still needs to be done before it's done.

Issue #395, opened 2024-03-12
Add semicolon as optional statement separator


mhmdanas:

erentar:
I would love this feature, at the moment making calculations with intermediate variables is rather unwieldy:
```
>>> 14.5*squares
error: while type checking
  ┌─ <input:1>:1:6
  │
1 │ 14.5*squares
  │      ^^^^^^^ unknown identifier
  │
  = Did you mean 'quarter'?
```

I have to go back and enter
```
>>> squares = 32
>>> 14.5*squares
```
when it would be nice to have the ability to be able to do this in a single line, for convenience reasons

PR #759, opened 2025-08-12
395 add semicolon as optional statement separator

Sondeyy:
#395


Goju-Ryu:
This looks really good. I think it is ready to merge as is, but could you add some numbat tests before I merge? Just some simple tests that show a newline separated and semicolon separated calculation is equivalent?

Issue #542, opened 2024-08-18
Return the position of runtime failure


irevoire:
Currently, numbat doesn't return anything except the error message when encountering a runtime error:
```
% numbat -e "12 / 0"
error: runtime error
 = Division by zero

Interpreter stopped
```

Ideally, it should return the line that caused the error (at least the line number and column).

Currently, this is not possible because once we generate the bytecode, we lose all information about the code source.
It seems like, in the industry, the way to make that work is to generate a second vector the sizes of the instructions that link to spans.
So while executing instruction we don't lose time with this information, but in case of an error we can find back what caused the issue.

There may be a better way to do it, I’m not an expert.

sharkdp:
> It seems like, in the industry, the way to make that work is to generate a second vector the sizes of the instructions that link to spans.

I don't know where I read about this, but that's also what I heard. It sounds very reasonable. And maybe not too hard.

sharkdp:
Or... maybe not so easy after all. Wouldn't that require stack traces as well?
```
fn f(x) = 1 / x

# 100 lines later

f(0)
```

The division by zero happens in an instruction that belongs to the code of `f`. But what would be more useful for the user would be to see the call site (one level up in the call stack) and it's corresponding span.

irevoire:
Yeah, we definitely need the stack trace as well

PR #761, opened 2025-10-07
Improve runtime errors

irevoire:
Hey! 👋 

Fixes https://github.com/sharkdp/numbat/issues/542

## Changes

- I updated the VM to store the span associated with the bytecode generated.
- I made implemented a basic stacktrace that is retrieved in case of runtime error
- At the very end I print the backtrace

## What's missing

My biggest issue is that most spans are bad. Usually, we don't save the full span, which blocks us from doing a proper print.
I don't want to fix this in this issue, as it's a large refacto of the whole AST, but if you're interested is this kind of work I think I could start working on it later. Let me know.

### Simple top-level error

Let's say I have a failure in my own code at the top level:
```rust
let a = 45
let b = 15
let c = 83 / (b * 3 - a)
print(c)
```

image

The error are composed of 3 parts:
1. The kind of error: Here it's a `runtime error` and that's the very first line
2. The code blocks, we'll show between 1 and 3 (including) pieces of code that triggered the error, starting from the "deepest" one in the stack
3. The backtrace, that's what you usually expect in case of error. It shows the level of the stack first, then a one-line "summary" of what caused the error. And on the next line, the exact file and position that caused the error

Here, since we didn't have any function, the backtrace only refers to the top-level `<main>` function.

> [!NOTE]
> You may be wondering why the first line of the backtrace is almost empty?
> That's because we don't have the full_span corresponding to the whole binop so we can only output the `/` safely

### Error originating from the prelude

```rust
use math::combinatorics
fn my_fibo(n) = fibonacci(n)

print(my_fibo(-1))
```

image

This time on the second part with the code blocks we have two code blocks.
The first one is what actually caused the error, the call to the prelude fibonacci function with a negative number.
The second one is what called the function that triggered the error.

> [!NOTE]
> Did you notice how we didn't print the codeblock of the "internal" code of the prelude?
> We can see it in the backtrace because the backtrace is complete. 
> But in the code blocks, we filter out all the `<builtin>` code to reduce the noise and only return the useful bits.

### Recursive function

```rust
fn fibo(n) =
  if n < 0
  then error("oh no")
  else fibo(n - 1) + fibo(n - 2)

print(fibo(5))
```

image

The code blocks are deduped, which means we won't generate too much noise. But the backtrace is full. This could be an issue in the future, maybe.
For now, I think we should keep it like that. 

As a reminder that's what we have currently:
image

## Performances

On main: 24.6ms to import the prelude.
On this branch 24.7ms to import the prelude.

-----

Outdated previous message:

Currently, it looks like that on a simple case:
image

The source code:
```rust
let a = 42
let b = 84

let c = a * 2

# print(c / (c - b))

fn truc(n) = 1235423 / n

truc(0)
```

I'm always displaying the trace that outputed the error, but I'm wondering if I should not just display the latest one that was made by the user? Like, if the error arise from the prelude, we don't need to see its source code. We just want to see the caller code right?

Maybe I could even display more than one piece of code? What info do you think I should add to the backtrace?

### DIFF
--- a/numbat/src/span.rs
+++ b/numbat/src/span.rs
@@ -58,9 +58,12 @@ impl Span {
     #[inline]
     pub fn in_between(left: Span, right: Span) -> Span {
         debug_assert_eq!(left.code_source_id, right.code_source_id);
+        debug_assert!(left.end <= right.start);
         crate::span::Span {
             start: left.end,
-            end: right.start,
+            // The right.start must be > left.end. But just in case it is not
+            // we'll make the new span points to left.end
+            end: right.start.max(left.end),
             // lhs and rhs should share the same code_source_id
             code_source_id: left.code_source_id,
         }


irevoire:
> Each stack entry takes up two lines. I usually search stack traces by recognizing visual patterns in the text, this becomes harder when a stack takes up multiple lines.

Yeah, about that one, I kinda think the same, but also the Rust stacktrace takes two lines, and I'm fully accustomed to it. So I'm not sure having two lines is a big deal 🤔 

About this part more specifically:
> I usually search stack traces by recognizing visual patterns in the text

I agree 1000%, since there is two kind of information to display in a stacktrace:
1. The cause of the error
2. The location in the file

I went with two lines to be sure it's always aligned where you think.

But for reference, in rust they don't show the cause of the error in the stacktrace, it looks like that:
```
thread 'main' panicked at numbat/src/diagnostic.rs:581:62:
argument of integer logarithm must be positive
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/panicking.rs:75:14
   2: core::num::int_log10::panic_for_nonpositive_argument
             at /rustc/6b00bc3880198600130e1cf62b8f8a93494488cc/library/core/src/num/int_log10.rs:147:5
   3: core::num::<impl usize>::ilog10
             at /Users/irevoire/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/num/uint_macros.rs:1356:17
...
```

So, the format is basically:
```
x: name_of_the_function_returning_the_error
      at file_location:line:col
```

While in my numbat backtrace it's:
```
x: summary_of_line_causing_the_error
    at function_name in file_location:line:col
```

Since the summary can get pretty large, I wanted to have as much space as possible, so I shoved it on its own line, alone.
Then the function_name followed by the location is not ideal as the function name can push the file location (I guess that's why they are on two different line currently).
But since numbat doesn't show the full path to a function currently, I expect it to always be pretty short 🤔 

-----

That being said, I'm very open to changing it, we just have to decide what is the most useful bits 👍 

> The error message in [0] displays as error(\"oh no\"). I’m not sure the error function should be displayed at all

I think it should; the error function is pretty oblivious, but in many cases, that will be something less obvious, like above, where it's just a `/`.
We could filter that one out as well, but I don't think it's a good idea, IMO the stacktrace should be "full" and share _all_ the details with all the file locations so you can double check everything for yourself.

> but in any case I would prefer the quotation marks to appear unescaped to make reading easier.

Good catch! I used the debug implementation, but actually, I only want to escape the `\n`, I think.

> I was also wondering, would it be possible to show prelude stack traces in non-release builds and maybe when running numbat in debug mode?

Oops, yes, that's pretty smart, especially considering I initially wrote that to help me work on the prelude 😂 
I think we could definitely enable it by default on all debug builds (when we're sure someone did a `cargo run` basically).
But maybe we should also add a CLI flag or an env variable to enable it even in production, what do you think?
If we push toward something like configurable, I'm thinking that maybe we could also output **all** the code block instead of the last three ones, what do you think?

PR #795, opened 2026-01-09
Add adaptive Runge-Kutta ODE solver (Dormand-Prince)

Ryan-D-Gast:
## Summary

Adds `dsolve_adaptive_runge_kutta`, an adaptive ODE solver using the Dormand-Prince 5(4) method.

## Changes

- New `dsolve_adaptive_runge_kutta(f, x_0, x_e, y_0, atol, rtol)` function
- Hairer & Wanner initial step size estimator
- Updated documentation and tests

## Why add this?

The existing `dsolve_runge_kutta` requires users to guess an appropriate step count. Too few steps gives poor accuracy; too many wastes computation. This is especially problematic for problems where the solution changes rapidly in some regions and slowly in others.

An adaptive solver automatically adjusts step sizes to maintain accuracy while minimizing work. Users specify error tolerances (`atol`, `rtol`) rather than step counts, making it both easier to use and more reliable.

Dormand-Prince is the standard choice for adaptive RK methods (used by MATLAB's `ode45`, SciPy's `solve_ivp` default, and Julia's `OrdinaryDiffEq.jl`).

## Side Notes

1. Numbat currently lacks optional/default parameters and keyword arguments, making it difficult to expose solver configuration (safety factor, min/max step scaling, etc.) in a clean API. The current implementation uses SciPy's defaults. If these language features are added in the future, the solver API could be extended to allow user customization.
2. Second if the language supports it which I think it does in the future I could update this function to include a "solout" function which would allow for user configurable control of the output between steps (useful for getting even points and/or recording key events).

### DIFF
--- a/book/src/prelude/functions/math.md
+++ b/book/src/prelude/functions/math.md
@@ -873,6 +873,22 @@ More information [here](https://en.wikipedia.org/wiki/Runge-Kutta_methods).
 fn dsolve_runge_kutta<X: Dim, Y: Dim>(f: Fn[(X, Y) -> Y / X], x_0: X, x_e: X, y_0: Y, steps: Scalar) -> RungeKuttaResult<X, Y>
 ```
 
+### `dsolve_adaptive_runge_kutta` (Adaptive Runge-Kutta method)
+Solve the ordinary differential equation \( y' = f(x, y) \) on the interval \( x \in [x_0, x_e] \) with initial conditions \( y(x_0) = y_0 \) using the fifth-order Dormand-Prince method with adaptive step size control.
+More information [here](https://en.wikipedia.org/wiki/Dormand%E2%80%93Prince_method).
+
+```nbt
+fn dsolve_adaptive_runge_kutta<X: Dim, Y: Dim>(f: Fn[(X, Y) -> Y / X], x_0: X, x_e: X, y_0: Y, atol: Y, rtol: Scalar) -> RungeKuttaResult<X, Y>
+```
+
+!!! example "Solve \( y' = -0.5y \) with \( y(0) = 1 \)."
+    ```nbt
+    use numerics::diff
+    fn decay(x, y) = -0.5 × y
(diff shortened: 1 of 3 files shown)


Ryan-D-Gast:
> we don't have keyword arguments or default values, but we do have [structs](https://numbat.dev/docs/basics/structs/). there's not a lot of syntactic support for working with them so far, but maybe they could be a good option for implementing this options struct, and then we could provide functions that generate common sets of options for ease of use?

Yes an option/setting struct would solve the problem. I did experiment with it but noticed like you said, that the syntactic support didn't seem easy enough for users to use (as in it felt like it made the API hard to work with). So if you add more syntactic support that would be amazing. 

My preferred addition (if not already) is a way to initialize a struct of type Foo only listing items that you want to define as different then some predefined default. If that is already possible or added then I can make add a set high quality differential equation solvers for the standard library.

I can work on making do with what is available already and implement some solvers or wait tell some numbat language editions so I don't have to redo any work. Let me know what works for you.

sharkdp:
> My preferred addition (if not already) is a way to initialize a struct of type Foo only listing items that you want to define as different then some predefined default. If that is already possible or added then I can make add a set high quality differential equation solvers for the standard library.

That's not yet possible, but I definitely would like to have that, too.



> I can work on making do with what is available already and implement some solvers or wait tell some numbat language editions so I don't have to redo any work. Let me know what works for you.

It's probably best to wait a bit. I am planning to merge this PR in it's current form soon.

Ryan-D-Gast:
> It's probably best to wait a bit. I am planning to merge this PR in it's current form soon.

I agree. 

I think this language has a lot of potential for use in numerical computing. For now this implementation should be sufficient for most use cases. In the future when new language features are added, I will work on a more comprehensive implementation for the standard library (assuming that is ok with you).

I enjoyed our back and forth conversations on this matter. I look forward to working more with you in the future.

PR #821, opened 2026-02-06
Add base_unit_of

sharkdp:
closes https://github.com/sharkdp/numbat/discussions/798


GrigorenkoPV:
Why check for zero here?

PR #840, opened 2026-02-14
Display litres as `L` instead of `l`

### LINKED ISSUE #839: Litres show up as lowercase `l`
Litres show up as a lowercase `l` which is, to my knowledge, a lot less commonly used than the uppercase `L`, and harder to read in some fonts:

```
> 20 L/s

    = 20 l/s    [FlowRate]
```

Could this be fixed by reversing the order from `l: short, L: short` to `L: short, l: short` here? https://github.com/sharkdp/numbat/blob/7d0365d37f52f6217f7bbcb0243a2a3389e06945/numbat/modules/units/si.nbt#L232

solson:
Fixes #839.


sharkdp:
Thank you!

PR #845, opened 2026-02-28
Add struct methods

Ryan-D-Gast:
# Add Inline Struct Methods With Struct-Scoped Dispatch

This PR adds inline struct methods (definition and calling) while keeping method resolution scoped to the struct.

## Example Syntax

```numbat
struct Vec<X: Dim> {
    x: X,
    y: X,

    fn scale(self, factor: Scalar) -> Self =
        Vec { x: self.x * factor, y: self.y * factor }

    fn dot<Y: Dim>(self, other: Vec<Y>) -> X * Y =
        self.x * other.x + self.y * other.y

    fn new(x: X, y: X) -> Self =
        Vec { x: x, y: y }
}

let v1 = Vec::new(1 m, 2 m)          # constructor-style call
let v2 = Vec { x: 3 m, y: 4 m }
let v2_cm = Vec { x: 300 cm, y: 400 cm }

let v3 = v1.scale(2)          # Vec { x: 2 m, y: 4 m }
let dp_m = v1.dot(v2)         # 11 m²
let dp_cm = v1.dot(v2_cm)     # 110_000 cm²
```

## Calling Modes

- Constructor-style calls: `Type::method(...)`
- Instance-style calls: `value.method(...)`

Type checking enforces that constructor methods are called with `::` and instance methods are called with `.`.

## Why This Design

I decided against a separate `impl` block approach (like Rust/Swift) because I wanted methods defined alongside the struct body and to avoid introducing additional language surface and constraints.

## Method Metadata + Dispatch (Simple Overview)

Struct methods are represented and executed through a struct-scoped path from typechecking to runtime:

- Each struct records method metadata (method name, definition span, and whether it is constructor-style or instance-style).
- Typed statements separate global functions from methods:
  - `DefineFunction` for global functions
  - `DefineMethod` for struct methods
- Typed method-call expressions carry an explicit method reference:
  - owner struct
  - method name
  - call kind (constructor vs instance)
- At runtime, the VM keeps a method callable table keyed by `(struct_name, method_name)`.
  - entries can point to either compiled method bytecode or foreign/native method callables.

This gives predictable struct-local lookup and avoids name collisions with global/prelude/imported functions.

## MISC

This PR relates with #795 and the broader direction discussed with @sharkdp around stronger struct features, which are needed for larger library work (including differential-equation-oriented code). I recognize this might not be a desired feature for the language, but I think it's a natural extension of the current struct syntax and semantics, and it unlocks a lot of expressiveness and code organization benefits.

### DIFF
--- a/book/src/basics/structs.md
+++ b/book/src/basics/structs.md
@@ -41,7 +41,7 @@ struct Element {
     atomic_number: Scalar,
     density: MassDensity,
 
-    fn tungsten() -> Self = Element {
+    fn tungsten() -> Self = Self {
         name: "Tungsten",
         atomic_number: 74,
         density: 19.25 g/cm³,
@@ -55,6 +55,8 @@ let tungsten = Element::tungsten()
 let side_length = tungsten.cube_side_length(1 kg) -> cm
 print("A 1 kg tungsten cube has side length {side_length:.2}.")
 ```
+
+Note that constructor methods can return `Self` or `Element` in this case, but the former is more concise and is preferred by convention.
 
 ## Generic structs
 
--- a/book/src/examples/example-numbat_syntax.md
+++ b/book/src/examples/example-numbat_syntax.md
@@ -138,7 +138,7 @@ struct Element {                 # Define a struct
     density: MassDensity,
 
     fn neptunium() -> Self =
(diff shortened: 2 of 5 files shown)


Ryan-D-Gast:
Closing this in favor of #847. The struct-method work from this branch has been consolidated into the draft PR, so this branch is now superseded.

PR #846, opened 2026-03-02
Add struct field defaults

Ryan-D-Gast:
# Add Struct Field Defaults

This PR adds field defaults for structs.

Depends on PR #845 branch. If/Once #845 is merged this can be rebased to main for merging. Or if #845 is not desired I can rebase this on main just LMK.

## Summary

- Adds syntax for field defaults in struct definitions:
  - `field_name: Type = expression`
- Allows omitted fields during struct instantiation when those fields have defaults.
- Keeps existing behavior for required fields:
  - missing fields without defaults still produce an error.
- Updates parser, AST, prefix transformation, and typechecker flow to support defaults.
- Adds interpreter and parser test coverage.
- Updates docs/examples (including syntax overview/book pages).

## Example

```numbat
struct PaperSize {
    width: Length = 210 mm,
    height: Length = 297 mm,
    name: String,
}

let a4 = PaperSize { name: "A4" }
let custom = PaperSize { name: "Custom", width: 100 mm }
let tall = PaperSize { name: "Tall", height: 400 mm }

assert_eq(a4.width, 210 mm)
assert_eq(a4.height, 297 mm)
assert_eq(custom.width, 100 mm)
assert_eq(custom.height, 297 mm)
assert_eq(tall.width, 210 mm)
assert_eq(tall.height, 400 mm)
```

## Semantics

- Field order in struct literals remains flexible.
- Provided values override defaults.
- Omitted defaulted fields are filled automatically.
- Omitted non-defaulted fields still error.

## Implementation Notes

- Field defaults are parsed from struct field declarations.
- Struct literals are normalized before typechecking by filling in omitted defaulted fields.
- Prefix/unit transformation is applied to default expressions as well.

Issue #462, opened 2024-06-11
Proper operator overloading for addition and subtraction


sharkdp:
We currently use operator overloading for `DateTime`s. We can perform the following three operations:

* `DateTime` - `DateTime` = `Time`
* `DateTime` + `Time` = `DateTime`
* `DateTime` - `Time` = `DateTime`

Every other addition/subtraction operation with `DateTime`s is disallowed. This is important because constructions like `DateTime` + `DateTime` are not meaningful.

The currently implemented operator overloading works fine whenever we have concrete types:
```
>>> now() + 4 days

    = Sat, 15 Jun 2024 18:44:56 +0200    [DateTime]
```
but if fails if type variables are involved (see similar issue for structs here: #459). For example, we currently can not do this — even though that used to work prior to #443:
```rs
fn id(x) = x
id(now()) + 4 days
```
Similarly, if we write the function
```rs
fn add_four_days(dt) = dt + 4 days
```
without a type annotation for the parameter type, we will not get the most general type inferred (which cannot even be expressed in Numbats type system at the moment).

Ideally, we would have something like a (multiparameter) type class or trait (`Add`, `Sub`), that we could use to model operator overloading. This would allow us to infer the most-general type (using a Rust-like syntax again):
```rs
fn add_four_days<A, B>(dt: A) -> B
  where A: Add<Time, Output = B>
  = …
```
which would allow us to use this function both with `DateTime` and `Time`.

---

If it were only for `DateTime`, I think we might be able to live with the current restriction. But there might be some other cool examples where we could make use of proper `Add`/`Sub` constraint handling. For example, we could introduce specialized types for `Celsius` and `Fahrenheit` that would potentially implement (in close analogy to what we have for `DateTime` and `Time`):
* `Celsius` - `Celsius` = `Kelvin`
* `Celsius` + `Kelvin` = `Celsius`
* `Celsius` - `Kelvin` = `Celsius` (we could think about disallowing this as well…)
but disallow things like `Celsius` + `Celsius` because they make no sense.

Ref: #184 

Another use case might be logarithmic units like decibel, for which we could implement specialized addition/subtraction operations.

Ref: #387 

We could start by only allowing those constraints to be resolved internally by the type checker, without going all in on user-definable type classes. But those would be interesting as well, obviously:
```rs
struct Vec {
  x: Length,
  y: Length,
}

impl Add<Vec> for Vec {
  type Output = Vec;

[...]

PR #847, opened 2026-03-13
Add struct defaults, methods, and operator overloading decorators

### LINKED ISSUE #459: Struct field access fails if expression type is not yet concrete
I was playing around with the new type system and came across this issue:
```
>>> struct F { a: Current, b: Current }
>>> head([F { a: 1A, b: 2A }]).b
error: while type checking
  ┌─ <input:24>:1:28
  │
1 │ head([F { a: 1A, b: 2A }]).b
  │ -------------------------- ^ Can not access field 'b' of non struct type 'T185'
  │ │
  │ T185
```

My guess is there just needs to be a zonking step performed before looking up the struct type for field accesses?

Ryan-D-Gast:
This draft consolidates the changes from #845 and #846 into one branch. Read this PR as the combined follow-up, not as an independent line of work.

## Summary

This adds struct-owned operator overloads via inline method decorators and moves `+/-/*//` resolution onto the same deferred-constraint path used for other inference-sensitive operations.

This branch is currently stacked on top of:

- #845, which introduced inline struct methods and struct-scoped dispatch
- #846, which added struct field defaults

So this PR should be read as a stacked draft for now, not as an independently mergeable branch against `main` in its current form.

It is also closely related to:

- #459, which exposed the same general inference timing problem for struct field access
- #462, which asks for proper deferred handling of overloaded `Add`/`Sub` operations

## Motivation

The immediate motivation is #462: built-in `DateTime` operator overloading currently fails when an intermediate type variable is involved, for example:

```numbat
fn id(x) = x
id(now()) + 4 days
```

This change fixes that deferred-resolution case and also exposes a constrained, coherence-friendly way for user-defined structs to participate in operator overloading.

The implementation follows the same general direction that fixed #459: do not require the expression type to already be concrete at the first lookup site, and instead carry a deferred constraint until more type information is available.

## Syntax

Operator overloads are declared on instance methods inside a `struct` body:

```numbat
struct Point {
    x: Scalar,
    y: Scalar,

    @add
    fn add(self, rhs: Self) -> Self =
        Point { x: self.x + rhs.x, y: self.y + rhs.y }

    @add
    fn add_scalar(self, rhs: Scalar) -> Self =
        Point { x: self.x + rhs, y: self.y + rhs }
}

let p1 = Point { x: 1, y: 2 }
let p2 = Point { x: 3, y: 4 }

let a = p1 + p2
let b = p1 + 3
```

Supported decorators in this PR:

- `@add`
- `@sub`
- `@mul`
- `@div`

For these decorators, the compiler infers:

- the lhs type from `self`
- the rhs type from the method's second parameter
- the output type from the method return type

These decorators are only valid on struct instance methods. Top-level operator decorators are rejected, and overloads remain directional rather than generating an implicit reversed form.

## Design

This intentionally takes a narrower path than a full trait/typeclass system.

Instead of introducing user-visible `Add`/`Sub` traits and qualified operator bounds, operator implementations are owned by the lhs struct and lowered into an internal operator-implementation table. Binary expressions can emit deferred operator constraints, which are resolved after more type information is known.

That gives us:

- a fix for the deferred `DateTime` case from #462
- a solver pattern that mirrors the successful deferred-resolution approach from #459
- struct-local coherence without orphan-style free-floating overloads
- support for multiple overloads on the same lhs with different rhs types
- a much smaller language and type-system expansion than full traits/typeclasses

What this does **not** do is infer or expose most-general operator-bounded function types such as a hypothetical:

```numbat
fn add_four_days<A, B>(dt: A) -> B
  where A: Add<Time, Output = B>
```

That remains future work if Numbat eventually grows a real trait/typeclass mechanism.

## Implementation Notes

- Adds parser/decorator support for `@add/@sub/@mul/@div` on struct methods.
- Infers rhs and output operator types from the decorated method signature instead of requiring them to be restated in the decorator.
- Validates that decorated methods are instance methods with exactly one rhs parameter.
- Introduces deferred binary-operator constraints in the typechecker.
- Resolves built-in `DateTime` operator behavior and struct-owned operator methods through the same deferred path.
- Rewrites resolved deferred struct operators into explicit typed method calls before bytecode compilation.

## Tests

Added coverage for:

- parser acceptance/rejection of operator decorators
- typechecking of valid operator methods
- invalid decorated method signatures
- multiple overloads for the same operator with different rhs types
- runtime dispatch for struct operator overloads
- the `id(now()) + 4 days` deferred-resolution case from #462

Addresses #462.
Related: #459, #845, #846.

### DIFF
--- a/book/src/prelude/functions/lists.md
+++ b/book/src/prelude/functions/lists.md
@@ -152,8 +152,6 @@ fn drop<A>(n: Scalar, xs: List<A>) -> List<A>
 ### `element_at`
 Get the element at index `i` in a list.
 
-This is equivalent to the indexing syntax `xs[i]`.
-
 ```nbt
 fn element_at<A>(i: Scalar, xs: List<A>) -> A
 ```
@@ -399,3 +397,4 @@ fn split(input: String, separator: String) -> List<String>
         = ["Numbat", "is", "a", "statically", "typed", "programming", "language."]    [List<String>]
     ```
     [:material-play-circle: Run this example](https://numbat.dev/?q=split%28%22Numbat%20is%20a%20statically%20typed%20programming%20language%2E%22%2C%20%22%20%22%29){ .md-button }
+


Ryan-D-Gast:
Added two follow-up commits on top of this draft:

- `ba0594e8` adds indexing support for structs via `@index` and builtin indexing for `List`, so `xs[i]` now works directly for lists.
- `e61cff4a` adds a runnable matrix example that uses nested lists for storage, struct indexing for `matrix[row, col]`, and `@mul` for matrix multiplication.

Example:
```numbat
struct Matrix {
    rows: Scalar,
    cols: Scalar,
    data: List<List<Scalar>>,

    fn from_rows(data: List<List<Scalar>>) -> Self =
        Self {
            rows: len(data),
            cols: if is_empty(data) then 0 else len(data[0]),
            data: data,
        }

    @index
    fn get(self, row: Scalar, col: Scalar) -> Scalar =
        self.data[row][col]

    fn dot_product(lhs: Self, rhs: Self, row: Scalar, col: Scalar, i: Scalar) -> Scalar =
        if i == lhs.cols
            then 0
            else lhs[row, i] * rhs[i, col] + Matrix::dot_product(lhs, rhs, row, col, i + 1)

    fn product_row(lhs: Self, rhs: Self, row: Scalar, col: Scalar) -> List<Scalar> =
        if col == rhs.cols
            then []
            else cons(
                Matrix::dot_product(lhs, rhs, row, col, 0),
                Matrix::product_row(lhs, rhs, row, col + 1),
            )

    fn product_rows(lhs: Self, rhs: Self, row: Scalar) -> List<List<Scalar>> =
        if row == lhs.rows
            then []
            else cons(
                Matrix::product_row(lhs, rhs, row, 0),
                Matrix::product_rows(lhs, rhs, row + 1),
            )

    @mul
    fn mul(self, rhs: Self) -> Self =
        Matrix::from_rows(Matrix::product_rows(self, rhs, 0))
}

let identity_2 = Matrix::from_rows([
    [1, 0],
    [0, 1],
])

let wide = Matrix::from_rows([
    [1, 2, 3],
    [4, 5, 6],
])

let tall = Matrix::from_rows([
    [1, 2],
    [3, 4],
    [5, 6],
])

let product = wide * tall
let unchanged = identity_2 * identity_2

assert_eq(identity_2.rows, 2)
assert_eq(identity_2.cols, 2)
assert_eq(identity_2[1, 1], 1)

assert_eq(wide.rows, 2)
assert_eq(wide.cols, 3)
assert_eq(wide[0, 2], 3)
assert_eq(wide[1, 0], 4)

assert_eq(tall.rows, 3)
assert_eq(tall.cols, 2)
assert_eq(tall[2, 1], 6)

assert_eq(product.rows, 2)
assert_eq(product.cols, 2)
assert_eq(product[0, 0], 22)
assert_eq(product[0, 1], 28)
assert_eq(product[1, 0], 49)
assert_eq(product[1, 1], 64)

assert_eq(unchanged[0, 0], 1)
assert_eq(unchanged[1, 1], 1)

print("2x2 identity: {identity_2}")
print("2x3 matrix:   {wide}")
print("3x2 matrix:   {tall}")
print("product:      {product}")

```

This is mainly meant to demonstrate the intended use case for the new indexing/operator-overloading features. I didn't want to do PR spam so I just included it in this draft. Obviously a built in Matrix/Tensor type is optimal so this example of a pure native numbat matrix type is more so to show the use case rather then recommending this be in the standard library.

Ryan-D-Gast:
This draft PR consolidates the work from #845 and #846 into one branch. I also ran `cd book && uv run build` locally to address the documentation-check failure, and the build completed successfully in this worktree.

Issue #851, opened 2026-03-24
`bin(inf)` and others cause hang by infinite loop


GideonBear:
```console
$ numbat -e "bin(1e308)"
"0x8e679c2f5e450000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
$ numbat -e "bin(inf)"
# hangs forever
```

Same problem for `hex`, `dec`, `oct`, `bin`, `base`, and with `NaN` instead of `inf`.

This is especially annoying in an interactive prompt, where Ctrl+C does not work.

GideonBear:
There are other functions that have this issue (I've found `str_repeat` so far). I am unsure what to do here, I wanted to add a check to `base` but maybe a language-level change is in order.

sharkdp:
> maybe a language-level change is in order

Hm yes, it seems reasonable to do something like Python with a maximum recursion limit. Until then, I've merged #853

sidsri14:
Picking up the str_repeat half of this: #889 guards both inf and NaN repetition counts with a user error instead of hanging, with regression examples for each branch (finite behavior unchanged).

PR #853, opened 2026-04-04
Fix infinite recursion for bin(inf), hex(NaN), etc.

amogusussy:
As mentioned in #851, when `bin(inf)` is called, numbat just enters a recursive loop that never ends. This PR fixes that in the one case of converting bases.

Perhaps a maximum recursion depth could be introduced to stop this from happening with other functions like `str_repeat`.


sharkdp (approved):
Thank you!

PR #863, opened 2026-06-04
Respect XDG dirs on macOS when set

### LINKED ISSUE #744: Follow XDG dirs if set on macos

ded-furby:
Fixes #744.

## Summary
- prefer `XDG_CONFIG_HOME` on macOS when it is set before falling back to the platform config directory
- prefer `XDG_DATA_HOME` on macOS when it is set before falling back to the platform data directory
- add focused tests for the new path-resolution behavior

## Validation
- `cargo fmt --all`
- `cargo test -p numbat-cli app_dir_`
- `cargo test -p numbat-cli --test integration`


sharkdp:

PR #864, opened 2026-06-05
Respect XDG dirs on macOS when set

### LINKED ISSUE #744: Follow XDG dirs if set on macos

ded-furby:
Closes #744

## Summary
- prefer `XDG_CONFIG_HOME` and `XDG_DATA_HOME` on macOS when those variables are set
- keep the existing macOS defaults when the XDG variables are unset
- document the config-path behavior and add coverage for the helper that reads XDG overrides

## Validation
- cargo fmt --all
- cargo test -p numbat-cli
- (cd book && uv run build)

I noticed the issue comment pointing at `etcetera`, but kept this change scoped to the two existing path lookups so the default macOS layout stays unchanged unless the user explicitly opts into XDG.


sharkdp:

PR #870, opened 2026-07-07
Fix binom for negative integer n

upuddu:
`binom` already computes the generalized binomial coefficient (it works for fractional `n`), but the `k > n` shortcut returned `0` for every integer `n`, including negatives - so `binom(-1, 0)` gave `0` instead of `1`, and `binom(-1, 2)` gave `0` instead of the binomial-series coefficient `1`. The shortcut is only valid for non-negative integer `n`, so I restricted it to that case and added tests for negative `n`.


sharkdp (approved):
Thank you!

PR #872, opened 2026-07-12
Fix Unix timestamp conversion precision

### LINKED ISSUE #871: -1ms error from DateTime to unixtime_ms
Convert a `unixtime_ms` number to `DateTime` and varsus, the result has a minus 1ms error:
```nbt
>>> let ts = from_unixtime_ms(1783598716223)

  let ts: DateTime = from_unixtime_ms(1_783_598_716_223)

>>> ts

  ts

    = 2026-07-09 20:05:16 CST (UTC +08), Asia/Shanghai    [DateTime]

>>> let ts2 = from_unixtime_ms(1783591946381)

  let ts2: DateTime = from_unixtime_ms(1_783_591_946_381)

>>> ts2

  ts2

    = 2026-07-09 18:12:26 CST (UTC +08), Asia/Shanghai    [DateTime]

>>> ts2 -> unixtime_ms

  unixtime_ms(ts2)

    = 1_783_591_946_380

>>> ts -> unixtime_ms

  unixtime_ms(ts)

    = 1_783_598_716_222
```

environment:
```
> cargo install numbat-cli
    Updating `sysu-sparse` index
     Ignored package `numbat-cli v1.23.0` is already installed, use --force to override
```

This error is caused by [the convertion first to `unixtime` in float second then `floor_in(unix_ms)`](https://github.com/sharkdp/numbat/blame/88b2e81a35d06942101d74e9b368081f02555f4f/numbat/modules/datetime/unixtime.nbt#L30):
```
@description("Converts a `DateTime` to a UNIX timestamp. Can be used on the right hand side of a conversion operator: `now() -> unixtime`.")
@example("datetime(\"2022-07-20 21:52 +0200\") -> unixtime")
fn unixtime(input: DateTime) -> UnixTime = _unixtime_µs(input) * unix_µs -> unix_s

@description("Converts a `DateTime` to a UNIX timestamp in milliseconds.")
@example("datetime(\"2022-07-20 21:52:05.123 +0200\") -> unixtime_ms")
fn unixtime_ms(input: DateTime) -> Scalar = unixtime(input) |> floor_in(unix_ms) |> value_of
```

I advice to keep `unixtime` as `unix_µs`，then convert to `unix_ms`、`unix_s` at the final step.

arturict:
## What changed

- derive second and millisecond timestamps directly from the exact microsecond value returned by the DateTime FFI
- return that exact value for `unixtime_µs` instead of converting through floating-point seconds
- add regression coverage for the two millisecond timestamps from the report

## Why

Converting microseconds to `unix_s` and then back to milliseconds could produce a value just below an integer boundary. `floor_in(unix_ms)` would consequently lose one millisecond. Keeping integer-valued microseconds until the final division avoids that intermediate rounding.

Fixes #871

## Validation

- reproduced the original `-1 ms` results with the official v1.23.0 Windows binary
- loaded the modified module through `NUMBAT_MODULES_PATH` and verified exact round trips for both reported timestamps, epoch zero, a pre-epoch value, microseconds, and seconds
- `cargo fmt --check`
- `git diff --check`

The focused Rust test target could not be compiled locally because this Windows host has Rust but not the MSVC linker. No system SDK was installed; GitHub CI provides native build/test coverage across the supported targets.

### DIFF
--- a/examples/tests/datetime.nbt
+++ b/examples/tests/datetime.nbt
@@ -117,6 +117,10 @@ assert_eq(dt_unixtime_2 -> unixtime_s -> from_unixtime_s -> unixtime_s, 16583467
 assert_eq(dt_unixtime_2 -> unixtime_ms -> from_unixtime_ms -> unixtime_ms, 1658346725123)
 assert_eq(dt_unixtime_2 -> unixtime_µs -> from_unixtime_µs -> unixtime_µs, 1658346725123456)
 
+# Regression test for #871
+assert_eq(from_unixtime_ms(1783591946381) -> unixtime_ms, 1783591946381)
+assert_eq(from_unixtime_ms(1783598716223) -> unixtime_ms, 1783598716223)
+
 
 # Calendar arithmetic
 
--- a/numbat/modules/datetime/unixtime.nbt
+++ b/numbat/modules/datetime/unixtime.nbt
@@ -23,15 +23,15 @@ fn unixtime(input: DateTime) -> UnixTime = _unixtime_µs(input) * unix_µs -> un
 
 @description("Converts a `DateTime` to a UNIX timestamp in seconds.")
 @example("datetime(\"2022-07-20 21:52 +0200\") -> unixtime_s")
-fn unixtime_s(input: DateTime) -> Scalar = unixtime(input) |> floor_in(unix_s) |> value_of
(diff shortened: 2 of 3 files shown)


sharkdp (approved):
Thank you!

PR #876, opened 2026-07-23
Add Woxi to comparison table

ad-si:


sharkdp:
Thank you. I don't really think this adds a lot of value for Numbat users though, given that we have a Wolfram (Alpha) column already. Good luck with your project -- looks interesting!

ad-si:
All the other tools are free offline/local tools, whereas Wolfram Alpha is a commercial online tool. I'd argue that removing Wolfram Alpha, rather than Woxi, makes more sense.

Since Woxi has now become very capable, correctly (and efficiently) implements the majority of the Wolfram Language, is free and open source, and has many frontends, I think it's an even better representative of the Wolfram Language.

PR #888, opened 2026-09-01
Load currency units on demand for `info` and `list` commands

### LINKED ISSUE #635: `info USD` (or any alias of `USD`) doesn't work until a currency calculation is performed
```
>>> info USD
  Not found
>>> 1USD

  1 dollar

    = 1 $    [Money]

>>> info USD
  Unit: US dollar (https://en.wikipedia.org/wiki/United_States_dollar)
  Aliases: dollar, dollars, USD, usd, $
  A unit of: Money
  
  1 dollar = 0.92584 €
```

I assume that this is related to the lazy loading of currencies. Should currencies be loaded if `info` triggers a use?

Jorge-Polanco-Roque:
Fixes #635.

## Problem
`info USD` (or any currency alias) reports `Not found` until the user first performs a currency calculation. On a fresh session the currency units simply aren't there yet.

## Root cause
On-demand loading of `units::currencies` was wired **only** into the expression-evaluation path (`interpret_with_settings`, triggered by an `UnknownIdentifier` type error). The `info` and `list` commands bypass the interpreter, so currencies were never pulled in for them.

## Approach
Extracted the existing inline lazy-load block into reusable `Context` methods and call them from the command handlers:
- `info <keyword>` loads currencies on demand when the keyword names a currency.
- `list` / `list units` load unconditionally; `list functions/dimensions/variables` do not.

The expression path is unchanged — it now calls the same shared method. No-prelude mode is unaffected (`load_currency_module_on_demand` stays false → early return).

## Testing
- Added `test_info_and_list_load_currencies_on_demand` — verified it **fails without the fix** (`info USD` → `Not found`).
- `cargo fmt --check` clean · `cargo clippy -p numbat` no new warnings · `cargo test -p numbat` green (lib 176, interpreter 53, prelude 6).

A maintainer suggested exactly this approach (load for `info`/`list`, not for `save`/`help`/`quit`) in the issue thread.

### DIFF
--- a/numbat/src/command.rs
+++ b/numbat/src/command.rs
@@ -94,6 +94,21 @@ impl From<RuntimeError> for CommandError {
     }
 }
 
+impl From<Box<crate::NumbatError>> for CommandError {
+    fn from(err: Box<crate::NumbatError>) -> Self {
+        let runtime = match *err {
+            crate::NumbatError::RuntimeError(e) => e,
+            // The on-demand currency load only surfaces runtime errors; other variants
+            // would indicate a bug in the internal `use units::currencies` module.
+            other => RuntimeError {
+                kind: crate::RuntimeErrorKind::UserError(other.to_string()),
+                backtrace: Vec::new(),
+            },
+        };
+        Self::Runtime(Box::new(runtime))
+    }
+}
+
 impl ErrorDiagnostic for ResolverDiagnostic<'_, CommandError> {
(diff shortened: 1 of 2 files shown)


Jorge-Polanco-Roque:
Gentle ping — open about a week. It loads currency units on demand for the `info` and `list` commands so they don't pay the network fetch when it isn't needed. Happy to rebase or adjust anything. Thanks for numbat!

Issue #650, opened 2024-11-03
[Startup performance] Stop getting the `local_timezone` when it's not required


irevoire:
Part of https://github.com/sharkdp/numbat/issues/525

## The culprit

Hey, I profiled numbat once again while running `1 + 1` and noticed that 23% (~10ms out of 40) of the time was spent running the `ffi::get_local_timezone` function of numbat.
<img width="1203" alt="image" src="https://github.com/user-attachments/assets/bc2e3683-0fb7-40af-8d11-6d9431acf25d">

The call comes from this line: https://github.com/sharkdp/numbat/blob/84704073de791a8266653de9c09af80a29f21e16/numbat/modules/datetime/functions.nbt#L28-L29

And after commenting it and benchmarking the startup time with hyperfine here are the results I got:
```
% hyperfine './numbat-master -e "1 + 1"' './numbat-without-tz -e "1 + 1"' --warmup 10
Benchmark 1: ./numbat-master -e "1 + 1"
  Time (mean ± σ):      41.4 ms ±   1.0 ms    [User: 30.5 ms, System: 10.2 ms]
  Range (min … max):    39.9 ms …  45.0 ms    69 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 2: ./numbat-without-tz -e "1 + 1"
  Time (mean ± σ):      32.4 ms ±   1.6 ms    [User: 30.3 ms, System: 1.4 ms]
  Range (min … max):    30.6 ms …  36.1 ms    86 runs
 
Summary
  ./numbat-without-tz -e "1 + 1" ran
    1.28 ± 0.07 times faster than ./numbat-master -e "1 + 1"
```

So it’s pretty much what I measured with my profiler, 1.28 times or ~10ms faster.

## The solution

I’m not sure of what should be the solution here though, maybe make the whole `local` variable a ffi function?
Not a fan though, maybe you’ll have a better idea

sharkdp:
Thank you! Not sure how we missed this in profiles so far. I guess it became slower with #511 because now we're loading the tz database from disk.

Conceptually, this is similar to the problem we had with currency conversion factors. We want to initialize variables/units with a value that is expensive to look up. Lazy loading would be an obvious solution here, but might be a bit tricky to implement? I haven't thought deeply about it but maybe, instead of calling `get_local_timezone()` at startup time, we could set
```
[...]

PR #909, opened 2026-09-19
perf: lazily resolve the local timezone

efegokdemir:
## Summary

Addresses #650.

- Replace the eager `tz(get_local_timezone())` initialization with a lightweight `_local()` FFI function.
- Resolve the system timezone only when the resulting conversion function is used.
- Preserve the existing `datetime -> local` syntax and output.
- Add a regression test and update the affected snapshot.

This follows the lightweight FFI-wrapper approach suggested in the issue discussion, without introducing general-purpose lazy evaluation.

## Validation

- `cargo fmt --all -- --check`
- `cargo test --workspace`
- `git diff --check`
- Compared six expressions against the original binary; all outputs and exit statuses matched.

## Local benchmark

Paired measurements of `numbat -e '1 + 1'` on one Mac, using debug builds:

- Before: median 187.99 ms
- After: median 186.38 ms
- Difference: 1.61 ms (0.9%)

These measurements include full process startup. The observed improvement is modest, and the benchmark does not establish statistical significance.

Opening as a draft for implementation review and CI validation.


efegokdemir:
The failing musl targets exposed a real portability issue: `TimeZone::system()` can have no IANA name when timezone data is unavailable, while the new regression compared it with `tz(get_local_timezone())`. Commit `9e4761d` makes the existing `get_local_timezone_or_utc()` fallback explicit: named system zones are preserved, otherwise both local conversion and the exposed timezone use UTC.

Validation: `cargo test -p numbat --test interpreter test_lazy_local_timezone_conversion` passed, `cargo fmt --all` passed, and `git diff --check` passed. The commit is DCO-signed and GitHub-verified. The full workflow is rerunning.

